PR #1253, opened 2022-12-07
Typed literals support in lexer.

Alon-Ti:
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/1253)


orizi (changes requested):
> Reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/1253)* status: all files reviewed, 1 unresolved discussion (waiting on @Alon-Ti)

*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/1253#-NIgoMRCEnyjEqmIwfXX:-NIgoMRCEnyjEqmIwfXY:bv2vike) (no related file):*
add a failing test to the syntax diagnostics as well.

___

Alon-Ti (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/1253)* status: all files reviewed, 1 unresolved discussion (waiting on @orizi)

*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/1253#-NIgoMRCEnyjEqmIwfXX:-NIh-8966ROye8V-vU03:b-i17epk) (no related file):*

add a failing test to the syntax diagnostics as well.
</blockquote></details>

With what input? This expands the current syntax, nothing new should fail.

___

orizi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/1253)* status: all files reviewed, 1 unresolved discussion (waiting on @yuvalsw)

*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/1253#-NIgoMRCEnyjEqmIwfXX:-NIh8aku77Ik_NAsiFHd:bmmasd1) (no related file):*
<details><summary><i>Previously, Alon-Ti wrote…</i></summary><blockquote>

With what input? This expands the current syntax, nothing new should fail.
</blockquote></details>

adding @yuvalsw - as we maybe should probably support this in the parsing stage.

___

yuvalsw (changes requested):
> Reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/1253)* status: all files reviewed, 3 unresolved discussions (waiting on @Alon-Ti and @orizi)

*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/1253#-NIgoMRCEnyjEqmIwfXX:-NIkuVCa7k4hvb5EKiY1:bhv2y6d) (no related file):*

adding @yuvalsw - as we maybe should probably support this in the parsing stage.
</blockquote></details>

I was thinking about this. I am still not sure if one of the ways have significant pros/cons.
Rust do it in the lexer as here (they don't enforce _, which I am not sure we should, too, but we can change this later).
I am OK with continuing this way.

___
___
*[`crates/parser/src/lexer.rs` line 122 at r1](https://reviewable.io/reviews/starkware-libs/cairo/1253#-NIl2Z6i5R5fBNXMzHtH:-NIl2Z6i5R5fBNXMzHtI:b-26bud8) ([raw file](https://github.com/starkware-libs/cairo/blob/ee82ba982e66cc84b76d645cfcee2b892a84a800/crates/parser/src/lexer.rs#L122)):*
> ```XML
> 
>         // The token does not start with '0x'. Parse the token as a decimal number.
>         self.take_while(|c| c.is_ascii_digit());
> ```

It should still work, but to make it easier to read/understand, if you remove the return, please put this in the else.
(also otherwise the comment is wrong)



_Code quote:_
```XML
self.take_while(|c| c.is_ascii_digit());
```
___
*[`crates/parser/src/lexer.rs` line 125 at r1](https://reviewable.io/reviews/starkware-libs/cairo/1253#-NIl2sBo8BgVpKHAgIi5:-NIl2sBo8BgVpKHAgIi6:b-jb73ig) ([raw file](https://github.com/starkware-libs/cairo/blob/ee82ba982e66cc84b76d645cfcee2b892a84a800/crates/parser/src/lexer.rs#L125)):*
> ```XML
> 
>         // Parse _type suffix.
>         if self.peek() == Some('_') {
> ```

do we want to enforce _?
Why not support 3u128 as in Rust?



_Code quote:_
```XML
'_
```

Alon-Ti (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/1253)* status: all files reviewed, 2 unresolved discussions (waiting on @yuvalsw)

___
*[`crates/parser/src/lexer.rs` line 122 at r1](https://reviewable.io/reviews/starkware-libs/cairo/1253#-NIl2Z6i5R5fBNXMzHtH:-NIlgtzn2Q9790LrgBmJ:b-i8cfwv) ([raw file](https://github.com/starkware-libs/cairo/blob/ee82ba982e66cc84b76d645cfcee2b892a84a800/crates/parser/src/lexer.rs#L122)):*

It should still work, but to make it easier to read/understand, if you remove the return, please put this in the else.
(also otherwise the comment is wrong)
</blockquote></details>

This doesn't work because the `else` refers only to the first `peek` (that looks for a `0`), so this won't catch numbers like `0123`. I could unify the if with short circuiting and it would do the same thing, but I think that's even less readable.

The comment is still correct because anything this iterator can read will have been exhausted previously if the number does start with `0x`.
___
*[`crates/parser/src/lexer.rs` line 125 at r1](https://reviewable.io/reviews/starkware-libs/cairo/1253#-NIl2sBo8BgVpKHAgIi5:-NIlTFy50WKUKHCfmxHA:bg8xp3v) ([raw file](https://github.com/starkware-libs/cairo/blob/ee82ba982e66cc84b76d645cfcee2b892a84a800/crates/parser/src/lexer.rs#L125)):*

do we want to enforce _?
Why not support 3u128 as in Rust?
</blockquote></details>

I think we should keep the `_` (Lior agrees), because `123_u128` is more readable than `123u128` (and to a lesser extent because if we ever want to support literals of types starting with `_` it can introduce an ambiguity).

yuvalsw (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/1253)* status: all files reviewed, 2 unresolved discussions (waiting on @Alon-Ti)

___
*[`crates/parser/src/lexer.rs` line 122 at r1](https://reviewable.io/reviews/starkware-libs/cairo/1253#-NIl2Z6i5R5fBNXMzHtH:-NIlkvgmCbIUDxgcV2Xd:bz352pu) ([raw file](https://github.com/starkware-libs/cairo/blob/ee82ba982e66cc84b76d645cfcee2b892a84a800/crates/parser/src/lexer.rs#L122)):*
<details><summary><i>Previously, Alon-Ti wrote…</i></summary><blockquote>

This doesn't work because the `else` refers only to the first `peek` (that looks for a `0`), so this won't catch numbers like `0123`. I could unify the if with short circuiting and it would do the same thing, but I think that's even less readable.

The comment is still correct because anything this iterator can read will have been exhausted previously if the number does start with `0x`.
</blockquote></details>

Ok, you can leave it this way. But I disagree that the comment it correct. You get to the comment even for tokens that start with 0x.
You should say something like:
"if the token starts with 0x, it is already fully consumed. Parse the token as a decimal number."
___
*[`crates/parser/src/lexer.rs` line 125 at r1](https://reviewable.io/reviews/starkware-libs/cairo/1253#-NIl2sBo8BgVpKHAgIi5:-NIllRK78zCwbC5D_sNR:b-650554) ([raw file](https://github.com/starkware-libs/cairo/blob/ee82ba982e66cc84b76d645cfcee2b892a84a800/crates/parser/src/lexer.rs#L125)):*
<details><summary><i>Previously, Alon-Ti wrote…</i></summary><blockquote>

I think we should keep the `_` (Lior agrees), because `123_u128` is more readable than `123u128` (and to a lesser extent because if we ever want to support literals of types starting with `_` it can introduce an ambiguity).
</blockquote></details>

1. OK, I don't really mind.
2. I don't think we ever want to have types starting with _...

Alon-Ti (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/1253)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @orizi and @yuvalsw)

___
*[`crates/parser/src/lexer.rs` line 122 at r1](https://reviewable.io/reviews/starkware-libs/cairo/1253#-NIl2Z6i5R5fBNXMzHtH:-NIlpX1D8rwSJGtWBE0l:b-896fix) ([raw file](https://github.com/starkware-libs/cairo/blob/ee82ba982e66cc84b76d645cfcee2b892a84a800/crates/parser/src/lexer.rs#L122)):*

Ok, you can leave it this way. But I disagree that the comment it correct. You get to the comment even for tokens that start with 0x.
You should say something like:
"if the token starts with 0x, it is already fully consumed. Parse the token as a decimal number."
</blockquote></details>

Done.
___
*[`crates/parser/src/lexer.rs` line 125 at r1](https://reviewable.io/reviews/starkware-libs/cairo/1253#-NIl2sBo8BgVpKHAgIi5:-NIlpI9n1-lZ5P88Rk4H:b-xryxi4) ([raw file](https://github.com/starkware-libs/cairo/blob/ee82ba982e66cc84b76d645cfcee2b892a84a800/crates/parser/src/lexer.rs#L125)):*

1. OK, I don't really mind.
2. I don't think we ever want to have types starting with _...
</blockquote></details>

I agree, this is mostly relevant if in some distant future we want to allow literals for user-defined types.

orizi (approved):
> Reviewed 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/1253)* status: all files reviewed, 1 unresolved discussion (waiting on @yuvalsw)

yuvalsw (approved):
:lgtm:

> Reviewed 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/1253#-:-NIlzg6BDe2RUtgG7c_O:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @Alon-Ti)

Alon-Ti (commented):
> Reviewed 1 of 2 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/1253)* status: all files reviewed, 1 unresolved discussion (waiting on @Alon-Ti)

Alon-Ti (commented):
> Dismissed @yuvalsw from a discussion.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/1253)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Alon-Ti)

Issue #1595, opened 2023-01-01
bug: cairo_filesystem detect_corelib fails when used as a git dependency


shramee:
# Bug Report

**Cairo version:**

74b4d128ccb199045abde158ee247e5a351f3cae

**Current behavior:**

thread 'main' panicked at 'Corelib not found.', /Users/shramee/.cargo/git/checkouts/cairo-f086c7e6d4098a68/74b4d12/crates/cairo-filesystem/src/detect.rs:34:5

When using `cairo_compiler::db::RootDatabase;` to get default `let mut db_val = RootDatabase::default();`

Using crates from the git repo like so,

```toml
# . . .
[dependencies]

# Members of starkware-libs/cairo
# . . .
cairo-filesystem = { git = "https://github.com/starkware-libs/cairo" }
cairo-compiler = { git = "https://github.com/starkware-libs/cairo" }
# . . .
```

**Expected behavior:**

Corelib dir to be found from installed git crate path.

**Steps to reproduce:**

**Related code:**

Tiniest reproducible version:

Issue is in `cairo_filesystem::detect::detect_corelib`.

#### Cargo.toml
```toml
[package]
name = "cairo-corelibs-bug"
version = "0.1.0"
edition = "2021"

[dependencies]
cairo-filesystem = { git = "https://github.com/starkware-libs/cairo" }

[[bin]]
name = "test"
path = "test.rs"
```

#### test.rs
```rs
use cairo_filesystem::detect::detect_corelib;
fn main() {
    let path = detect_corelib();
    print!("Found corelibs at {:#?}.", path);
}
```

**Other information:**

mkaput:
@maciektr will work on this. Our goal is to provide an API to provide custom paths to the `corelib` and in the future I foresee giving an ability to provide the `corelib` via virtual filesystem objects.

shramee:
I second virtual FS corelib. Projects shouldn't have to include `corelibs` in their project files. They should be a part of... err... Cairo `core` crate...

spapinistarkware:
I think this is fixed now. You can probably check out how Scarb are doing it. If it's still an issue, I'll reopen

PR #1596, opened 2023-01-01
Check current file path parents for corelib

shramee:
Addresses #1595

### Background

I am writing a creative coding framework for Cairo [shramee/creative-cairo](https://github.com/shramee/creative-cairo). And I need to use Cairo crates as a git dependency. The issue I am facing is highlighted in #1595.

### Changes

1. Added `check_parents_for_corelib`macro to test `PathBuf` parents for `corelib`
2. Added current file (and effectively current `crate`) path candidate to look at while finding `corelib`.

Here are some logs to show how this works (logs have now been removed from code.

Screenshot 2023-01-02 at 2 26 11 AM

This is where we had those logs...
```diff
macro_rules! check_parents_for_corelib {
    ($dir:ident) => {
-       println!("Checking directory {:#?}", $dir);
        $dir.pop();
        $dir.push("corelib");
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/1596)


mkaput (commented):
I have exactly the same issue as you have in the package manager, and I think the change here don't *really* solve the problem.

1. For your project, I believe a much better solution would be to include corelib in the program's memory and let the compiler read from it instead from the fs. Imagine how could you deploy your project as a website via WASM?
2. For package manager/build tool, the corelib should reside in some global directory ("compiler installation") managed by this tool.

I originally planned to discuss this topic further with the team internally, but given this PR I'll post this here:

I think the proper solution would be to remove this detection logic entirely and let users of the `cairo-compiler` crate provide API to give a virtual filesystem object for each crate user might want to provide builtin. In your framework, you could then bundle corelib compile-time in Golang's `embed` fashion, while the build-tool could point the compiler to well-known path.

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/1596#-:-NKqd9lA74YMJVjJuNyk:b-tvrgn2)* status: 0 of 1 files reviewed, all discussions resolved

mkaput (commented):
This doesn't of course mean we cannot provide ready-to use virtual fs that lookup parent directories :)

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/1596#-:-NKqebe18jJtjGBjpmZy:bddygyd)* status: 0 of 1 files reviewed, all discussions resolved

spapinistarkware:
This issue is being worked on by @mkaput

PR #1970, opened 2023-01-30
Make format path/stdin API public 

maciektr:
Justification: In order to implement project formatting API in Scarb (https://github.com/software-mansion/scarb/issues/31) we suggest making the path formatting currently under formatter binary public to allow usage as a crate.

- - -
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/1970)

### DIFF
--- a/crates/cairo-lang-formatter/src/bin/cli.rs
+++ b/crates/cairo-lang-formatter/src/bin/cli.rs
@@ -1,218 +1,9 @@
-use std::fs;
-use std::io::{stdin, Read};
-use std::path::{Path, PathBuf};
 use std::process::ExitCode;
-use std::sync::Arc;
 
-use anyhow::{bail, Result};
-use cairo_lang_filesystem::db::FilesGroup;
-use cairo_lang_filesystem::ids::{FileId, FileLongId, VirtualFile};
-use cairo_lang_formatter::{get_formatted_file, FormatterConfig};
-use cairo_lang_parser::utils::{get_syntax_root_and_diagnostics, SimpleParserDatabase};
+use cairo_lang_formatter::file_formatter::{format_path, format_stdin, FileFormatterArgs};
+use cairo_lang_formatter::FormatterConfig;
 use cairo_lang_utils::logging::init_logging;
 use clap::Parser;
-use colored::Colorize;
-use diffy::{create_patch, PatchFormatter};
-
-#[derive(Debug)]
-enum Input<'a> {
-    Stdin,
-    File { path: &'a str },
-}
-
-#[derive(Debug)]
-enum FormatResult {
-    Identical,
-    DiffFound,
-}
-
-impl<'a> Input<'a> {
-    pub fn to_file_id<D: FilesGroup>(&self, db: &D) -> Result<FileId, std::io::Error> {
-        match self {
-            Self::Stdin => {
-                let mut buffer = String::new();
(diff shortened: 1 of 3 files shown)


mkaput (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/1970)* status: all files reviewed, 2 unresolved discussions (waiting on @maciektr)

*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/1970#-NN0pHX4EEHC_8rkr7ih:-NN0pHX53Jh_APR3E_UR:b-tbjnrq) (no related file):*
I think we will want to more selectively publish functionality from the CLI:
1. Not everybody uses `eprint` for printing stuff to the user. For example, Scarb has the `Ui` system.
2. This code in current shape mixes various input kinds and formatter config with CLI-specific one.

What I think would be more usable is to provide the following API:

```rust
let config = FormatterConfig { .. };
let mut fmt = CairoFormatter::new(config);
// fmt.walk(path) returns preconfigured `ignore::WalkBuilder`,
// enabling use of its parallel variant if user wishes to
for path = fmt.walk("./").build() {
    // Input struct is converted to a trait and is implemented for &Path, &str and ZST `struct StdinFmt;`
    let result = fmt.check(&path)?;

    let result = fmt.format_in_place(&path)?;

    let (result, string) = fmt.format_to_string(&path)?;
}
```

what do you think?

FYI take a look at the https://docs.rs/ignore/latest/ignore/ crate, it implements much of the stuff going on here in a conventional way :smile: cc @gilbens-starkware By using this the notable difference would be that there would be no messages about symlinks or ignored paths, but I personally find this unnecessary noise.

___
___
*[`crates/cairo-lang-formatter/src/lib.rs` line 17 at r1](https://reviewable.io/reviews/starkware-libs/cairo/1970#-NN0uHhK9_6BUtgnVh_k:-NN0uHhK9_6BUtgnVh_l:b-3n2rwl) ([raw file](https://github.com/starkware-libs/cairo/blob/03ff18cf997968190e158f9fc82fb55dcfffd028/crates/cairo-lang-formatter/src/lib.rs#L17)):*
> ```XML
> use crate::formatter::Formatter;
> 
> pub mod file_formatter;
> ```

I think most of the stuff should be re-exported here?

gilbens-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/1970)* status: all files reviewed, 2 unresolved discussions (waiting on @maciektr and @mkaput)

*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/1970#-NN0pHX4EEHC_8rkr7ih:-NN14KWr7NxyOdE1kmxl:bg483gj) (no related file):*
<details><summary><i>Previously, maciektr (maciektr) wrote…</i></summary><blockquote>

@mkaput Thank you :raised_hands:! As of now existence of `cairofmtignore` file in a directory makes whole directory ignored. Should we change this behaviour to a solution based on filtering according to the contents of ignore files, or is it expected to be upheld with the new implementation?
</blockquote></details>

The behaviour of `cairofmtignore` should definitely be changed. It was a quick fix to ignore the formatter test files. Something similar to `.gitingnore` is preferable, and the `ignore` crate seems to give this exact solution.
As for the API, it looks good to me.

___

PR #2210, opened 2023-02-20
feat - Improving debug by  adding PrintTrait<T>

gaetbout:
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/2210)


gaetbout:
Maybe it is possible to enhance it by doing smthng like:

```
trait PrintTrait<T> {
    fn print(ref self: T);
}


impl FeltPrintImpl<felt> of PrintTrait::<felt> {
    fn print(ref self: felt){
        let mut arr = ArrayTrait::<felt>::new();
        arr.append(self);
        print(arr);
    }
}


impl ContractAddressPrintImpl<ContractAddress> of PrintTrait::<ContractAddress> {
    fn print(ref self: ContractAddress){
        self.into().print()
    }
}
```
Even though atm the trait would collide with 

`extern fn print(message: Array::<felt>) nopanic;`

But this would allow to have it for all types

15_u8.print();

spapinistarkware (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/2210)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @gaetbout)

___
*[`corelib/src/debug.cairo` line 14 at r2](https://reviewable.io/reviews/starkware-libs/cairo/2210#-NOx-RBG-EIQWYgTsSik:-NOx-RBG-EIQWYgTsSil:b-na5c1o) ([raw file](https://github.com/starkware-libs/cairo/blob/ca304a6629c93985f23917756586555ffcf0e544/corelib/src/debug.cairo#L14)):*
> ```Cairo
> }
> 
> fn print_contract_address(message: ContractAddress) {
> ```

I don't think that is the right approach. How about a Debug trait?

gaetbout:
I proposed such a Trait to be able to do 
`some_type.print()`
But this implementing such a trait would cause some issue with:
`extern fn print(message: Array::<felt>) nopanic;`
As it would collide

gaetbout:
Also trying to make the trait with a different name makes an issue:

`error: Unexpected argument type. Expected: "core::felt", found: "felt".` 
Here is the code:

```
use array::ArrayTrait;

extern fn print(message: Array::<felt>) nopanic;

fn print_felt(message: felt) {
    let mut arr = ArrayTrait::new();
    arr.append(message);
    print(arr);
}

trait PrintTrait<T> {
    fn print_it(self: T);
}

impl FeltPrintImpl<felt> of PrintTrait::<felt> {
    fn print_it(self: felt){
        print_felt(self);
    }
}
```

Is it linked on the fact that Trait on Generic type <T> isn't supported yet?
Or is it now?

gaetbout:
Oh nvm I found my mistake

gaetbout:
If CI passes, it should be fine now 👍

gaetbout:
Let me know if there is a way to specify like in rust that a generic type implements a trait
So that the 
`Array<felt> `
can be changed to
` Array<T: PrintTrait>`
And be made more generic

gaetbout:
I'd very much like to be able to pass a structure in the debugger to be able to specify some stuff such as:
 - Should show short string value  
 - Should show raw value
 - The type of the element to be printed
    - To be able to have stuff like:
    [DEBUG] felt:                           123123
    [DEBUG] ContractAddress:    654654
    [DEBUG] Array at 0:                54
    [DEBUG] Array at 1:                54
    [DEBUG] Array at 2:                54

But I don't know how to do that...
So being able to pass and Array of struct instead of an Array of felt would be very helpful. 
Could you point out on how to do this?
I wouldn't mind adding more debug possibilities :) 

Is it even relevant to do this or will it change?

gaetbout:
Also I notice that as_cairo_short_string always returns a String and can never return None()
So maybe it can be simplified?

spapinistarkware (approved):
:lgtm:

> Reviewed 1 of 2 files at r5, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/2210#-:-NPHIjFpCZjM93dw6wNN:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @gaetbout)

PR #2331, opened 2023-03-01
feat/ ArrayTrait: new_with

gaetbout:
Often in testing we have to create an array with one element, it would be easier if we could do it in a one liner instead of 
```
let mut data = ArrayTrait::new();
data.append('Out of gas');
panic(data);
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/2331)


spapinistarkware (commented):
I don't like this. Doesn't sound like the right thing to do. Sounds like something we might regret.
We don't have yet inline macros like rust either.
Until we have a good concrete design, let's wait with this.

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/2331#-:-NQsVk5ODOC2o863rY97:b-2oy7a6)* status: 0 of 2 files reviewed, all discussions resolved

PR #2412, opened 2023-03-08
Add array storage interface for smart contracts

punk5736:
Having arrays in storage is useful so I decided to try my hand at adding support in the compiler. Here are some thoughts I have about the implementation:

* The interface is effectively a Queue but with the added ability to index into the underlying array
* The memory layout scheme I implemented does not use a continuous segment of memory and consequently requires hashing on insert/read
* The implementation does not yet include any tests

Hopefully this PR can guide a discussion about what the proper interface and memory layout scheme should be.

- - -
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/2412)


punk5736:
Opened prematurely.

PR #2457, opened 2023-03-12
Add array concat

0xChqrles:
Adds generic array concat feature. Quite useful to create L1 <-> L2 message payloads, or ERC721 URI for example.

- - -
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/2457)

### DIFF
--- a/corelib/src/array.cairo
+++ b/corelib/src/array.cairo
@@ -128,3 +128,32 @@ fn clone_loop<T, impl TClone: Clone::<T>, impl TDrop: Drop::<T>>(
         Option::None(_) => {},
     }
 }
+
+trait ArrayTConcatTrait<T> {
+    fn concat(self: Array<T>, arr_2: @Array<T>) -> Array<T>;
+}
+
+impl ArrayTConcatImpl<T, impl TCopy: Copy::<T>> of ArrayTConcatTrait::<T> {
+    fn concat(mut self: Array<T>, arr_2: @Array<T>) -> Array<T> {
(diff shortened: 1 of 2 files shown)


spapinistarkware (commented):
There is another PR here: https://github.com/starkware-libs/cairo/pull/2493.
Perhaps it satisfies your needs?

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/2457#-:-NQsWFoa0Q9anyMZ7qff:b-fyt8yt)* status: 0 of 2 files reviewed, all discussions resolved

0xChqrles:
@spapinistarkware I'm not 100% satisfied by the implementation as we've discussed in #2493 but I'll close my PR if you would rather merge this one ^^

spapinistarkware:
Can you elaborate here please?

0xChqrles:
I think this method should be implemented using a new trait and not added to the `ArrayTrait` because of the unnecessary constraints it adds to the `T` type.

spapinistarkware:
But ArrayTrait adds no constraints on the T type!

0xChqrles:
Concatenation only works with array of types implementing Copy and Drop, thus I think it's better to use a new trait to implement such methods. But I may understand it wrong 😅

spapinistarkware:
You could add the constraints on the function itself though! No need to pit at on the impl.
Also, i dont think it actually requires that. It just needs ownership of the arrays.
Though, to support both cases we might need a new type. Some Owning Span...

0xChqrles:
Ok I understand better now. Thanks !

PR #2493, opened 2023-03-15
append_all in array

gaetbout:
Adding a way to append all the elements of an array to another array.
It'll consume the array passed as parameter

- - -
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/2493)

### DIFF
--- a/corelib/src/array.cairo
+++ b/corelib/src/array.cairo
@@ -12,7 +12,7 @@ extern fn array_len<T>(arr: @Array<T>) -> usize nopanic;
 trait ArrayTrait<T> {
     fn new() -> Array<T>;
     fn append(ref self: Array<T>, value: T);
-    fn append_all(ref self: Array::<T>, ref arr: Array::<T>);
+    fn append_all<T, impl TDrop: Drop::<T>>(ref self: Array::<T>, ref arr: Array::<T>);
     fn pop_front(ref self: Array<T>) -> Option<T>;
     fn get(self: @Array<T>, index: usize) -> Option<@T>;
     fn at(self: @Array<T>, index: usize) -> @T;
@@ -20,7 +20,7 @@ trait ArrayTrait<T> {
     fn is_empty(self: @Array<T>) -> bool;
     fn span(self: @Array<T>) -> Span<T>;
 }
-impl ArrayImpl<T, impl TDrop: Drop::<T>> of ArrayTrait::<T> {
+impl ArrayImpl<T> of ArrayTrait::<T> {
     #[inline(always)]
     fn new() -> Array<T> {
         array_new()
@@ -29,7 +29,7 @@ impl ArrayImpl<T, impl TDrop: Drop::<T>> of ArrayTrait::<T> {
     fn append(ref self: Array<T>, value: T) {
         array_append(ref self, value)
     }
-    fn append_all(ref self: Array::<T>, ref arr: Array::<T>) {
+    fn append_all<T, impl TDrop: Drop::<T>>(ref self: Array::<T>, ref arr: Array::<T>) {
         match get_gas() {
             Option::Some(_) => {},
             Option::None(_) => {
--- a/corelib/src/array.cairo
+++ b/corelib/src/array.cairo
@@ -12,6 +12,7 @@ extern fn array_len<T>(arr: @Array<T>) -> usize nopanic;
 trait ArrayTrait<T> {
     fn new() -> Array<T>;
     fn append(ref self: Array<T>, value: T);
+    fn append_all(ref self: Array::<T>, ref arr: Array::<T>);
     fn pop_front(ref self: Array<T>) -> Option<T>;
     fn get(self: @Array<T>, index: usize) -> Option<@T>;
     fn at(self: @Array<T>, index: usize) -> @T;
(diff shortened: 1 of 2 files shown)


gaetbout:
For some reason I need to specify T as Implementing Drop

0xChqrles:
I came up with a similar proposition, #2457 if you want to check

gaetbout:
I'd make it part of the ArrayTrait as it is more relevant there, Wouldn't want to use array::ArrayTrait and anotherone?
But I agree it should definitely be there, and making it part of the corelib avoids a lot of people doing their own (and potential errors)...
Just unsure how to solve the `cargo test` issue

0xChqrles:
I didn't add it to the `ArrayTrait` because of the unnecessary constraints it adds to the `T` type 🤷‍♀️
But I also think that it should be in the corelib (or at least in quaireaux). It's a very common function + as I state in my PR I already had to use it to develop basic contracts features.

enitrat:
I would name it `extend` to keep it consistent with rust's method. Eventually with proper intellisense it won't matter but it makes more sense semantically.
Also why not pass a Span?

spapinistarkware (changes requested):

spapinistarkware (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/2493)* status: 0 of 2 files reviewed, 3 unresolved discussions (waiting on @gaetbout)

___
*[`corelib/src/array.cairo` line 23 at r1](https://reviewable.io/reviews/starkware-libs/cairo/2493#-NRSJNuODgwlNQ-aOKu8:-NRSJNuODgwlNQ-aOKu9:bs33b4b) ([raw file](https://github.com/starkware-libs/cairo/blob/c982e521194bcd339d40ff014c10c528849d534a/corelib/src/array.cairo#L23)):*
> ```Cairo
>     fn span(self: @Array<T>) -> Span<T>;
> }
> impl ArrayImpl<T, impl TDrop: Drop::<T>> of ArrayTrait::<T> {
> ```

This unecessarily restricts all of the functions here to need T to be droppable. It is better to only add this generic param to append_all.



_Code quote:_
```Cairo
 impl TDrop: Drop::<T
```
___
*[`corelib/src/array.cairo` line 23 at r1](https://reviewable.io/reviews/starkware-libs/cairo/2493#-NRSJZvi5bHjz9Da55LW:-NRSJZvi5bHjz9Da55LX:b-uienl9) ([raw file](https://github.com/starkware-libs/cairo/blob/c982e521194bcd339d40ff014c10c528849d534a/corelib/src/array.cairo#L23)):*
> ```Cairo
>     fn span(self: @Array<T>) -> Span<T>;
> }
> impl ArrayImpl<T, impl TDrop: Drop::<T>> of ArrayTrait::<T> {
> ```

Also, add a TODO to require T to be destructible instead of droppable. You can also do it now, though, you would need to implement ArrayDestruct (which we need anyway).

gaetbout:
@spapinistarkware 

> This unecessarily restricts all of the functions here to need T to be droppable. It is better to only add this generic param to append_all.

How should I do that?
Should I simply put this in another trait?

enitrat:
> How should I do that?
> Should I simply put this in another trait?

```
impl ArrayImpl<T> of ArrayTrait::<T> {
    fn append_all<T, impl TDrop: Drop::<T>>(ref self: Array<T>, ref other: Array<T>);
}

**Comment by gaetbout** [comment]
> > How should I do that?
> > Should I simply put this in another trait?
> 
> ```rust
> impl ArrayImpl<T> of ArrayTrait::<T> {
>     fn append_all<T, impl TDrop: Drop::<T>>(ref self: Array<T>, ref other: Array<T>);
> }
> ```

Tried but it tells me smthng like
error: Wrong number of generic arguments. Expected 0, found: 2

**Comment by gaetbout** [comment]
@spapinistarkware 
I updated the file but now for some reason the tests don't find the append_all fn anymore. I also have the same with another setup on quaireaux.

**Review by orizi — CHANGES_REQUESTED** [review]
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/2493)* status: 0 of 2 files reviewed, 4 unresolved discussions (waiting on @gaetbout and @spapinistarkware)

*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/2493#-NRlBv2O0lr2q0w4ofzG:-NRlBv2O0lr2q0w4ofzH:bkgxnsp) (no related file):*
i think this should be implemented with iterators - so this should be put on hold for now (later this can be extended to supporting anything with the trait `IntoIterator` so it would work for all.)

___

**Review by spapinistarkware — COMMENTED** [review]
It didnt work becuase you redefined T at append_all. It was essentialy a different T. This is the right siganture:
```
fn append_all<impl TDrop: Drop::<T>>(ref self: Array::<T>, ref arr: Array::<T>);
```

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/2493#-:-NRmj1th5JEJPKnofXzh:b7t77qj)* status: 0 of 2 files reviewed, 4 unresolved discussions (waiting on @gaetbout and @orizi)

spapinistarkware:
Closed until we have iterators.

PR #2533, opened 2023-03-17
Print impl for generic types.

enitrat:
Implements print for all types that implement the `Into::<T,felt252>` trait


spapinistarkware:
I don't think this is the right way.
I don't want to force anyone that has IntoFelt to have this simple print IMO, it makes sense to have a different function for each.

PR #2570, opened 2023-03-21
Debug use generic traits

gaetbout:
Tested with all of this:
```rust
use debug::Print;
use starknet::contract_address_const;

#[test]
#[available_gas(1000000)]
fn test_print() {
    1.print();
    true.print();
    'msg'.print();
    8_u8.print();
    16_u16.print();
    32_u32.print();
    64_u64.print();
    128_u128.print();
    (contract_address_const::<42>()).print();
    let x = as_u256(1_u128, 0_u128);
    x.print();

    let mut arr = ArrayTrait::new();
    arr.append('1234567890123456789012345678901');
    arr.append('Sca');
    arr.append('SomeVeryLongMessage');
    arr.print();

    let mut arr1 = ArrayTrait::new();
    arr1.append(true);
    arr1.append(true);
    arr1.append(false);
    arr1.print();

    let mut arr2 = ArrayTrait::new();
    arr2.append(1_u8);
    arr2.append(8_u8);
    arr2.append(32_u8);
    arr2.print();

    let mut arr3 = ArrayTrait::new();
    arr3.append((contract_address_const::<42>()));
    arr3.append((contract_address_const::<84>()));
    arr3.append((contract_address_const::<12>()));
    arr3.print();
}
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/2570)

### DIFF
--- a/corelib/src/debug.cairo
+++ b/corelib/src/debug.cairo
@@ -1,11 +1,9 @@
 use array::ArrayTrait;
 use traits::Into;
-use starknet::ContractAddressIntoFelt252;
-use option::Option;
 
 // Usage:
 //
-// use debug::PrintTrait;
+// use debug::Print;
 //
 // 1.print();
 //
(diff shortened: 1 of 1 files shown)


enitrat:
It was refused in #2533

gaetbout:
> It was refused in #2533

Thanks for the notice!
I was hoping we could have this until there is more Debug support, I'll close 😃

gaetbout:
Even though the part about printing an Array of T could be handy, I'll leave it open and wait for more info!

spapinistarkware:
Like I said there, I am not convinced this is the right way. I would like to have some better design for debug prints in mind. When we have a general understanding of this, then it makes sense to have some early approximations for it. You are welcome to start a discussion in the git for it:)

spapinistarkware:
Closing, as this will probably wait until we have inline macros. @tarrencev is working on it I think,

PR #2589, opened 2023-03-24
Push config options from `ConfigPlugin` state to `FilesGroup`

mkaput:
This brings two improvements:
1. It is now possible to access `cfg` options from other plugins.
2. It is possible to set `cfg` options without bothering with plugins.

While implementing this I have followed a way how flags were done.

PR #2590, opened 2023-03-24
Establish a convention that plugins do not take construction arguments

mkaput:
Instead, I suggest keeping all necessary information in databases, like I did with `cfg` in #2589.
This eases reasoning on plugin construction.
For example, I have now idea how could I gather construction parameters from Scarb in an elegant way.

Additionally, test runner now uses `get_default_plugins`, because it can so.

- - -
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/2590)

### DIFF
--- a/crates/cairo-lang-defs/src/test.rs
+++ b/crates/cairo-lang-defs/src/test.rs
@@ -34,9 +34,9 @@ impl Default for DatabaseForTesting {
         let mut res = Self {
             storage: Default::default(),
             plugins: vec![
-                Arc::new(FooToBarPlugin {}),
-                Arc::new(RemoveOrigPlugin {}),
-                Arc::new(DummyPlugin {}),
+                Arc::new(FooToBarPlugin),
+                Arc::new(RemoveOrigPlugin),
+                Arc::new(DummyPlugin),
             ],
         };
         init_files_group(&mut res);
@@ -219,7 +219,7 @@ impl GeneratedFileAuxData for DummyAuxData {
 }
 
 #[derive(Debug)]
-struct DummyPlugin {}
+struct DummyPlugin;
 impl MacroPlugin for DummyPlugin {
     fn generate_code(&self, db: &dyn SyntaxGroup, item_ast: ast::Item) -> PluginResult {
         match item_ast {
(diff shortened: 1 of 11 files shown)


orizi (approved):
:lgtm:

> Reviewed 10 of 10 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/2590#-:-NRRRGph7k55HWi3C2T5:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @mkaput)

spapinistarkware:
@mkaput This works for our own plugins, but what about user plugins? They cannot modify the DB. What if they want to have configuration to their plugins?

mkaput (commented):
@spapinistarkware I tried to find actual use cases for passing configuration upon plugin construction and I couldn't find any, that have probably better solution.

1. Plugins taking configuration options on struct construction are basically unusable from CLIs, unless some layer is put on top of that that reads these values from CLI-accessible means, either a file or environment variables.
2. The only real-world usage of plugins, with configuration, that I am aware of is Dojo (cc @tarrencev), which is loading configuration from a `world.toml` file, because their interface is a CLI, not Rust library. We discussed this privately, and the conclusion is that it would make a sense anyway to let the plugin lazily load `world.toml` into a `OnceCell` upon first `SemanticPlugin` call.
3. I'd like to highlight that I am not opposed to plugins having a state in struct -- for this I'd recommend defining a `new` function. Perhaps should I redo this PR to use `new` constructor?

Additionally, I'm thinking of putting another layer on top of this, called `PluginDefinition` and pass this to `RootDatabaseBuilder`. The reason behind is that I am looking for a way to isolate plugins in separate processes, so that I could eventually allow injecting custom plugins without having to compile compiler/fork Scarb. But I need to iterate on this, the fact that plugins have access to the database is making this pretty hard.

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/2590#-:-NRX3cfK4vlpZLc8UX_J:b-u6cgui)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @mkaput)

mkaput (commented):
Added `Default` derive and used it wherever applicable + I marked structs as `non_exhaustive` just in case.

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/2590#-:-NRX_FAi2jnux2QxM17c:b3jymli)* status: 2 of 11 files reviewed, all discussions resolved (waiting on @orizi and @spapinistarkware)

orizi (approved):
:lgtm:

> Reviewed 9 of 9 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/2590#-:-NRXaCjy6YDmq29IrQBH:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @spapinistarkware)

PR #2739, opened 2023-04-07
debug trait for u16 and u32

notV4l:
just wonder why its missing


spapinistarkware:
Approved another PR that does the same

PR #2948, opened 2023-04-27
Add Abi ContractAddress/ClassHashIntoDispatcher implementations.

maciejka:
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/2948)


spapinistarkware:
I'm not 100% sire about this, since in the not far future, dispatcher might contain another field (this is related to the new contract syntax suggestion, where we have impl_offset)

maciejka:
> I'm not 100% sire about this, since in the not far future, dispatcher might contain another field (this is related to the new contract syntax suggestion, where we have impl_offset)

So knowing ContractAddress will not be enough to interact with the contract?

Issue #3178, opened 2023-05-17
feat: crate keyword


taco-paco:
# Feature Request

**Describe the Feature Request**

Have `crate` keyword as in Rust.

**Describe Preferred Solution**

using `crate` instead of possibly long crate names + its more explicit that imports are from current package.

**Describe Alternatives**

**Related Code**

`crates/cairo-lang-semantic/src/resolve/mod.rs`

**Additional Context**

**If the feature request is approved, would you be willing to submit a PR?**
_(Help can be provided if you need assistance submitting a PR)_

- [X] Yes
- [ ] No

Jinwei1987:
does help is needed for this issue?

enitrat:
@orizi bumping this

orizi:
checking if this can happen without particular breaking changes.

PR #3180, opened 2023-05-18
Added support for crate keyword

taco-paco:
#3178

### DIFF
--- a/crates/cairo-lang-semantic/src/resolve/mod.rs
+++ b/crates/cairo-lang-semantic/src/resolve/mod.rs
@@ -37,6 +37,9 @@ use crate::{
     TypeId, TypeLongId,
 };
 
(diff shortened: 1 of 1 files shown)


taco-paco:
where shall I add "test" contract?

orizi (commented):
maybe we should even disallow the current default as well - so that this would be more similar to rust.
@spapinistarkware WDYT?

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3180#-:-NWfgqL6DhlWiEblvmi9:bg8c3yg)* status: 0 of 1 files reviewed, all discussions resolved

taco-paco:
> maybe we should even disallow the current default

Just curious what is 'current default'? Inside the crate one will be able to use only `crate` keyword?

PR #3231, opened 2023-05-23
dev: fully qualified args in StorageAccess derive

milancermak:
Currently, when trying to use the `storage_access::StorageAccess` derive, we have to also import `use starknet::StorageBaseAddress;` because the function args in `read_at_offset_internal` and `write_at_offset_internal` are not fully qualified. This PR fixes that.

- - -
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3231)


milancermak:
Possibly superseded by #3247

milancermak:
Fixed in #3259

PR #3276, opened 2023-05-30
only use env_logger  in binaries

tdelabro:
`env_logger` is used in the cairo-lang crates, a bit indiscriminately in libs and bin.

The [crate documentation](https://github.com/rust-cli/env_logger/tree/main#in-libraries) states that it should only be used in binaries. This PR makes sure it does.

The problem is that those binaries are part of the same crates as the lib they use, making them share the same `[dependencies]`. This is problematic because we don't want it imported at all while building the lib, but we want it when building the binary.

There is multiple ways to fix this, as explained in [this StackOverflow answer](https://stackoverflow.com/questions/35711044/how-can-i-specify-binary-only-dependencies). I chose the second option: defining a specific feature for this import and adding it as `required-features` in the `[[bin]]` section.

I did it because it is the less invasive solution for now. But my recommendation would be to separate entirely binaries from libs, by making them exist in a separate crate where we can have full control over their dependencies (it will be more handy when we start to support `no_std` builds).

Wdyt? Is it good enough for now? Do you prefer the standalone binary crate solutions? If so where should those crates exist, and how should they be named (`<name of the lib>-bin`? Something else?).

- - -
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3276)


tdelabro:
@spapinistarkware Following the exchange you had with @abdelhamidbakhta about `no_std` support in the rust cairo libraries, I started working on it.
This is the first thing that will require a change (`env_logger` not being `no_std` compatible).

spapinistarkware (approved):
:lgtm:

> Reviewed 9 of 9 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3276#-:-NWi4DBSAygqr6PutFgR:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @tdelabro)

spapinistarkware:
Looks good

spapinistarkware (changes requested):
> Reviewed 8 of 8 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3276)* status: all files reviewed, 1 unresolved discussion (waiting on @tdelabro)

___
*[`scripts/cairo_fmt.sh` line 3 at r2](https://reviewable.io/reviews/starkware-libs/cairo/3276#-NWjvFhp1NTOOFCIv-u6:-NWjvFhp1NTOOFCIv-u7:bmx6okz) ([raw file](https://github.com/starkware-libs/cairo/blob/bb6a341bdb51d1dc03af08688e89523d6f714def/scripts/cairo_fmt.sh#L3)):*
> ```Shell
> #!/bin/bash
> 
> cargo run --bin cairo-format --features="binary" -- --recursive "$@"
> ```

Do I need to specify this when I run? Why is that, I thought defining in the toml is enough.
I'd really like to avoid this.

mkaput (commented):
Personally, I believe extracting binaries to a separate crate is a way to go upfront here. I suspect this play with features will bring issues to compiler crates users :/

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3276#-:-NWlLdrmAwllryJ8qsa-:bkyg9p4)* status: all files reviewed, 1 unresolved discussion (waiting on @tdelabro)

spapinistarkware (commented):
@tdelabro Can you try to separate the binaries into different crate then?
Remember to add the new crate to the ci yml file.

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3276#-:-NWi4EiV6L0S-uNx3MNY:b7qgrsf)* status: all files reviewed, 1 unresolved discussion (waiting on @tdelabro)

spapinistarkware:
No need to rebase. I want @orizi to also tal, and then we'll merge.

orizi (approved):
> Reviewed 2 of 9 files at r1, 2 of 8 files at r2, 27 of 29 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3276)* status: all files reviewed, 1 unresolved discussion (waiting on @tdelabro)

orizi (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3276#-:-NWnbxBl7Wsn1Mlt21Oq:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @tdelabro)

tdelabro:
@spapinistarkware ok perfect

mkaput (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3276)* status: all files reviewed, 1 unresolved discussion (waiting on @tdelabro)

___
*[`crates/cairo-lang-language-server/src/lib.rs` line 72 at r3](https://reviewable.io/reviews/starkware-libs/cairo/3276#-NWqcOrb4xfzv9ZxAwZC:-NWqcOrb4xfzv9ZxAwZD:b-mcdxt4) ([raw file](https://github.com/starkware-libs/cairo/blob/4e0cf74084a5f8fccff3dbffca216cc5a4d07275/crates/cairo-lang-language-server/src/lib.rs#L72)):*
> ```XML
> 
> pub async fn serve_language_service() {
>     init_logging(log::LevelFilter::Warn);
> ```

@maciektr please take a look at this, I think we will have to adapt Scarb's LS to also include logging, perhaps via tracing which we use there?

PR #3294, opened 2023-06-02
Derive common traits for `ListSelector`

mkaput:
The most important is `Clone` and `Eq` which would quite useful in Scarb. Adding other traits for future proofing.

- - -
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3294)

### DIFF
--- a/crates/cairo-lang-starknet/src/allowed_libfuncs.rs
+++ b/crates/cairo-lang-starknet/src/allowed_libfuncs.rs
@@ -33,7 +33,7 @@ pub enum AllowedLibfuncsError {
 }
 
 /// A selector for the allowed libfunc list.
-#[derive(Default)]
+#[derive(Clone, Debug, Default, Eq, PartialEq)]
 pub enum ListSelector {
     /// A list with one of the predefined names.
     ListName(String),


orizi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3294)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @mkaput)

___
*[`-- commits` line 4 at r1](https://reviewable.io/reviews/starkware-libs/cairo/3294#-NX3hq9PAOOJugi17FD9):*
In what way is adding unneeded traits future proofing?
Especially equality when you may not want to rely on the default.

mkaput (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3294)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @orizi)

___
*[`-- commits` line 4 at r1](https://reviewable.io/reviews/starkware-libs/cairo/3294#-NX3hq9PAOOJugi17FD9):*

In what way is adding unneeded traits future proofing?
Especially equality when you may not want to rely on the default.
</blockquote></details>

This is future proofing, because although these traits are unused within the project, this struct is **a part of public API of the compiler** and who knows what use cases will people find.

And I do actually have a need to use Clone and Eq in Scarb: 
- Clone: https://github.com/software-mansion/scarb/blob/77eedb6fdc56799cc21d07c0b780bc956ffffc4e/scarb/src/compiler/compilers/starknet_contract.rs#L193
- Eq: https://github.com/software-mansion/scarb/blob/77eedb6fdc56799cc21d07c0b780bc956ffffc4e/scarb/src/compiler/compilers/starknet_contract.rs#L212-L223

mkaput (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3294)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @mkaput)

___
*[`-- commits` line 4 at r1](https://reviewable.io/reviews/starkware-libs/cairo/3294#-NX3hq9PAOOJugi17FD9):*

If you need the eq, you need the eq, but again, this is not future proofing, this is the exact opposite 
This is prevents future updates, this does not protect against them.
</blockquote></details>

I actually thought the opposite, just from a different perspective: user of the API :) This makes the type more usable in contexts writer of the API couldn't imagine.

But OK, I agree I could probably use different words. Sorry for misunderstanding

PR #3297, opened 2023-06-02
No std support

tdelabro:
This PR aims to make as much as the cairo-lang libs no_std compatible.

In order to keep coherence across the crates, I opted for the `cairo-lang-std` approach. 
I added a crate that will export a custom "std" lib, either from the rust "std" one or from the "core" one (+ hashbrown).

This approach reduces the amount of conditional compilation (`#[cfg(feature = "std")]` vs `#[cfg(not(feature = "std"))]` ) to be expressed in the subsequent crates.

The drawback is that we have to import all our basics from this `cairo-lang-std` crate, rather than using `core`. Regardless, I think that it's the best option for such a diverse workspace.

I also added an `ensure_no_std` crate. It imports the different libs that support no-std and compiles, with them as dependencies. If even a single symbol across the dependencies is provided by the `std` lib, the compilation will fail. It is to be used in the CI in order to make sure no regressions are introduced.
It has to be compile with:
```shell
cargo +nightly build --target wasm32-unknown-unknown
```
meaning it cannot be part of the `workspace`. I added a `.cargo/config.toml` and a `rust-toolchain.toml` so running `cargo build` will be enough.

In it's current state `cairo-lang-std`, `cairo-lang-utils` and `cairo-lang-casm` are compilable in no-std.

- - -
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3297)


spapinistarkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3297)* status: 0 of 42 files reviewed, 1 unresolved discussion (waiting on @tdelabro)

___
*[`crates/cairo-lang-project/Cargo.toml` line 11 at r2](https://reviewable.io/reviews/starkware-libs/cairo/3297#-NXFAnAZ0B_ENd6LA5yo:-NXFAnAZ0B_ENd6LA5yp:b-4m48pj) ([raw file](https://github.com/starkware-libs/cairo/blob/ff9a60eae3439fc72a9e79d43f31875ccc240a3e/crates/cairo-lang-project/Cargo.toml#L11)):*
> ```TOML
> [dependencies]
> cairo-lang-filesystem = { path = "../cairo-lang-filesystem", version = "1.1.0" }
> cairo-lang-utils = { path = "../cairo-lang-utils", default-features = false }
> ```

Why did u remove the version?

tdelabro:
> Why did you remove the version?

This answer describes what happens if you combine `path` and `version`. If we are using the same version as the `workspace` version, it is not useful at all. And it could lead to errors if not updated. When I believe we always want to use the file from our current workspace commit.
[in this answer](https://users.rust-lang.org/t/cargo-toml-override-includes/52124/2)

> Depending on git commits is probelamtci, and messes with resolution a lot of times. Why isn't a version enough?

`indexmap` has a problem with `no-std` in its `1.X.X` version, it is described [here](https://github.com/bluss/indexmap/issues/258). So we have to use code that is on version `2.0.0-pre`, it could be using `version` or the branch `main`. But they are not respecting `semantic versioning` yet for this version. So they could include breaking changes over time, that is why I preferred to fix on a commit that does what I need and is `no-std` compatible until they release a fixed `2.0.0` tag.

> Format this better?

Yes my toml formating tool does this. Which is not the case of most people. I have to change it

mkaput:
> `indexmap` has a problem with `no-std` in its `1.X.X` version. It is described [here](https://github.com/bluss/indexmap/issues/258). So we have to use code that is on version `2.0.0-pre`, it could be achieved using `version` or the branch `main`. But they are not respecting `semantic versioning` yet for this version (because it's still `-pre`). So they could include breaking changes over time, that is why I preferred to fix on a commit that does what I need and is `no-std` compatible until they release a fixed `2.0.0` tag.

The issue is that using git-based dependencies makes the compiler unpublishable to crates.io which many projects depend on :(

tdelabro:
> The issue is that using git-based dependencies makes the compiler unpublishable to crates.io which many projects depend on :(

The only solution I see is for `indexmap` to publish a tag `2.0.0-pre`. I will try to talk with them and understand what is the feasibility of this.

tdelabro:
On my demand indexmap released the 2.0.0 version

tdelabro:
@spapinistarkware wdyt?

spapinistarkware (changes requested):
> Reviewed 5 of 57 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3297)* status: 5 of 59 files reviewed, 4 unresolved discussions (waiting on @tdelabro)

___
*[`crates/cairo-lang-casm-contract-class/Cargo.toml` line 22 at r4](https://reviewable.io/reviews/starkware-libs/cairo/3297#-NYvinW7DqiLC31l5N7Q:-NYvinW7DqiLC31l5N7R:b1iv6eq) ([raw file](https://github.com/starkware-libs/cairo/blob/5dd74104dc841c166db89d39da99c912a347abf9/crates/cairo-lang-casm-contract-class/Cargo.toml#L22)):*
> ```TOML
>     "num-bigint/std",
>     "serde/std",
> ]
> ```

Add newlines at the end of files

spapinistarkware (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3297)* status: 5 of 59 files reviewed, 5 unresolved discussions (waiting on @tdelabro)

___
*[`crates/cairo-lang-casm-contract-class/Cargo.toml` line 2 at r4](https://reviewable.io/reviews/starkware-libs/cairo/3297#-NYvit-V9gghAWtUKPS3:-NYvit-V9gghAWtUKPS4:b-myxgv6) ([raw file](https://github.com/starkware-libs/cairo/blob/5dd74104dc841c166db89d39da99c912a347abf9/crates/cairo-lang-casm-contract-class/Cargo.toml#L2)):*
> ```TOML
> [package]
> name = "cairo-lang-casm-contract-class"
> ```

I wonder if this is the right name for this package.
We might want to add things in the future here, right? Ant iot won't be just for the contract class

tdelabro:
@spapinistarkware Yes I wasn't sure about the naming, but I don't know what will happen in the future for those crates.
We could always deprepate thoses and create new ones, with the correct name for their content.
Or do you prefer a more vague crate name, like `cairo-lang-casm-types` maybe ?

PR #3339, opened 2023-06-08
Split len and is_empty methods to a different trait

wraitii:
This splits the `len` and `is_empty` methods to a different trait (which I called `Container`, but that name is not something I'm particularly attached to).

This will make it easier to write generic algorithms, since we can know expect 'container' types that implement `Index` and `Container` instead of having to rely on specific types.

Unfortunately, there are still some awkward bits:

- Despite requiring Container as a super-trait of ArrayTrait and SpanTrait, the super-trait is not automatically put in scope for inference, and so needs to be manually imported. This is rather clunky and would break backwards-compatibility in a bad way. I am not sure off-hand how easy it would be to fix inference for this case.
- I'm not entirely sure why Span `len` / `is_empty` wasn't implemented for snapshots ?
- this means we cannot use the recently added `generate_trait` macro.

IMO `get` and `at` should be removed in favour of just the indexing trait.
At this point, all that remains is to split off `append` / `pop_front` / `pop_back` in various traits. I'm not sure it's a good idea to bundle these too much.

- - -
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3339)


spapinistarkware (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3339)* status: 0 of 31 files reviewed, 1 unresolved discussion (waiting on @wraitii)

___
*[`corelib/src/array.cairo` line 34 at r1](https://reviewable.io/reviews/starkware-libs/cairo/3339#-Na6vBFE3sineEgItTNx:-Na6vBFE3sineEgItTNy:bftr34c) ([raw file](https://github.com/starkware-libs/cairo/blob/2460ab991c64d97e9af5b132765ba795f00989e4/corelib/src/array.cairo#L34)):*
> ```Cairo
> }
> 
> impl ArrayImpl<T> of ArrayTrait<T, ArrayContainer<T>> {
> ```

I don't think there is a need to defined the dependency on ArrayContainer. Array implements Container, and it is known. We can bring back the generate_trait, right?

spapinistarkware (commented):
I dont think we need to define the super-trait here, as ArrayTrait is specific to Array.
len and is_empty can be chagned to take `@Span`.
get() returns a Box, while Index does't necessarily return a Box.
at() can probably be united with the index() method of IndexView.

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3339#-:-Na6vdfL4A6jux2gNx3s:bt9s911)* status: 0 of 31 files reviewed, 1 unresolved discussion (waiting on @wraitii)

PR #3457, opened 2023-06-20
feat: pub visibility

Jinwei1987:
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3457)


spapinistarkware (changes requested):
> Reviewed 8 of 11 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3457)* status: 8 of 11 files reviewed, 8 unresolved discussions (waiting on @Jinwei1987)

___
*[`crates/cairo-lang-defs/src/db.rs` line 30 at r1](https://reviewable.io/reviews/starkware-libs/cairo/3457#-NYSI_Wa6TfP9kgfBZeK:-NYSI_Wa6TfP9kgfBZeL:b7l326l) ([raw file](https://github.com/starkware-libs/cairo/blob/7769549807ee4b670d3302543ef2f3f551828960/crates/cairo-lang-defs/src/db.rs#L30)):*
> ```XML
> {
>     #[salsa::interned]
>     fn intern_visibility(&self, id: VisibilityLongId) -> VisibilityId;
> ```

I don't think you need this
___
*[`crates/cairo-lang-defs/src/ids.rs` line 414 at r1](https://reviewable.io/reviews/starkware-libs/cairo/3457#-NYSIaNE5OEPJEBSYYVZ:-NYSIaNE5OEPJEBSYYV_:b7l326l) ([raw file](https://github.com/starkware-libs/cairo/blob/7769549807ee4b670d3302543ef2f3f551828960/crates/cairo-lang-defs/src/ids.rs#L414)):*
> ```XML
> 
> define_language_element_id!(
>     VisibilityId,
> ```

I don't think you need this
___
*[`crates/cairo-lang-parser/src/parser.rs` line 480 at r1](https://reviewable.io/reviews/starkware-libs/cairo/3457#-NYSIhfRAs4Ko5e_4UV-:-NYSIhfRAs4Ko5e_4UV0:b-xgzcl1) ([raw file](https://github.com/starkware-libs/cairo/blob/7769549807ee4b670d3302543ef2f3f551828960/crates/cairo-lang-parser/src/parser.rs#L480)):*
> ```XML
>     fn parse_visibility(&mut self) -> VisibilityGreen {
>         if self.peek().kind != SyntaxKind::TerminalPub {
>             VisibilityDefault::new_green(self.db).into()
> ```

Make this an early return
___
*[`crates/cairo-lang-semantic/src/substitution.rs` line 187 at r1](https://reviewable.io/reviews/starkware-libs/cairo/3457#-NYSI7RS-91koBa4QFqQ:-NYSI7RS-91koBa4QFqR:b7l326l) ([raw file](https://github.com/starkware-libs/cairo/blob/7769549807ee4b670d3302543ef2f3f551828960/crates/cairo-lang-semantic/src/substitution.rs#L187)):*
> ```XML
>         $crate::prune_single!(__identitity_helper, MemberId, $($exclude)*);
>         $crate::prune_single!(__identitity_helper, LocalVarId, $($exclude)*);
>         $crate::prune_single!(__identitity_helper, VisibilityId, $($exclude)*);
> ```

I don't think you need this
___
*[`crates/cairo-lang-semantic/src/expr/inference.rs` line 11 at r1](https://reviewable.io/reviews/starkware-libs/cairo/3457#-NYSI9g4BoUH3phd_I3f:-NYSI9g4BoUH3phd_I3g:b7l326l) ([raw file](https://github.com/starkware-libs/cairo/blob/7769549807ee4b670d3302543ef2f3f551828960/crates/cairo-lang-semantic/src/expr/inference.rs#L11)):*
> ```XML
>     ConstantId, EnumId, ExternFunctionId, ExternTypeId, FreeFunctionId, GenericParamId,
>     ImplAliasId, ImplDefId, ImplFunctionId, LanguageElementId, LocalVarId, MemberId, ParamId,
>     StructId, TraitFunctionId, TraitId, VarId, VariantId, VisibilityId
> ```

I don't think you need this
___
*[`crates/cairo-lang-semantic/src/expr/inference/canonic.rs` line 4 at r1](https://reviewable.io/reviews/starkware-libs/cairo/3457#-NYSIA_A4n_6vFXXiO-Z:-NYSIA_A4n_6vFXXiO-_:b7l326l) ([raw file](https://github.com/starkware-libs/cairo/blob/7769549807ee4b670d3302543ef2f3f551828960/crates/cairo-lang-semantic/src/expr/inference/canonic.rs#L4)):*
> ```XML
>     EnumId, ExternFunctionId, ExternTypeId, FreeFunctionId, GenericParamId, ImplAliasId, ImplDefId,
>     ImplFunctionId, LocalVarId, MemberId, ParamId, StructId, TraitFunctionId, TraitId, VarId,
>     VariantId, VisibilityId
> ```

I don't think you need this
___
*[`crates/cairo-lang-semantic/src/items/structure.rs` line 123 at r1](https://reviewable.io/reviews/starkware-libs/cairo/3457#-NYSIDj8E1F4aVm5LpPB:-NYSIDj8E1F4aVm5LpPC:b-lqd2xv) ([raw file](https://github.com/starkware-libs/cairo/blob/7769549807ee4b670d3302543ef2f3f551828960/crates/cairo-lang-semantic/src/items/structure.rs#L123)):*
> ```XML
> pub struct Member {
>     pub id: MemberId,
>     pub vis: VisibilityId,
> ```

You can annotate with `#[dont_rewrite]`
___
*[`crates/cairo-lang-semantic/src/items/structure.rs` line 159 at r1](https://reviewable.io/reviews/starkware-libs/cairo/3457#-NYSIJI56ACVLs0N-w8a:-NYSIJI56ACVLs0N-w8b:b-hybula) ([raw file](https://github.com/starkware-libs/cairo/blob/7769549807ee4b670d3302543ef2f3f551828960/crates/cairo-lang-semantic/src/items/structure.rs#L159)):*
> ```XML
>             &member.type_clause(syntax_db).ty(syntax_db),
>         );
>         let vis = db.intern_visibility(
> ```

I don't think this needs to be interned. It's kind of small an not used on its own. Just make one type Visibility, and store it on each member.

spapinistarkware (commented):
> Reviewed 3 of 11 files at r1.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3457)* status: all files reviewed, 8 unresolved discussions (waiting on @Jinwei1987)

spapinistarkware (changes requested):
> Reviewed 7 of 9 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3457)* status: 12 of 14 files reviewed, 4 unresolved discussions (waiting on @Jinwei1987)

___
*[`crates/cairo-lang-semantic/src/items/visibilities.rs` line 8 at r2](https://reviewable.io/reviews/starkware-libs/cairo/3457#-NYT71t98ywe2HG5DvAZ:-NYT71t98ywe2HG5DvA_:b-oiugx8) ([raw file](https://github.com/starkware-libs/cairo/blob/a0d408cb59fb17e2852e21ee146080adc9db2fc7/crates/cairo-lang-semantic/src/items/visibilities.rs#L8)):*
> ```XML
> use crate::Visibility;
> 
> pub fn compute_visibility(
> ```

I don't understand this logic.
1. Why are u adding submodules?
2. I think the visibility struct shouldn't contain module ids. It's kinda memory expensive.
Instead, it's enough to only save the bool, and compute later ifbthay accesses are kegal 
Wdyt?
___
*[`crates/cairo-lang-semantic/src/items/visibilities.rs` line 18 at r2](https://reviewable.io/reviews/starkware-libs/cairo/3457#-NYT6zc7BE3p6-fojJeJ:-NYT6zc8Ep7lQtx5SS2a:bcm6lty) ([raw file](https://github.com/starkware-libs/cairo/blob/a0d408cb59fb17e2852e21ee146080adc9db2fc7/crates/cairo-lang-semantic/src/items/visibilities.rs#L18)):*
> ```XML
>             let mut nodes = vec![module_id.clone()];
>             while !nodes.is_empty() {
>                 if let Some(module_id) = nodes.pop() {
> ```

U can do while let Some(...)
___
*[`crates/cairo-lang-semantic/src/items/visibilities.rs` line 32 at r2](https://reviewable.io/reviews/starkware-libs/cairo/3457#-NYT6xGt3NF-me0052ke:-NYT6xGuEWR4lFF_FlXZ:b-kvff0n) ([raw file](https://github.com/starkware-libs/cairo/blob/a0d408cb59fb17e2852e21ee146080adc9db2fc7/crates/cairo-lang-semantic/src/items/visibilities.rs#L32)):*
> ```XML
>         }
>         ast::Visibility::Public(_) => Ok(Visibility { is_public: true, module_ids: vec![] })
>     }
> ```

End file with a newline.

Jinwei1987 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3457)* status: 12 of 14 files reviewed, 4 unresolved discussions (waiting on @spapinistarkware)

___
*[`crates/cairo-lang-semantic/src/items/visibilities.rs` line 8 at r2](https://reviewable.io/reviews/starkware-libs/cairo/3457#-NYT71t98ywe2HG5DvAZ:-NYV_QUo5lTCESzIUfpi:b-pyjirt) ([raw file](https://github.com/starkware-libs/cairo/blob/a0d408cb59fb17e2852e21ee146080adc9db2fc7/crates/cairo-lang-semantic/src/items/visibilities.rs#L8)):*
<details><summary><i>Previously, spapinistarkware (Shahar Papini) wrote…</i></summary><blockquote>

I don't understand this logic.
1. Why are u adding submodules?
2. I think the visibility struct shouldn't contain module ids. It's kinda memory expensive.
Instead, it's enough to only save the bool, and compute later ifbthay accesses are kegal 
Wdyt?
</blockquote></details>

For 1, inspired by rust, the default visibility is actually inherent. That is the submodule has the access to items defined in parent module. It’s a convenient if writing test in sub module.

For 2, I was thinking different implementations. Another way is to hook to db for incremental competition while the semantic model contains only some simple attributes like bool etc. 

All this is aimed at a more complex visibility e.g. pub pub crate ext. if we only want pub or private, of course things can be simplified.

By the way I have removed intern stuff after a second thought.

spapinistarkware (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3457)* status: 8 of 17 files reviewed, 5 unresolved discussions (waiting on @Jinwei1987)

___
*[`crates/cairo-lang-semantic/src/items/visibilities.rs` line 28 at r3](https://reviewable.io/reviews/starkware-libs/cairo/3457#-NYmy6Zg7OkMTYw8SE1u:-NYmy6Zg7OkMTYw8SE1v:b-t9sdtq) ([raw file](https://github.com/starkware-libs/cairo/blob/7877d68e58b42edcce7150425aee6ba5e27f5c18/crates/cairo-lang-semantic/src/items/visibilities.rs#L28)):*
> ```XML
>                     return Ok(true)
>                 }
>                 if !visited.contains(&current) {
> ```

1. HashSet seems like a better data structure for this.
2. No need for `visited`. parent() is guaranteed to not yield cycles.

spapinistarkware (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3457)* status: 8 of 17 files reviewed, 6 unresolved discussions (waiting on @Jinwei1987)

___
*[`crates/cairo-lang-semantic/src/semantic.rs` line 73 at r3](https://reviewable.io/reviews/starkware-libs/cairo/3457#-NYmyLlX3xpRItsvCwkQ:-NYmyLlX3xpRItsvCwkR:b-wnpyom) ([raw file](https://github.com/starkware-libs/cairo/blob/7877d68e58b42edcce7150425aee6ba5e27f5c18/crates/cairo-lang-semantic/src/semantic.rs#L73)):*
> ```XML
> pub enum Visibility {
>     Public,
>     SubModule(ModuleId),
> ```

I think 
___
*[`crates/cairo-lang-semantic/src/items/visibilities.rs` line 8 at r2](https://reviewable.io/reviews/starkware-libs/cairo/3457#-NYT71t98ywe2HG5DvAZ:-NYmyXrNChqi88xbGojo:b-5sw0p3) ([raw file](https://github.com/starkware-libs/cairo/blob/a0d408cb59fb17e2852e21ee146080adc9db2fc7/crates/cairo-lang-semantic/src/items/visibilities.rs#L8)):*
<details><summary><i>Previously, Jinwei1987 (Jinwei) wrote…</i></summary><blockquote>

For 1, inspired by rust, the default visibility is actually inherent. That is the submodule has the access to items defined in parent module. It’s a convenient if writing test in sub module.

For 2, I was thinking different implementations. Another way is to hook to db for incremental competition while the semantic model contains only some simple attributes like bool etc. 

All this is aimed at a more complex visibility e.g. pub pub crate ext. if we only want pub or private, of course things can be simplified.

By the way I have removed intern stuff after a second thought.
</blockquote></details>

I'm in favor of supporting mopre intricate visibility models, but I would like the semantic model of the item visibility to contain a non redundant local-representation.
So, I don't thing that Visiblity::Submodule() should contain a submodule id - that is redudant, since the item module_id is locally achievable.

spapinistarkware (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3457)* status: 8 of 17 files reviewed, 6 unresolved discussions (waiting on @Jinwei1987)

___
*[`crates/cairo-lang-semantic/src/items/visibilities.rs` line 8 at r2](https://reviewable.io/reviews/starkware-libs/cairo/3457#-NYT71t98ywe2HG5DvAZ:-NYmz09G0YKaBrapbDIa:b-rxy1ue) ([raw file](https://github.com/starkware-libs/cairo/blob/a0d408cb59fb17e2852e21ee146080adc9db2fc7/crates/cairo-lang-semantic/src/items/visibilities.rs#L8)):*
<details><summary><i>Previously, spapinistarkware (Shahar Papini) wrote…</i></summary><blockquote>

I'm in favor of supporting mopre intricate visibility models, but I would like the semantic model of the item visibility to contain a non redundant local-representation.
So, I don't thing that Visiblity::Submodule() should contain a submodule id - that is redudant, since the item module_id is locally achievable.

</blockquote></details>

(you can do this with `item_id.module_id()`)

spapinistarkware (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3457)* status: 8 of 17 files reviewed, 6 unresolved discussions (waiting on @Jinwei1987)

___
*[`crates/cairo-lang-semantic/src/semantic.rs` line 73 at r3](https://reviewable.io/reviews/starkware-libs/cairo/3457#-NYmyLlX3xpRItsvCwkQ:-NYmz7AH-pIUe9pa0R9S:b-cjauro) ([raw file](https://github.com/starkware-libs/cairo/blob/7877d68e58b42edcce7150425aee6ba5e27f5c18/crates/cairo-lang-semantic/src/semantic.rs#L73)):*
<details><summary><i>Previously, spapinistarkware (Shahar Papini) wrote…</i></summary><blockquote>

I think 
</blockquote></details>

I think (ModuleId) is unecessary here

Jinwei1987:
> > _[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3457)_ status: 8 of 17 files reviewed, 6 unresolved discussions (waiting on @Jinwei1987)
> 
> _[`crates/cairo-lang-semantic/src/semantic.rs` line 73 at r3](https://reviewable.io/reviews/starkware-libs/cairo/3457#-NYmyLlX3xpRItsvCwkQ:-NYmyLlX3xpRItsvCwkR:b-wnpyom) ([raw file](https://github.com/starkware-libs/cairo/blob/7877d68e58b42edcce7150425aee6ba5e27f5c18/crates/cairo-lang-semantic/src/semantic.rs#L73)):_
> 
> > ```
> > pub enum Visibility {
> >     Public,
> >     SubModule(ModuleId),
> > ```
> 
> I think
> 
> _[`crates/cairo-lang-semantic/src/items/visibilities.rs` line 8 at r2](https://reviewable.io/reviews/starkware-libs/cairo/3457#-NYT71t98ywe2HG5DvAZ:-NYmyXrNChqi88xbGojo:b-5sw0p3) ([raw file](https://github.com/starkware-libs/cairo/blob/a0d408cb59fb17e2852e21ee146080adc9db2fc7/crates/cairo-lang-semantic/src/items/visibilities.rs#L8)):_
> 
> Previously, Jinwei1987 (Jinwei) wrote…
> I'm in favor of supporting mopre intricate visibility models, but I would like the semantic model of the item visibility to contain a non redundant local-representation. So, I don't thing that Visiblity::Submodule() should contain a submodule id - that is redudant, since the item module_id is locally achievable.

sure, with a redundant module_id, it will save some trivial computation steps. if you prefer memory efficiency, I will change it.

Jinwei1987:
most of the computations have been implemented, except pub use and  pub impl. I think pub use will introduce some reexport logic. For pub impl, trait will search all possible impl, good thing is that module visible impl can be hidden, but it also means that somewhere there is some logic to filter impl by visibility.

Jinwei1987 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3457)* status: 8 of 28 files reviewed, 6 unresolved discussions (waiting on @spapinistarkware)

___
*[`crates/cairo-lang-parser/src/parser.rs` line 480 at r1](https://reviewable.io/reviews/starkware-libs/cairo/3457#-NYSIhfRAs4Ko5e_4UV-:-NYuL5rk8FXGOoHaDR3p:bskmyul) ([raw file](https://github.com/starkware-libs/cairo/blob/7769549807ee4b670d3302543ef2f3f551828960/crates/cairo-lang-parser/src/parser.rs#L480)):*
<details><summary><i>Previously, spapinistarkware (Shahar Papini) wrote…</i></summary><blockquote>

Make this an early return
</blockquote></details>

not sure what do you mean "early return". do you mean use return keyword instead of using if result for return?
___
*[`crates/cairo-lang-semantic/src/semantic.rs` line 73 at r3](https://reviewable.io/reviews/starkware-libs/cairo/3457#-NYmyLlX3xpRItsvCwkQ:-NYuLUpAAagQQYG9zacf:b-896fix) ([raw file](https://github.com/starkware-libs/cairo/blob/7877d68e58b42edcce7150425aee6ba5e27f5c18/crates/cairo-lang-semantic/src/semantic.rs#L73)):*
<details><summary><i>Previously, spapinistarkware (Shahar Papini) wrote…</i></summary><blockquote>

I think (ModuleId) is unecessary here
</blockquote></details>

Done.
___
*[`crates/cairo-lang-semantic/src/items/visibilities.rs` line 18 at r2](https://reviewable.io/reviews/starkware-libs/cairo/3457#-NYT6zc7BE3p6-fojJeJ:-NYuLrgC4eaD7WVLe18i:b-896fix) ([raw file](https://github.com/starkware-libs/cairo/blob/a0d408cb59fb17e2852e21ee146080adc9db2fc7/crates/cairo-lang-semantic/src/items/visibilities.rs#L18)):*
<details><summary><i>Previously, spapinistarkware (Shahar Papini) wrote…</i></summary><blockquote>

U can do while let Some(...)
</blockquote></details>

Done.
___
*[`crates/cairo-lang-semantic/src/items/visibilities.rs` line 32 at r2](https://reviewable.io/reviews/starkware-libs/cairo/3457#-NYT6xGt3NF-me0052ke:-NYuMg-82jsj6eMV9Rxf:b-896fix) ([raw file](https://github.com/starkware-libs/cairo/blob/a0d408cb59fb17e2852e21ee146080adc9db2fc7/crates/cairo-lang-semantic/src/items/visibilities.rs#L32)):*
<details><summary><i>Previously, spapinistarkware (Shahar Papini) wrote…</i></summary><blockquote>

End file with a newline.
</blockquote></details>

Done.
___
*[`crates/cairo-lang-semantic/src/items/visibilities.rs` line 28 at r3](https://reviewable.io/reviews/starkware-libs/cairo/3457#-NYmy6Zg7OkMTYw8SE1u:-NYuLtWvDkIgpdBjyS31:b-896fix) ([raw file](https://github.com/starkware-libs/cairo/blob/7877d68e58b42edcce7150425aee6ba5e27f5c18/crates/cairo-lang-semantic/src/items/visibilities.rs#L28)):*
<details><summary><i>Previously, spapinistarkware (Shahar Papini) wrote…</i></summary><blockquote>

1. HashSet seems like a better data structure for this.
2. No need for `visited`. parent() is guaranteed to not yield cycles.
</blockquote></details>

Done.

spapinistarkware (changes requested):
Please base this PR on the cairo-major-update branch

> Reviewed 10 of 17 files at r4, 2 of 6 files at r5.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3457#-:-NYvNM1D5JKbL1GT0ewm:b-rjwwae)* status: 20 of 28 files reviewed, 4 unresolved discussions (waiting on @Jinwei1987)

___
*[`crates/cairo-lang-parser/src/parser.rs` line 480 at r1](https://reviewable.io/reviews/starkware-libs/cairo/3457#-NYSIhfRAs4Ko5e_4UV-:-NYvMaWE4guZD7UkMKeU:b-e2qju7) ([raw file](https://github.com/starkware-libs/cairo/blob/7769549807ee4b670d3302543ef2f3f551828960/crates/cairo-lang-parser/src/parser.rs#L480)):*
<details><summary><i>Previously, Jinwei1987 (Jinwei) wrote…</i></summary><blockquote>

not sure what do you mean "early return". do you mean use return keyword instead of using if result for return?
</blockquote></details>

Yes.
```
if self.peek().kind != SyntaxKind::TerminalPub {
  return ...;
}
let pub_kw = ...;
```

spapinistarkware (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3457)* status: 20 of 28 files reviewed, 5 unresolved discussions (waiting on @Jinwei1987)

___
*[`crates/cairo-lang-semantic/src/items/visibilities.rs` line 15 at r5](https://reviewable.io/reviews/starkware-libs/cairo/3457#-NYvNWsl2MRDo7gDlctz:-NYvNWsl2MRDo7gDlcu-:b-pej8w4) ([raw file](https://github.com/starkware-libs/cairo/blob/7be4a797734c97411aec6d2102e31545599cea8d/crates/cairo-lang-semantic/src/items/visibilities.rs#L15)):*
> ```XML
> }
> 
> /// Query implementation of [SemanticGroup::constant_visible_in].
> ```

Don't make each a separate query.
They won't be able to reuse each other's query cache.
Instead, consider just making a single query for somehting like privates_visible_in(module_a, module_b).
Or maybe call it is_submodule.

Also, instead of having a lot fo free functions here, extend the lookup_item.rs module to incorporate these as methods.

For example, extend `HasResolverData` with a visibility() function and a visible_in() function.

spapinistarkware (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3457)* status: 20 of 28 files reviewed, 3 unresolved discussions (waiting on @Jinwei1987)

___
*[`crates/cairo-lang-semantic/src/items/visibilities.rs` line 101 at r5](https://reviewable.io/reviews/starkware-libs/cairo/3457#-NYvOkZJ7eWpaAqjon9h:-NYvOkZJ7eWpaAqjon9i:budyjrj) ([raw file](https://github.com/starkware-libs/cairo/blob/7be4a797734c97411aec6d2102e31545599cea8d/crates/cairo-lang-semantic/src/items/visibilities.rs#L101)):*
> ```XML
>                 }
>             }
>             Ok(false)
> ```

like I wrote above, I suggest extracting this to a query.
Actually, it would be even better to make a query like module_ancesstors(ModuleId)->OrderedHashSet<ModuleId>, for memory and lookup efficiency.



_Code quote:_
```XML
            let mut nodes = vec![source_module_id];
            while let Some(current) = nodes.pop() {
                if module_id == current {
                    return Ok(true)
                }
                for submodule_id in db.module_submodules_ids(current.clone())? {
                    nodes.push(ModuleId::Submodule(submodule_id));
                }
            }
            Ok(false)
```

spapinistarkware (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3457)* status: 20 of 28 files reviewed, 3 unresolved discussions (waiting on @Jinwei1987)

___
*[`crates/cairo-lang-semantic/src/items/visibilities.rs` line 101 at r5](https://reviewable.io/reviews/starkware-libs/cairo/3457#-NYvOkZJ7eWpaAqjon9h:-NYvP-yD8SM5rt8cyRUe:bgu1f1l) ([raw file](https://github.com/starkware-libs/cairo/blob/7be4a797734c97411aec6d2102e31545599cea8d/crates/cairo-lang-semantic/src/items/visibilities.rs#L101)):*
<details><summary><i>Previously, spapinistarkware (Shahar Papini) wrote…</i></summary><blockquote>

like I wrote above, I suggest extracting this to a query.
Actually, it would be even better to make a query like module_ancesstors(ModuleId)->OrderedHashSet<ModuleId>, for memory and lookup efficiency.
</blockquote></details>

Either this, or not make any query at all for the visible_in() flow.

Jinwei1987 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3457)* status: 20 of 28 files reviewed, 3 unresolved discussions (waiting on @spapinistarkware)

___
*[`crates/cairo-lang-semantic/src/items/visibilities.rs` line 15 at r5](https://reviewable.io/reviews/starkware-libs/cairo/3457#-NYvNWsl2MRDo7gDlctz:-NZ-AGMb2_YLMhJsWMcV:b-i2e8f9) ([raw file](https://github.com/starkware-libs/cairo/blob/7be4a797734c97411aec6d2102e31545599cea8d/crates/cairo-lang-semantic/src/items/visibilities.rs#L15)):*
<details><summary><i>Previously, spapinistarkware (Shahar Papini) wrote…</i></summary><blockquote>

Don't make each a separate query.
They won't be able to reuse each other's query cache.
Instead, consider just making a single query for somehting like privates_visible_in(module_a, module_b).
Or maybe call it is_submodule.

Also, instead of having a lot fo free functions here, extend the lookup_item.rs module to incorporate these as methods.

For example, extend `HasResolverData` with a visibility() function and a visible_in() function.
</blockquote></details>

Found a more serious problem.
Currently priv_* -> inference.finalize -> core* -> priv_use -> resolver.
By any mean if we introduce resolver -> priv_*, we will introduce a cyclic call.
Put visibility to ResolveData won't fix this problem.

There are two solutions:
1) Attach the visibility to the LongId like module_file_id. intern happens in defs group before resolve. So resolver don't need to call priv_* to get visibility. Visibility is a simple enum, so it won't take up a lot memory in the cache.
2) Fix inference.finalize. Somehow let inference knows felt252 has been loaded.

I prefer solution 1. It's better maintain single direction call. Bi-direction call will be a nightmare in the future.

spapinistarkware (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3457)* status: 20 of 28 files reviewed, 3 unresolved discussions (waiting on @Jinwei1987)

___
*[`crates/cairo-lang-semantic/src/items/visibilities.rs` line 15 at r5](https://reviewable.io/reviews/starkware-libs/cairo/3457#-NYvNWsl2MRDo7gDlctz:-NZ0cvwf8sDphT_q9H23:bd06jzg) ([raw file](https://github.com/starkware-libs/cairo/blob/7be4a797734c97411aec6d2102e31545599cea8d/crates/cairo-lang-semantic/src/items/visibilities.rs#L15)):*
<details><summary><i>Previously, Jinwei1987 (Jinwei) wrote…</i></summary><blockquote>

Found a more serious problem.
Currently priv_* -> inference.finalize -> core* -> priv_use -> resolver.
By any mean if we introduce resolver -> priv_*, we will introduce a cyclic call.
Put visibility to ResolveData won't fix this problem.

There are two solutions:
1) Attach the visibility to the LongId like module_file_id. intern happens in defs group before resolve. So resolver don't need to call priv_* to get visibility. Visibility is a simple enum, so it won't take up a lot memory in the cache.
2) Fix inference.finalize. Somehow let inference knows felt252 has been loaded.

I prefer solution 1. It's better maintain single direction call. Bi-direction call will be a nightmare in the future.


</blockquote></details>

Yeah, the essence of this issue it that resolver needs to not know the semantic models of the items it resolves, onyl their existance.
So we need the visibility to be attained before the semantic model of the item.
I think it can be a query that gets the item id, and the implementation is fetching the ast and retrieving the visibility.

Jinwei1987 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3457)* status: 17 of 30 files reviewed, 3 unresolved discussions (waiting on @spapinistarkware)

___
*[`crates/cairo-lang-semantic/src/items/visibilities.rs` line 15 at r5](https://reviewable.io/reviews/starkware-libs/cairo/3457#-NYvNWsl2MRDo7gDlctz:-NZ0lsduDNWFMQo-0uc7:b-utc9mw) ([raw file](https://github.com/starkware-libs/cairo/blob/7be4a797734c97411aec6d2102e31545599cea8d/crates/cairo-lang-semantic/src/items/visibilities.rs#L15)):*
<details><summary><i>Previously, spapinistarkware (Shahar Papini) wrote…</i></summary><blockquote>

Yeah, the essence of this issue it that resolver needs to not know the semantic models of the items it resolves, onyl their existance.
So we need the visibility to be attained before the semantic model of the item.
I think it can be a query that gets the item id, and the implementation is fetching the ast and retrieving the visibility.
</blockquote></details>

I have some implementation to attach visibility to id, e.g. SubmoduleId. If you are happy with this solution, I will go forward. Another way we can add a bunch of free functions to query ast by item in defs group. But I still think attach to id is cleaner.

spapinistarkware (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3457)* status: 17 of 30 files reviewed, 3 unresolved discussions (waiting on @Jinwei1987)

___
*[`crates/cairo-lang-semantic/src/items/visibilities.rs` line 15 at r5](https://reviewable.io/reviews/starkware-libs/cairo/3457#-NYvNWsl2MRDo7gDlctz:-NZ1BzJd0K-kGCBas1YU:b8wb7l) ([raw file](https://github.com/starkware-libs/cairo/blob/7be4a797734c97411aec6d2102e31545599cea8d/crates/cairo-lang-semantic/src/items/visibilities.rs#L15)):*
<details><summary><i>Previously, Jinwei1987 (Jinwei) wrote…</i></summary><blockquote>

I have some implementation to attach visibility to id, e.g. SubmoduleId. If you are happy with this solution, I will go forward. Another way we can add a bunch of free functions to query ast by item in defs group. But I still think attach to id is cleaner.
</blockquote></details>

WDYM by attach? It musn't be a part of SubmoduleId itself, that would affect the id when the visibility is changed, and will make it hard to construct these ids outside.

Jinwei1987 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3457)* status: 17 of 30 files reviewed, 3 unresolved discussions (waiting on @spapinistarkware)

___
*[`crates/cairo-lang-semantic/src/items/visibilities.rs` line 15 at r5](https://reviewable.io/reviews/starkware-libs/cairo/3457#-NYvNWsl2MRDo7gDlctz:-NZ1P4v8CDdPpqWylaC7:b-okj045) ([raw file](https://github.com/starkware-libs/cairo/blob/7be4a797734c97411aec6d2102e31545599cea8d/crates/cairo-lang-semantic/src/items/visibilities.rs#L15)):*
<details><summary><i>Previously, spapinistarkware (Shahar Papini) wrote…</i></summary><blockquote>

WDYM by attach? It musn't be a part of SubmoduleId itself, that would affect the id when the visibility is changed, and will make it hard to construct these ids outside.
</blockquote></details>

Sure, I see what you mean.  Will change accordingly. Another as far as I understand now the resolver will use ast visibility to do the computation. Do we still need to keep a semantic visibility in the semantic models? Don’t see any use of the semantic visibility.

Jinwei1987 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3457)* status: 5 of 68 files reviewed, 3 unresolved discussions (waiting on @spapinistarkware)

___
*[`crates/cairo-lang-parser/src/parser.rs` line 480 at r1](https://reviewable.io/reviews/starkware-libs/cairo/3457#-NYSIhfRAs4Ko5e_4UV-:-NZ4KeTe4RAt4AAX4qxL:b-896fix) ([raw file](https://github.com/starkware-libs/cairo/blob/7769549807ee4b670d3302543ef2f3f551828960/crates/cairo-lang-parser/src/parser.rs#L480)):*
<details><summary><i>Previously, spapinistarkware (Shahar Papini) wrote…</i></summary><blockquote>

Yes.
```
if self.peek().kind != SyntaxKind::TerminalPub {
  return ...;
}
let pub_kw = ...;
```
</blockquote></details>

Done.

Jinwei1987 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3457)* status: 5 of 69 files reviewed, 3 unresolved discussions (waiting on @spapinistarkware)

___
*[`crates/cairo-lang-semantic/src/items/visibilities.rs` line 101 at r5](https://reviewable.io/reviews/starkware-libs/cairo/3457#-NYvOkZJ7eWpaAqjon9h:-NZ4cJyF79XDSqF_52jZ:b-9yxy82) ([raw file](https://github.com/starkware-libs/cairo/blob/7be4a797734c97411aec6d2102e31545599cea8d/crates/cairo-lang-semantic/src/items/visibilities.rs#L101)):*
<details><summary><i>Previously, spapinistarkware (Shahar Papini) wrote…</i></summary><blockquote>

Either this, or not make any query at all for the visible_in() flow.
</blockquote></details>

I think pub visibility is working now. I have implemented in the way as suggested I think.

Jinwei1987 (commented):
Done

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3457#-:-NZ4klrx23ebmpS3m16V:bb6un1j)* status: 4 of 69 files reviewed, 3 unresolved discussions (waiting on @spapinistarkware)

spapinistarkware (changes requested):
> Reviewed 2 of 5 files at r6, 4 of 55 files at r7, 6 of 26 files at r8.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3457)* status: 16 of 71 files reviewed, 8 unresolved discussions (waiting on @Jinwei1987)

___
*[`crates/cairo-lang-defs/src/db.rs` line 668 at r8](https://reviewable.io/reviews/starkware-libs/cairo/3457#-NZ4v36uAUl4-edrlQpt:-NZ4v36uAUl4-edrlQpu:bin52e1) ([raw file](https://github.com/starkware-libs/cairo/blob/660e5ab76059041f80e67240545b4e7c9bc08841/crates/cairo-lang-defs/src/db.rs#L668)):*
> ```XML
> }
> 
> pub fn struct_member_visibility(db: &dyn DefsGroup, struct_id: StructId, member_name: SmolStr) -> Maybe<Option<ast::Visibility>> {
> ```

You should make this query get MemberId, and fetch the ast of the member from it.
___
*[`crates/cairo-lang-parser/src/parser.rs` line 477 at r8](https://reviewable.io/reviews/starkware-libs/cairo/3457#-NZ4vUeD8x8S2mcWJUT6:-NZ4vUeE55OD2dM36qdJ:bgr59ql) ([raw file](https://github.com/starkware-libs/cairo/blob/660e5ab76059041f80e67240545b4e7c9bc08841/crates/cairo-lang-parser/src/parser.rs#L477)):*
> ```XML
>         if self.peek().kind != SyntaxKind::TerminalPub {
>             return VisibilityDefault::new_green(self.db).into()
>         } else {
> ```

No need for the `else` now. Unindent it.
___
*[`crates/cairo-lang-semantic/src/lookup_item.rs` line 17 at r8](https://reviewable.io/reviews/starkware-libs/cairo/3457#-NZ50QAz271rGYg8VMWT:-NZ50QAz271rGYg8VMWU:b-nrlr19) ([raw file](https://github.com/starkware-libs/cairo/blob/660e5ab76059041f80e67240545b4e7c9bc08841/crates/cairo-lang-semantic/src/lookup_item.rs#L17)):*
> ```XML
>     fn resolver_data(&self, db: &dyn SemanticGroup) -> Maybe<Arc<ResolverData>>;
>     
>     fn visible_in(&self, db: &dyn SemanticGroup, module_id: ModuleId) -> Maybe<bool>;
> ```

We should somehow document that this is not the entire visibility logic.
i.e. when we import `a::b::c`, c might be pub (and thus, visible_in will be true), but b might not be visible.
so c is still inaccessible.

Can you try to document this? Maybe we should also rename `visible` to reflect this? Maybe something like visible_given_parent, or something... WDYT?
___
*[`crates/cairo-lang-semantic/src/lookup_item.rs` line 119 at r8](https://reviewable.io/reviews/starkware-libs/cairo/3457#-NZ4vyxy-JGC8ERZ-f2k:-NZ4vyxy-JGC8ERZ-f2l:b-kbg7p8) ([raw file](https://github.com/starkware-libs/cairo/blob/660e5ab76059041f80e67240545b4e7c9bc08841/crates/cairo-lang-semantic/src/lookup_item.rs#L119)):*
> ```XML
>     }
> 
>     fn visible_in(&self, _db: &dyn SemanticGroup, _module_id: ModuleId) -> Maybe<bool> {
> ```

why are these 3 returning true, instead of visible_in()?
___
*[`crates/cairo-lang-semantic/src/expr/compute.rs` line 1962 at r8](https://reviewable.io/reviews/starkware-libs/cairo/3457#-NZ524_k5ApKmzpz8oHm:-NZ524_k5ApKmzpz8oHn:b-48bdtu) ([raw file](https://github.com/starkware-libs/cairo/blob/660e5ab76059041f80e67240545b4e7c9bc08841/crates/cairo-lang-semantic/src/expr/compute.rs#L1962)):*
> ```XML
> ) -> Maybe<()> {
>     let source_module_id = struct_id.parent_module(ctx.db.upcast());
>     if let Some(visibility) = ctx.db.struct_member_visibility(struct_id, member_name.clone())? {
> ```

Rephrase this with early returns.
```
let Some(visibility) = .. else { return; };
if visible_in(..) { return; }
...
```
___
*[`crates/cairo-lang-semantic/src/expr/compute.rs` line 1964 at r8](https://reviewable.io/reviews/starkware-libs/cairo/3457#-NZ51wWO20WRcuN2hR5V:-NZ51wWO20WRcuN2hR5W:b-mrphse) ([raw file](https://github.com/starkware-libs/cairo/blob/660e5ab76059041f80e67240545b4e7c9bc08841/crates/cairo-lang-semantic/src/expr/compute.rs#L1964)):*
> ```XML
>     if let Some(visibility) = ctx.db.struct_member_visibility(struct_id, member_name.clone())? {
>         if !visibilities::visible_in(ctx.db, &visibility, source_module_id, module_id) {
>             return Err(ctx.diagnostics.report(
> ```

It's better to report the diagnostics, but not fail the computation. This way, we stay with more relevant semantic information.

spapinistarkware (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3457)* status: 16 of 71 files reviewed, 9 unresolved discussions (waiting on @Jinwei1987)

___
*[`crates/cairo-lang-semantic/src/lookup_item.rs` line 14 at r8](https://reviewable.io/reviews/starkware-libs/cairo/3457#-NZ53Elq3ddi4SQwpI1n:-NZ53Elq3ddi4SQwpI1o:bl1fulr) ([raw file](https://github.com/starkware-libs/cairo/blob/660e5ab76059041f80e67240545b4e7c9bc08841/crates/cairo-lang-semantic/src/lookup_item.rs#L14)):*
> ```XML
> use crate::resolve::ResolverData;
> 
> pub trait HasResolverData {
> ```

Let's rename this.
Maybe `ItemEx`?



_Code quote:_
```XML
asResolverData
```
___
*[`crates/cairo-lang-semantic/src/items/visibilities.rs` line 5 at r8](https://reviewable.io/reviews/starkware-libs/cairo/3457#-NZ52QpE14GkCKmXRMqW:-NZ52QpE14GkCKmXRMqX:b-a73bxd) ([raw file](https://github.com/starkware-libs/cairo/blob/660e5ab76059041f80e67240545b4e7c9bc08841/crates/cairo-lang-semantic/src/items/visibilities.rs#L5)):*
> ```XML
> use crate::db::SemanticGroup;
> 
> pub fn visible_in(db: &dyn SemanticGroup, visibility: &ast::Visibility, source_module_id: ModuleId, module_id: ModuleId) -> bool {
> ```

Same documentation here as what I wrote in lookup_item.rs. Also, make the other doc point here.
___
*[`crates/cairo-lang-semantic/src/resolve/mod.rs` line 907 at r8](https://reviewable.io/reviews/starkware-libs/cairo/3457#-NZ53SrZBQ9MKnpNAeyg:-NZ53SrZBQ9MKnpNAeyh:b-kbwc9p) ([raw file](https://github.com/starkware-libs/cairo/blob/660e5ab76059041f80e67240545b4e7c9bc08841/crates/cairo-lang-semantic/src/resolve/mod.rs#L907)):*
> ```XML
>     }
> 
>     define_check_is_visible!(check_constant_is_visible, constant_id, &ConstantId, ConstantNotVisible);
> ```

Do we need all of this? Can't we use the HasResolver trait and its visible_in()?

spapinistarkware (changes requested):
> Reviewed 1 of 26 files at r8.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3457)* status: 17 of 71 files reviewed, 10 unresolved discussions (waiting on @Jinwei1987)

___
*[`corelib/src/box.cairo` line 2 at r8](https://reviewable.io/reviews/starkware-libs/cairo/3457#-NZ53sv15n1qjU1cI9wu:-NZ53sv15n1qjU1cI9wv:b-fwo46y) ([raw file](https://github.com/starkware-libs/cairo/blob/660e5ab76059041f80e67240545b4e7c9bc08841/corelib/src/box.cairo#L2)):*
> ```Cairo
> pub extern type Box<T>;
> impl BoxTCopy<T, impl TCopy: Copy<T>> of Copy<Box<T>>;
> ```

We should think of how to introduce visibility on impls.
Please add a TODO(spapini) for this somewhere.
___
*[`crates/cairo-lang-semantic/src/resolve/mod.rs` line 907 at r8](https://reviewable.io/reviews/starkware-libs/cairo/3457#-NZ53SrZBQ9MKnpNAeyg:-NZ55Fb53xerY36w7uQ_:bgoz2vo) ([raw file](https://github.com/starkware-libs/cairo/blob/660e5ab76059041f80e67240545b4e7c9bc08841/crates/cairo-lang-semantic/src/resolve/mod.rs#L907)):*
<details><summary><i>Previously, spapinistarkware (Shahar Papini) wrote…</i></summary><blockquote>

Do we need all of this? Can't we use the HasResolver trait and its visible_in()?
</blockquote></details>

Also, I tihnk we don't want a different diagnostic kind for each. Make one diagnostic kind with a LookupItemId.

Jinwei1987 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3457)* status: 13 of 72 files reviewed, 10 unresolved discussions (waiting on @spapinistarkware)

___
*[`corelib/src/box.cairo` line 2 at r8](https://reviewable.io/reviews/starkware-libs/cairo/3457#-NZ53sv15n1qjU1cI9wu:-NZ96f7A17WnXdD6gebz:b-896fix) ([raw file](https://github.com/starkware-libs/cairo/blob/660e5ab76059041f80e67240545b4e7c9bc08841/corelib/src/box.cairo#L2)):*
<details><summary><i>Previously, spapinistarkware (Shahar Papini) wrote…</i></summary><blockquote>

We should think of how to introduce visibility on impls.
Please add a TODO(spapini) for this somewhere.
</blockquote></details>

Done.
___
*[`crates/cairo-lang-defs/src/db.rs` line 668 at r8](https://reviewable.io/reviews/starkware-libs/cairo/3457#-NZ4v36uAUl4-edrlQpt:-NZ8nk32CXpg7fCEQ1lD:bl6t5v8) ([raw file](https://github.com/starkware-libs/cairo/blob/660e5ab76059041f80e67240545b4e7c9bc08841/crates/cairo-lang-defs/src/db.rs#L668)):*
<details><summary><i>Previously, spapinistarkware (Shahar Papini) wrote…</i></summary><blockquote>

You should make this query get MemberId, and fetch the ast of the member from it.
</blockquote></details>

Not sure how this is going to work. right now member ast is part of struct ast. And member id is interned only in the semantic stage. Alternative solution would be put ast visibility in the member semantic model. I think core issue is right now member ast is not indexed by member id in any place. Another solution will be add extra field in the module data, maybe called "struct_members" and index member ast by id in the defs group 
___
*[`crates/cairo-lang-parser/src/parser.rs` line 477 at r8](https://reviewable.io/reviews/starkware-libs/cairo/3457#-NZ4vUeD8x8S2mcWJUT6:-NZ96k-585eY3s4KddW9:b-896fix) ([raw file](https://github.com/starkware-libs/cairo/blob/660e5ab76059041f80e67240545b4e7c9bc08841/crates/cairo-lang-parser/src/parser.rs#L477)):*
<details><summary><i>Previously, spapinistarkware (Shahar Papini) wrote…</i></summary><blockquote>

No need for the `else` now. Unindent it.
</blockquote></details>

Done.
___
*[`crates/cairo-lang-semantic/src/lookup_item.rs` line 14 at r8](https://reviewable.io/reviews/starkware-libs/cairo/3457#-NZ53Elq3ddi4SQwpI1n:-NZ96oWPB4jsGyjoMNfz:b-896fix) ([raw file](https://github.com/starkware-libs/cairo/blob/660e5ab76059041f80e67240545b4e7c9bc08841/crates/cairo-lang-semantic/src/lookup_item.rs#L14)):*
<details><summary><i>Previously, spapinistarkware (Shahar Papini) wrote…</i></summary><blockquote>

Let's rename this.
Maybe `ItemEx`?
</blockquote></details>

Done.
___
*[`crates/cairo-lang-semantic/src/lookup_item.rs` line 17 at r8](https://reviewable.io/reviews/starkware-libs/cairo/3457#-NZ50QAz271rGYg8VMWT:-NZ8stIB71Fp2cQeU9cl:b5w6q1y) ([raw file](https://github.com/starkware-libs/cairo/blob/660e5ab76059041f80e67240545b4e7c9bc08841/crates/cairo-lang-semantic/src/lookup_item.rs#L17)):*
<details><summary><i>Previously, spapinistarkware (Shahar Papini) wrote…</i></summary><blockquote>

We should somehow document that this is not the entire visibility logic.
i.e. when we import `a::b::c`, c might be pub (and thus, visible_in will be true), but b might not be visible.
so c is still inaccessible.

Can you try to document this? Maybe we should also rename `visible` to reflect this? Maybe something like visible_given_parent, or something... WDYT?
</blockquote></details>

Basically it means ignoring or forgetting ancestors visibility, it is visible in the module with that particular module id. Of course not really ignoring the ancestor, because that module id may be a ancestor. Maybe rename to peek_visible_in? 
___
*[`crates/cairo-lang-semantic/src/lookup_item.rs` line 119 at r8](https://reviewable.io/reviews/starkware-libs/cairo/3457#-NZ4vyxy-JGC8ERZ-f2k:-NZ8ruSL5eRT0fyMmzUC:b-iy8zq5) ([raw file](https://github.com/starkware-libs/cairo/blob/660e5ab76059041f80e67240545b4e7c9bc08841/crates/cairo-lang-semantic/src/lookup_item.rs#L119)):*
<details><summary><i>Previously, spapinistarkware (Shahar Papini) wrote…</i></summary><blockquote>

why are these 3 returning true, instead of visible_in()?
</blockquote></details>

This three item are use, impl and impl alias. the pub logic for three are more complex. As I comment in the code, use need reexport logic, while pub impl will need to change inference logic. I think it's better to implement these two in next stage. So leave it true for now.
___
*[`crates/cairo-lang-semantic/src/expr/compute.rs` line 1962 at r8](https://reviewable.io/reviews/starkware-libs/cairo/3457#-NZ524_k5ApKmzpz8oHm:-NZ96wx--4fWmfiCw5UV:b-896fix) ([raw file](https://github.com/starkware-libs/cairo/blob/660e5ab76059041f80e67240545b4e7c9bc08841/crates/cairo-lang-semantic/src/expr/compute.rs#L1962)):*
<details><summary><i>Previously, spapinistarkware (Shahar Papini) wrote…</i></summary><blockquote>

Rephrase this with early returns.
```
let Some(visibility) = .. else { return; };
if visible_in(..) { return; }
...
```
</blockquote></details>

Done.
___
*[`crates/cairo-lang-semantic/src/expr/compute.rs` line 1964 at r8](https://reviewable.io/reviews/starkware-libs/cairo/3457#-NZ51wWO20WRcuN2hR5V:-NZ96zW27l4ZXRY6aTiR:b-896fix) ([raw file](https://github.com/starkware-libs/cairo/blob/660e5ab76059041f80e67240545b4e7c9bc08841/crates/cairo-lang-semantic/src/expr/compute.rs#L1964)):*
<details><summary><i>Previously, spapinistarkware (Shahar Papini) wrote…</i></summary><blockquote>

It's better to report the diagnostics, but not fail the computation. This way, we stay with more relevant semantic information.
</blockquote></details>

Done.
___
*[`crates/cairo-lang-semantic/src/items/visibilities.rs` line 5 at r8](https://reviewable.io/reviews/starkware-libs/cairo/3457#-NZ52QpE14GkCKmXRMqW:-NZ970Lk94R1VN5cfzHR:b-896fix) ([raw file](https://github.com/starkware-libs/cairo/blob/660e5ab76059041f80e67240545b4e7c9bc08841/crates/cairo-lang-semantic/src/items/visibilities.rs#L5)):*
<details><summary><i>Previously, spapinistarkware (Shahar Papini) wrote…</i></summary><blockquote>

Same documentation here as what I wrote in lookup_item.rs. Also, make the other doc point here.
</blockquote></details>

Done.
___
*[`crates/cairo-lang-semantic/src/resolve/mod.rs` line 907 at r8](https://reviewable.io/reviews/starkware-libs/cairo/3457#-NZ53SrZBQ9MKnpNAeyg:-NZ9-zhT9l26LI5EtgFQ:bhaci4) ([raw file](https://github.com/starkware-libs/cairo/blob/660e5ab76059041f80e67240545b4e7c9bc08841/crates/cairo-lang-semantic/src/resolve/mod.rs#L907)):*
<details><summary><i>Previously, spapinistarkware (Shahar Papini) wrote…</i></summary><blockquote>

Also, I tihnk we don't want a different diagnostic kind for each. Make one diagnostic kind with a LookupItemId.
</blockquote></details>

I think with ModuleItemId will be enough for now. In most language inference/trait function is always public. I think we will follow this rule as well, which means impl functions are always public as well in most case.

spapinistarkware (changes requested):
> Reviewed 1 of 55 files at r7, 5 of 49 files at r9.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3457)* status: 19 of 104 files reviewed, 2 unresolved discussions (waiting on @Jinwei1987)

___
*[`crates/cairo-lang-semantic/src/lookup_item.rs` line 120 at r9](https://reviewable.io/reviews/starkware-libs/cairo/3457#-NZL5cLvCF3dVf_sTEcH:-NZL5cLvCF3dVf_sTEcI:undefined) ([raw file](https://github.com/starkware-libs/cairo/blob/7bde4967d04f5cf4091e2f39655dd0ac72869e13/crates/cairo-lang-semantic/src/lookup_item.rs#L120)):*
> ```XML
> 
>     //TODO(spapini)
>     // Use need some reexport logic.
> ```





_Suggestion:_
```XML
    // TODO(spapini): Use need some reexport logic.
```
___
*[`crates/cairo-lang-semantic/src/expr/compute.rs` line 1962 at r9](https://reviewable.io/reviews/starkware-libs/cairo/3457#-NZL5ppR3izmiouBPSv7:-NZL5ppR3izmiouBPSv8:b3kqx) ([raw file](https://github.com/starkware-libs/cairo/blob/7bde4967d04f5cf4091e2f39655dd0ac72869e13/crates/cairo-lang-semantic/src/expr/compute.rs#L1962)):*
> ```XML
> ) -> Maybe<()> {
>     let source_module_id = struct_id.parent_module(ctx.db.upcast());
>     let Some(visibility) = ctx.db.struct_member_visibility(struct_id, member_name.clone())? else { return Ok(()) };
> ```

.



_Suggestion:_
```XML
let Some(visibility) = ctx.db.struct_member_visibility(struct_id, member_name.clone())? else {
        return Ok(())
        };
```

spapinistarkware (commented):
> Reviewed 10 of 55 files at r7, 2 of 26 files at r8, 26 of 49 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3457)* status: 57 of 104 files reviewed, 2 unresolved discussions (waiting on @Jinwei1987)

Jinwei1987 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3457)* status: 50 of 131 files reviewed, 2 unresolved discussions (waiting on @spapinistarkware)

___
*[`crates/cairo-lang-semantic/src/lookup_item.rs` line 120 at r9](https://reviewable.io/reviews/starkware-libs/cairo/3457#-NZL5cLvCF3dVf_sTEcH:-NZOlho_A0KU6moS9o8F:b-896fix) ([raw file](https://github.com/starkware-libs/cairo/blob/7bde4967d04f5cf4091e2f39655dd0ac72869e13/crates/cairo-lang-semantic/src/lookup_item.rs#L120)):*
> ```XML
> 
>     //TODO(spapini)
>     // Use need some reexport logic.
> ```

Done.
___
*[`crates/cairo-lang-semantic/src/expr/compute.rs` line 1962 at r9](https://reviewable.io/reviews/starkware-libs/cairo/3457#-NZL5ppR3izmiouBPSv7:-NZOlkVrBEmojZW4H8qN:b-896fix) ([raw file](https://github.com/starkware-libs/cairo/blob/7bde4967d04f5cf4091e2f39655dd0ac72869e13/crates/cairo-lang-semantic/src/expr/compute.rs#L1962)):*
<details><summary><i>Previously, spapinistarkware (Shahar Papini) wrote…</i></summary><blockquote>

.
</blockquote></details>

Done.

Jinwei1987:
I have fixed cargo test and cairo test

spapinistarkware (commented):
> Reviewed 1 of 40 files at r10.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3457)* status: 50 of 135 files reviewed, all discussions resolved

spapinistarkware (commented):
Looks good. Let's wait fo @orizi comments.

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3457#-:-NZQCObnEZe_jO1AHlYY:b-bxkcz0)* status: 50 of 135 files reviewed, all discussions resolved

orizi (approved):
:lgtm:

> Reviewed 1 of 11 files at r1, 3 of 9 files at r3, 17 of 55 files at r7, 11 of 26 files at r8, 18 of 49 files at r9, 39 of 40 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3457#-:-NZQLu9d3_AgZmgm-i5b:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Jinwei1987)

spapinistarkware (commented):
> Reviewed 3 of 9 files at r3, 14 of 55 files at r7, 10 of 26 files at r8, 6 of 49 files at r9, 38 of 40 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3457)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Jinwei1987)

spapinistarkware (approved):
> Reviewed 1 of 26 files at r8, 12 of 49 files at r9, 1 of 40 files at r10.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3457)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Jinwei1987)

Jinwei1987 (commented):
> Done
there is some issue with cairo fmt, will let you guys know once I fix it.

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3457#-:-NZ4krkF4ZtaBJJwT8zc:b-xfwie)* status: 123 of 137 files reviewed, all discussions resolved (waiting on @orizi and @spapinistarkware)

spapinistarkware (commented):
> Reviewed 6 of 14 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3457)* status: 129 of 137 files reviewed, all discussions resolved (waiting on @orizi)

spapinistarkware (approved):
> Reviewed 8 of 14 files at r11.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3457)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Jinwei1987)

Jinwei1987:
Everything shall be fixed right now. Please review and approve again.

spapinistarkware (approved):
> Reviewed 20 of 20 files at r12, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3457)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Jinwei1987)

spapinistarkware:
@Jinwei1987 if you sign up for onlydust, we can reward you

PR #3628, opened 2023-07-06
feat: Cairo Formatter Unwanted Change in Spacing

neotheprogramist:
When formatting the following piece of code using Cairo Formatter, it's observed that an unwanted change in spacing occurs.

***Original Code:***
```cairo
match array_pop_front(ref self) {
    Option::Some(x) => Option::Some(x.unbox()),
    Option::None => Option::None(()),
}
```

***Formatted Code:***
```cairo
match array_pop_front(ref self) {
    Option::Some(x) => Option::Some(x.unbox()),
    Option::None=> Option::None(()),
}
```

Please note that the spacing between Option::None and => is removed in the formatted version, which isn't expected. The spacing in the original code is correct according to the standard formatting guidelines and should not be modified.

- - -
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3628)


gilbens-starkware (commented):
Thanks for the contribution, however, the same issue was fixed in another PR (#3633) which also address another minor issue so we merged it.

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3628#-:-NZydawf0VumkeItAPYs:b-2d4vq4)* status: 0 of 12 files reviewed, all discussions resolved

PR #3704, opened 2023-07-18
Add derive for LegacyHash; Resolves #3684

### LINKED ISSUE #3684: feat: using composite types as map keys
# Feature Request

It would be awesome if there was an automagical way how we could use custom composite types as keys in `LegacyMap`.

For example, I have this type:

```rust
struct Pool {
    tokenA: ContractAddress,
    tokenB: ContractAddress,
    fee: u128
}
```

and I want to use it as a key in a storage variable, e.g. `is_enabled: LegacyMap<Pool, bool>`.

**Describe Preferred Solution**

I think the best solution would be if there was a something like a `MapKey` trait that would transform the type into a `felt252` by hashing all its members, which could then be used in the `LegacyMap`. This should, of course, be automatically implemented by the Starknet plugin via a derive arg:

```rust
trait MapKey<T> {
    fn as_map_key(self: T) -> felt252;
}

#[derive(MapKey)]
struct Pool {
    tokenA: ContractAddress,
    tokenB: ContractAddress,
    fee: u128
}

//
// MapKey derive would create an impl block of MapKey<Pool>, e.g.
//

impl PoolMapKeyImpl of MapKey<Pool> {
    fn as_map_key(self: Pool) -> felt252 {
        let mut members: Array<felt252> = Default::default();
        members.append(self.tokenA.into());
        members.append(self.tokenB.into());
        members.append(self.fee.into());

        poseidon_hash_span(members.span())
    }
}

//
// and then in my contract
//

fn disable_pool(ref self: ContractState, pool: Pool) {
    self.is_enabled.write(pool, false);
}
```

**Describe Alternatives**

I know I can use a tuple as a key in the `LegacyMap` declaration, but that has its limits.

ArielElp:
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3704)

### DIFF
--- a/crates/cairo-lang-plugins/src/plugins/derive.rs
+++ b/crates/cairo-lang-plugins/src/plugins/derive.rs
@@ -182,6 +182,7 @@ fn generate_derive_code_for_type(
                         message: "Unsupported trait for derive for extern types.".into(),
                     })
                 }
+                "LegacyHash" => impls.push(get_legacy_hash_impl(&name, &extra_info)),
                 _ => {
                     // TODO(spapini): How to allow downstream derives while also
                     //  alerting the user when the derive doesn't exist?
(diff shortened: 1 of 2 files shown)


orizi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3704)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @ArielElp)

*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/3704#-NakaegGAooHCqE9-GxR:-NakaegGAooHCqE9-GxS:bcpiogc) (no related file):*
rebase and make it for `Hash`

___

PR #3800, opened 2023-07-31
refactor(bigint): improve encode/decode implementation

LucasLvy:
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3800)

### DIFF
--- a/crates/cairo-lang-utils/src/bigint.rs
+++ b/crates/cairo-lang-utils/src/bigint.rs
@@ -112,22 +112,25 @@ where
 
 impl Encode for BigIntAsHex {
     fn size_hint(&self) -> usize {
-        // sign, len, data.
-        1 + 8 + self.value.to_bytes_be().1.len()
+        // sign + len packed in the same byte, it allows numbers of byte size up to 63 (2**504),
+        // data.
+        let bits = self.value.bits() as usize;
+        core::mem::size_of::<u8>() + bits / 8 + if bits % 8 != 0 { 1 } else { 0 }
     }
+
+    /// /!\ Warning this function panics if the number encoded is too big (>= 2**504)
     fn encode_to<T: parity_scale_codec::Output + ?Sized>(&self, dest: &mut T) {
-        let (sign, data) = self.value.to_bytes_be();
-        match sign {
-            num_bigint::Sign::Minus => 0u8.encode_to(dest),
-            num_bigint::Sign::NoSign => 1u8.encode_to(dest),
-            num_bigint::Sign::Plus => 2u8.encode_to(dest),
-        };
-        // TODO(yair): better way to encode vec?
-        let len = data.len() as u64;
-        len.encode_to(dest);
-        for b in data.as_slice() {
-            b.encode_to(dest);
-        }
+        let (sign, data) = self.value.to_bytes_le();
+        assert!(data.len() <= 63, "Can't encode numbers longer than 63 bytes");
+        // Pack sign + number byte size.
+        ((match sign {
+            num_bigint::Sign::Minus => 0u8,
+            num_bigint::Sign::NoSign => 1u8,
+            num_bigint::Sign::Plus => 2u8,
+        } << 6)
+            + data.len() as u8)
+            .encode_to(dest);
+        dest.write(&data);
     }
 }
 
@@ -135,8 +138,8 @@ impl Decode for BigIntAsHex {
     fn decode<I: parity_scale_codec::Input>(
         input: &mut I,
     ) -> Result<Self, parity_scale_codec::Error> {
-        let discriminant = input.read_byte()?;
-        let sign = match discriminant {
+        let sign_and_len = input.read_byte()?;
+        let sign = match sign_and_len >> 6 {
             0u8 => num_bigint::Sign::Minus,
             1u8 => num_bigint::Sign::NoSign,
             2u8 => num_bigint::Sign::Plus,
@@ -144,12 +147,9 @@ impl Decode for BigIntAsHex {
                 return Err(parity_scale_codec::Error::from("Bad sign encoding."));
             }
         };
-        // TODO(yair): better way to decode vec?
-        let data_len = u64::decode(input)?;
-        let mut bytes: Vec<u8> = Vec::with_capacity(data_len as usize);
-        for _ in 0..data_len {
-            bytes.push(input.read_byte()?);
-        }
-        Ok(Self { value: BigInt::from_bytes_be(sign, bytes.as_slice()) })
+        let len = sign_and_len & 0b00111111;
+        let mut buffer = vec![0; len as usize];
+        input.read(&mut buffer)?;
+        Ok(Self { value: BigInt::from_bytes_le(sign, buffer.as_slice()) })
     }
 }


crates/cairo-lang-utils/src/bigint.rs
```diff
@@ -112,44 +112,40 @@ where
 
 impl Encode for BigIntAsHex {
     fn size_hint(&self) -> usize {
-        // sign, len, data.
-        1 + 8 + self.value.to_bytes_be().1.len()
+        // sign (2 is the biggest sign value possible), data.
+        (2u8, self.value.to_bytes_le().1).size_hint()
```
tdelabro, line 116: What value does this return?
I expect it to be equal to `1 + 1 + self.value.to_bytes_le().1.len()`

tdelabro: I don't like this implementation uses `to_bytes_le` which is performing memory allocation internally (a new vec is crated).

I think `size_hint` should not do that. Just read the data and compute it's size.
YOu can instead use `mem::size_of_val`. That will look at the size of the inner vec

tdelabro (changes requested):

crates/cairo-lang-utils/src/bigint.rs
tdelabro: `it prunes the 0 bytes` what do you mean?

tdelabro: You can get the inner `data` by calling the `IntDigits` trait `digits` method.
And actually, it's true length. Then you just have to multiply by 8

```rust
impl IntDigits for BigUint {
    #[inline]
    fn digits(&self) -> &[BigDigit] {
        &self.data
    }
    #[inline]
    fn digits_mut(&mut self) -> &mut Vec<BigDigit> {
        &mut self.data
    }
    #[inline]
    fn normalize(&mut self) {
        self.normalize();
    }
    #[inline]
    fn capacity(&self) -> usize {
        self.data.capacity()
    }
    #[inline]
    fn len(&self) -> usize {
        self.data.len()
    }
}

```

LucasLvy: This is the implementation of `to_bytes_le()`
```rs
/// Returns the byte representation of the `BigUint` in little-endian byte order.
    ///
    /// # Examples
    ///
    /// ```
    /// use num_bigint::BigUint;
    ///
    /// let i = BigUint::parse_bytes(b"1125", 10).unwrap();
    /// assert_eq!(i.to_bytes_le(), vec![101, 4]);
    /// ```
    #[inline]
    pub fn to_bytes_le(&self) -> Vec<u8> {
        if self.is_zero() {
            vec![0]
        } else {
            convert::to_bitwise_digits_le(self, 8)
        }
    }
```
If i used `core::mem::size_of(val)` it would have returned 8 which would be wrong

yair-starkware (approved):
> Reviewed 1 of 1 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3800)* status: all files reviewed, 2 unresolved discussions (waiting on @LucasLvy, @orizi, and @tdelabro)

___
*[`crates/cairo-lang-utils/src/bigint.rs` line 146 at r1](https://reviewable.io/reviews/starkware-libs/cairo/3800#-NagL9tWAIH5ExDswqUs:-NagL9tWAIH5ExDswqUt:b-di5ja4) ([raw file](https://github.com/starkware-libs/cairo/blob/187a2e3ace50e4f708d8c5c9ce5f49fea4d78df4/crates/cairo-lang-utils/src/bigint.rs#L146)):*
> ```XML
>             }
>         };
>         let mut buffer = vec![0; len.0 as usize];
> ```

Isn't `Vec::with_capacity` better?



_Code quote:_
```XML
let mut buffer = vec![0; len.0 as usize];
```

LucasLvy (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3800)* status: all files reviewed, 2 unresolved discussions (waiting on @orizi, @tdelabro, and @yair-starkware)

___
*[`crates/cairo-lang-utils/src/bigint.rs` line 146 at r1](https://reviewable.io/reviews/starkware-libs/cairo/3800#-NagL9tWAIH5ExDswqUs:-NagLvy29scGYq1lGWAj:b-os7pxs) ([raw file](https://github.com/starkware-libs/cairo/blob/187a2e3ace50e4f708d8c5c9ce5f49fea4d78df4/crates/cairo-lang-utils/src/bigint.rs#L146)):*
<details><summary><i>Previously, yair-starkware (Yair) wrote…</i></summary><blockquote>

Isn't `Vec::with_capacity` better?
</blockquote></details>

Generally speaking it should but for this use case clippy is not happy because theoretically the input could be empty. Here i don't think it matters too much as it's done at compile time (the optimizer is probably also doing some dark magic)

crates/cairo-lang-utils/src/bigint.rs
```diff
@@ -112,44 +112,39 @@ where
 
 impl Encode for BigIntAsHex {
     fn size_hint(&self) -> usize {
-        // sign, len, data.
-        1 + 8 + self.value.to_bytes_be().1.len()
+        // sign, data.
+        core::mem::size_of::<u8>() + self.value.bits() as usize / 8
```
tdelabro, line 116: YOu are missing one u8 there.
Sign takes one, vec len takes one, and then the rest is `self.value.bits() as usize / 8`

crates/cairo-lang-utils/src/bigint.rs
```diff
@@ -112,44 +112,39 @@ where
...
             }
         };
-        // TODO(yair): better way to decode vec?
-        let data_len = u64::decode(input)?;
-        let mut bytes: Vec<u8> = Vec::with_capacity(data_len as usize);
-        for _ in 0..data_len {
-            bytes.push(input.read_byte()?);
-        }
-        Ok(Self { value: BigInt::from_bytes_be(sign, bytes.as_slice()) })
+        let mut buffer = vec![0; len.0 as usize];
```
tdelabro, line 146: I would rather use `Vec::with_capacity`, you don't have to initialize the memory to `0`. It will be overwritten just after without ever being read.

tdelabro (changes requested):

crates/cairo-lang-utils/src/bigint.rs
tdelabro: s there a domain constraint we can rely upon in order to be sure that the vec won't be more than X byte long?

If those BigInt are used to store Felt252, we know for sure that the data len won't exceed 32. Which is totally encodable on a single byte with compact scale codec

tdelabro: Is there a way to be sure that `self.value.bits() as usize / 8` is always equal to `self.value.to_bytes_le().len()`?
Can we have some unit tests for this?
Or at least a comment that says that it has been tested and is always the case. In a few month it will be hard to make the link between those two just by reading the code.

LucasLvy: It surely does not always equal as it's an integer div we need to ceil 🫠

LucasLvy: As for the total size as it's a hint we don't have any guarantees that it'll be a felt252 there (imo)

tdelabro: We should rather round-up than down, shouldn't we?

tdelabro:
> and in u8(up to 255) bytes we pretty much have an infinitely large number - so we should be quite fine.
we can probably even not use 2 bits and keep using sign encoded within it, so we'd have up to 64 bytes.

If the len is stored as compact, it already uses two bits for the scale codec, so the max value (if stored on one byte) is already `63`.

tdelabro:
@orizi 
Also if we are comfortable with those values never being larger than `Felt::MAX` (and it seems we are, since we are willing to `panic!`), it should probably be enforced at the type level itself. Through some kind of wrapper or a type wich internal representation is not a potential infinite Vec, but rather a `[u8; 32]`. Or even a `BoundedVec`. 

Wdyt?

orizi (approved):
:lgtm:

> Reviewed 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3800#-:-NauaeuT2O1OBo5qlrLL:bnfp4nl)* status: all files reviewed, 4 unresolved discussions (waiting on @LucasLvy, @tdelabro, and @yair-starkware)

LucasLvy (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/3800)* status: all files reviewed, 3 unresolved discussions (waiting on @tdelabro)

___
*[`crates/cairo-lang-utils/src/bigint.rs` line 116 at r1](https://reviewable.io/reviews/starkware-libs/cairo/3800#gh-1279783112:-NavyUPV0LObr6WojlQN:b-896fix) ([raw file](https://github.com/starkware-libs/cairo/blob/187a2e3ace50e4f708d8c5c9ce5f49fea4d78df4/crates/cairo-lang-utils/src/bigint.rs#L116)):*

it sounds ok for me to check the result and panic if there is any issue.
and in u8(up to 255) bytes we pretty much have an infinitely large number - so we should be quite fine.
we can probably even not use 2 bits and keep using sign encoded within it, so we'd have up to 64 bytes.
</blockquote></details>

Done.
___
*[`crates/cairo-lang-utils/src/bigint.rs` line 146 at r1](https://reviewable.io/reviews/starkware-libs/cairo/3800#gh-1279794430:-NavyTPR8HjheSVdAGMd:b-896fix) ([raw file](https://github.com/starkware-libs/cairo/blob/187a2e3ace50e4f708d8c5c9ce5f49fea4d78df4/crates/cairo-lang-utils/src/bigint.rs#L146)):*
<details><summary><i>Previously, tdelabro (Timothée Delabrouille) wrote…</i></summary><blockquote>

I would rather use `Vec::with_capacity`, you don't have to initialize the memory to `0`. It will be overwritten just after without ever being read.
</blockquote></details>

Done.

PR #4194, opened 2023-10-04
feat: Storage{Base}Address impls

### LINKED ISSUE #4195: feat: add missing implementations for Storage{Base}Access
# Feature Request

**Describe the Feature Request**
- Implement the `Into` trait to convert a `Storage{Base}Address` into a `felt252`
- Implement the `Into` trait to convert a `Storage{Base}Address` into a `u256` 
- Implement `PartialEq` trait to allow comparison of `StorageBaseAddress` and `StorageAddress`

**Describe Preferred Solution**

**Describe Alternatives**

**Related Code**

**Additional Context**

**If the feature request is approved, would you be willing to submit a PR?**
_(Help can be provided if you need assistance submitting a PR)_

- [x] Yes
- [ ] No

enitrat:
- Implements the `Into` trait to convert a `Storage{Base}Address` into a `felt252`
- Implements the `Into` trait to convert a `Storage{Base}Address` into a `u256` 
- Implements `PartialEq` trait to allow comparison of `StorageBaseAddress` and `StorageAddress`

Resolves #4195

- - -
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/4194)

### DIFF
--- a/corelib/src/starknet/storage_access.cairo
+++ b/corelib/src/starknet/storage_access.cairo
@@ -42,9 +42,23 @@ impl StorageAddressIntoFelt252 of Into<StorageAddress, felt252> {
     }
 }
 
+impl StorageAddressIntoU256 of Into<StorageAddress, u256> {
+    fn into(self: StorageAddress) -> u256 {
+        let self: felt252 = self.into();
+        self.into()
+    }
+}
+
 impl StorageBaseAddressIntoFelt252 of Into<StorageBaseAddress, felt252> {
     fn into(self: StorageBaseAddress) -> felt252 {
         storage_address_from_base(self).into()
+    }
+}
+
+impl StorageBaseAddressIntoU256 of Into<StorageBaseAddress, u256> {
+    fn into(self: StorageBaseAddress) -> u256 {
+        let self: felt252 = self.into();
+        self.into()
     }
 }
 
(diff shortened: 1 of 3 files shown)


orizi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/4194)* status: all files reviewed, 1 unresolved discussion (waiting on @enitrat)

___
*[`corelib/src/starknet/storage_access.cairo` line 50 at r1](https://reviewable.io/reviews/starkware-libs/cairo/4194#-NfuHy3h9Kmb3G89bu_X:-NfuHy3i2NwuL0lEGNMF:b-k46qts) ([raw file](https://github.com/starkware-libs/cairo/blob/c381ceea4c386b29dbfdc40bf0fe33bcd0f5f401/corelib/src/starknet/storage_access.cairo#L50)):*
> ```Cairo
>         self.into()
>     }
> }
> ```

this currently does not exist for most types - why add this?



_Code quote:_
```Cairo
impl StorageAddressIntoU256 of Into<StorageAddress, u256> {
    fn into(self: StorageAddress) -> u256 {
        let self: felt252 = self.into();
        self.into()
    }
}
```

enitrat:
I was thinking that adding them gradually to the corelibrary would be great. It's a bit frustrating to always have to doublecheck whether a type conversion is in the corelib or not, only to end up writing it ourselves.

Perhaps I can make the process more generic and write XIntoY where X can be converted to felt252 and felt252 can be converted to Y; and same for TryInto

That would actually solve most of these cases

enitrat:
If you don't want these kind of implementations in the core library then I can add them to Alexandria instead. 

I understand that it might increase considerably the size of the corelib if we start implementing all traits for corelib types. 
However, it's very convenient, devX wise, to be able to natively perform all of these operations without having to think about where all my`impl` are defined, and explicitly importing them

Issue #4231, opened 2023-10-09
feat: is_null function in NullableTrait


chachaleo:
# Feature Request

**Describe the Feature Request**

Function that returns true if the Nullable is null and false otherwise

**Related Code**
```rust
fn is_null<+Drop<T>>(self: Nullable<T>) -> bool;
```
**If the feature request is approved, would you be willing to submit a PR?**

- [x] Yes

PR #4232, opened 2023-10-09
addition function nullable is_null

chachaleo:
Resolves issue #4231

I added a function is_null function is the NullableTrait.

When I was solving the Node Guardian quest : racing riverboats, I thought it could be useful to have this methods to smoothly check if a value is present is a dict or not.


orizi (commented):
This requires to be snapshot function - and this isn't (and can't be, since we are missing a libfunc).
adding to add such a libfunc to the backlog.

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/4232#-:-NgYRhes4fgIRs83_Cbh:bazr4c9)* status: 0 of 1 files reviewed, all discussions resolved

PR #4301, opened 2023-10-25
Added prelude module option.

orizi:
Currently only backwards compatible for latest release.

---

**Stack**:
- #4311
- #4306
- #4305
- #4304
- #4301 ⬅

- - -
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/4301)


mkaput (approved):
> Reviewed 4 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/4301)* status: all files reviewed, 3 unresolved discussions (waiting on @orizi)

___
*[`corelib/src/lib.cairo` line 348 at r1](https://reviewable.io/reviews/starkware-libs/cairo/4301#-NhaTpfX3GcS_tWSQwn5:-NhaTpfX3GcS_tWSQwn6:ba0fw7i) ([raw file](https://github.com/starkware-libs/cairo/blob/732938ed247e1f213582dd199bf6a0ff66ddf341/corelib/src/lib.cairo#L348)):*
> ```Cairo
> 
> // Preludes.
> mod prelude_v0;
> ```

nit, rust has `prelude::v0` and thanks to this all preludes are grouped in fs. I'm not enforcing this suggestion, as this is bikeshedding-prone.

In `PreludeSubmodule::name` you can always assume the common prefix `core::prelude`, and return just `v0`/`v1` etc there.
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 25 at r1](https://reviewable.io/reviews/starkware-libs/cairo/4301#-NhaUi8_03vEYudlt9Tw:-NhaUi8_03vEYudlt9Tx:b-2wsleu) ([raw file](https://github.com/starkware-libs/cairo/blob/732938ed247e1f213582dd199bf6a0ff66ddf341/crates/cairo-lang-filesystem/src/db.rs#L25)):*
> ```XML
>     pub root: Directory,
>     /// The prelude version for the crate.
>     pub prelude_submodule: PreludeSubmodule,
> ```

Isn't this too specific? I think for greater good, it would be rather good to only allow users to specify Rust-style editions config knob. Wdyt?
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 38 at r1](https://reviewable.io/reviews/starkware-libs/cairo/4301#-NhaV2A-1ukGFJGPp5zM:-NhaV2A-1ukGFJGPp5zN:b24o9mh) ([raw file](https://github.com/starkware-libs/cairo/blob/732938ed247e1f213582dd199bf6a0ff66ddf341/crates/cairo-lang-filesystem/src/db.rs#L38)):*
> ```XML
> pub enum PreludeSubmodule {
>     #[default]
>     BackwardCompatibility,
> ```

nit, _backward_ but in relation to what?



_Suggestion:_
```XML
V0
```

orizi (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/4301)* status: all files reviewed, 2 unresolved discussions (waiting on @mkaput)

___
*[`corelib/src/lib.cairo` line 348 at r1](https://reviewable.io/reviews/starkware-libs/cairo/4301#-NhaTpfX3GcS_tWSQwn5:-NhaoIG-EVkcaAm6vKZC:b-q9hmds) ([raw file](https://github.com/starkware-libs/cairo/blob/732938ed247e1f213582dd199bf6a0ff66ddf341/corelib/src/lib.cairo#L348)):*

nit, rust has `prelude::v0` and thanks to this all preludes are grouped in fs. I'm not enforcing this suggestion, as this is bikeshedding-prone.

In `PreludeSubmodule::name` you can always assume the common prefix `core::prelude`, and return just `v0`/`v1` etc there.
</blockquote></details>

gonna probably just have v0 and than go the v2023_10 (more date based)
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 25 at r1](https://reviewable.io/reviews/starkware-libs/cairo/4301#-NhaUi8_03vEYudlt9Tw:-NhapK6KCeLuA0duXWEN:b-z4i776) ([raw file](https://github.com/starkware-libs/cairo/blob/732938ed247e1f213582dd199bf6a0ff66ddf341/crates/cairo-lang-filesystem/src/db.rs#L25)):*

Isn't this too specific? I think for greater good, it would be rather good to only allow users to specify Rust-style editions config knob. Wdyt?
</blockquote></details>

Hmm - changed to version - and made prelude a simple member function.

yuvalsw (changes requested):
> Reviewed 6 of 6 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/4301)* status: all files reviewed, 7 unresolved discussions (waiting on @orizi)

___
*[`crates/cairo-lang-filesystem/src/db.rs` line 47 at r2](https://reviewable.io/reviews/starkware-libs/cairo/4301#-NhfUIpl3tNAPG9pBIGK:-NhfUIpm7vebkzSS0spL:byg842r) ([raw file](https://github.com/starkware-libs/cairo/blob/759437ad2b130d971a27d1cca28180ba1e297e82/crates/cairo-lang-filesystem/src/db.rs#L47)):*
> ```XML
>         }
>     }
> }
> ```

This is all unclear:
- Compatibility with what?
- please add what's gonna happen when we add new versions.
-   Will the default change too?



_Code quote:_
```XML
/// The versions of compatibilty.
#[derive(Clone, Debug, Default, Hash, PartialEq, Eq)]
pub enum CompatibilityVersion {
    /// The base version for compatibility.
    #[default]
    V0,
}
impl CompatibilityVersion {
    pub fn prelude_submodule(&self) -> &str {
        match self {
            Self::V0 => "v0",
        }
    }
}
```
___
*[`crates/cairo-lang-semantic/src/corelib.rs` line 36 at r2](https://reviewable.io/reviews/starkware-libs/cairo/4301#-NhfUt_f2PmQ8tGkgdaI:-NhfUt_f2PmQ8tGkgdaJ:undefined) ([raw file](https://github.com/starkware-libs/cairo/blob/759437ad2b130d971a27d1cca28180ba1e297e82/crates/cairo-lang-semantic/src/corelib.rs#L36)):*
> ```XML
> }
> 
> /// Returns the submodule of `base` name `submodule_name`
> ```





_Suggestion:_
```XML
/// Returns the submodule of `base`, named `submodule_name`, if exists.
```
___
*[`crates/cairo-lang-semantic/src/corelib.rs` line 39 at r2](https://reviewable.io/reviews/starkware-libs/cairo/4301#-NhfVHhO0lVxLKDAatOG:-NhfVHhO0lVxLKDAatOH:b-x77oa6) ([raw file](https://github.com/starkware-libs/cairo/blob/759437ad2b130d971a27d1cca28180ba1e297e82/crates/cairo-lang-semantic/src/corelib.rs#L39)):*
> ```XML
> pub fn get_submodule(
>     db: &dyn SemanticGroup,
>     base: ModuleId,
> ```

consider renaming to `parent`/`parent_module`/`containing_module`.



_Code quote:_
```XML
base
```
___
*[`crates/cairo-lang-semantic/src/corelib.rs` line 42 at r2](https://reviewable.io/reviews/starkware-libs/cairo/4301#-NhfVXykE0dK-IkVY_oc:-NhfVXykE0dK-IkVY_od:b7zzpwj) ([raw file](https://github.com/starkware-libs/cairo/blob/759437ad2b130d971a27d1cca28180ba1e297e82/crates/cairo-lang-semantic/src/corelib.rs#L42)):*
> ```XML
>     submodule_name: &str,
> ) -> Option<ModuleId> {
>     let submodules = db.module_submodules(base).unwrap();
> ```

return None if given a non existing base.



_Code quote:_
```XML
.unwrap();
```
___
*[`crates/cairo-lang-semantic/src/corelib.rs` line 52 at r2](https://reviewable.io/reviews/starkware-libs/cairo/4301#-NhfV5Z6B4KphYpN5bxS:-NhfV5Z6B4KphYpN5bxT:b-8v89x6) ([raw file](https://github.com/starkware-libs/cairo/blob/759437ad2b130d971a27d1cca28180ba1e297e82/crates/cairo-lang-semantic/src/corelib.rs#L52)):*
> ```XML
> }
> 
> /// Returns a submodule of the corelib named `submodule_name`.
> ```

Panics if it doesn't exist.
___
*[`crates/cairo-lang-semantic/src/resolve/mod.rs` line 773 at r2](https://reviewable.io/reviews/starkware-libs/cairo/4301#-NhfkEIbEWbzmBa0hmXz:-NhfkEIbEWbzmBa0hmY-:b45h) ([raw file](https://github.com/starkware-libs/cairo/blob/759437ad2b130d971a27d1cca28180ba1e297e82/crates/cairo-lang-semantic/src/resolve/mod.rs#L773)):*
> ```XML
>             .map(|config| config.compatibility_version)
>             .unwrap_or_default();
>         let prelude_submodule_name = compatibility_version.prelude_submodule();
> ```





_Suggestion:_
```XML
prelude_submodule_name
```
___
*[`crates/cairo-lang-semantic/src/resolve/mod.rs` line 776 at r2](https://reviewable.io/reviews/starkware-libs/cairo/4301#-Nhfk939ALkSGFVs-pjD:-Nhfk939ALkSGFVs-pjE:undefined) ([raw file](https://github.com/starkware-libs/cairo/blob/759437ad2b130d971a27d1cca28180ba1e297e82/crates/cairo-lang-semantic/src/resolve/mod.rs#L776)):*
> ```XML
>         let core_prelude_submodule = core_submodule(self.db, "prelude");
>         Some(get_submodule(self.db, core_prelude_submodule, prelude_submodule_name).unwrap_or_else(
>             || panic!("expected prelude submodule `{prelude_submodule_name}` not found."),
> ```





_Suggestion:_
```XML
dule_name}` not found in `core::prelude`."),
```

yuvalsw (changes requested):
> Reviewed 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/4301)* status: all files reviewed, 2 unresolved discussions (waiting on @orizi)

___
*[`crates/cairo-lang-semantic/src/corelib.rs` line 42 at r2](https://reviewable.io/reviews/starkware-libs/cairo/4301#-NhfVXykE0dK-IkVY_oc:-Nhg5TV34YC4_d3l-VWL:b-mq45o) ([raw file](https://github.com/starkware-libs/cairo/blob/759437ad2b130d971a27d1cca28180ba1e297e82/crates/cairo-lang-semantic/src/corelib.rs#L42)):*

this can never happen - as this is a valid module.
this is the same code as before.
</blockquote></details>

before it was always for core module, which is indeed always a valid module. Now you may get a nonexisting module as the base module.
___
*[`crates/cairo-lang-semantic/src/resolve/mod.rs` line 773 at r2](https://reviewable.io/reviews/starkware-libs/cairo/4301#-NhfkEIbEWbzmBa0hmXz:-Nhg5pGK8vrqdTSbyjFr:bwynrd6) ([raw file](https://github.com/starkware-libs/cairo/blob/759437ad2b130d971a27d1cca28180ba1e297e82/crates/cairo-lang-semantic/src/resolve/mod.rs#L773)):*

this was already the name.
</blockquote></details>

LOL, it's reviewable. I pointed to the method name, which you now changed anyway.

orizi (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/4301)* status: all files reviewed, 2 unresolved discussions (waiting on @yuvalsw)

___
*[`crates/cairo-lang-semantic/src/corelib.rs` line 42 at r2](https://reviewable.io/reviews/starkware-libs/cairo/4301#-NhfVXykE0dK-IkVY_oc:-Nhg6MRB9yQiHJ_nZDMk:bsvdara) ([raw file](https://github.com/starkware-libs/cairo/blob/759437ad2b130d971a27d1cca28180ba1e297e82/crates/cairo-lang-semantic/src/corelib.rs#L42)):*

before it was always for core module, which is indeed always a valid module. Now you may get a nonexisting module as the base module.
</blockquote></details>

how can you get a non-existing module id?

yuvalsw (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/4301)* status: all files reviewed, 1 unresolved discussion (waiting on @orizi)

___
*[`crates/cairo-lang-semantic/src/corelib.rs` line 42 at r2](https://reviewable.io/reviews/starkware-libs/cairo/4301#-NhfVXykE0dK-IkVY_oc:-Nhg9wAv-TAVl0TXG-WL:bub0mg) ([raw file](https://github.com/starkware-libs/cairo/blob/759437ad2b130d971a27d1cca28180ba1e297e82/crates/cairo-lang-semantic/src/corelib.rs#L42)):*

how can you get a non-existing module id?
</blockquote></details>

Build it manually. In the end it's a u32 underneath. I think it's possible. Didn't try to, but I am pretty sure we don't assume this in the rest of the code. If we could, we could delete some '?'s which I'd love to do.

orizi (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/4301)* status: 5 of 6 files reviewed, 1 unresolved discussion (waiting on @yuvalsw)

___
*[`crates/cairo-lang-semantic/src/corelib.rs` line 42 at r2](https://reviewable.io/reviews/starkware-libs/cairo/4301#-NhfVXykE0dK-IkVY_oc:-NhgFO-PEmUlUUaIZlJm:bdwm7xo) ([raw file](https://github.com/starkware-libs/cairo/blob/759437ad2b130d971a27d1cca28180ba1e297e82/crates/cairo-lang-semantic/src/corelib.rs#L42)):*

Build it manually. In the end it's a u32 underneath. I think it's possible. Didn't try to, but I am pretty sure we don't assume this in the rest of the code. If we could, we could delete some '?'s which I'd love to do.
</blockquote></details>

Added in the doc and an expect.
Generally - whenever we know and can doc - i don't have an issue with unwraping.
here we know the only 2 callers of this function.

yuvalsw (changes requested):
> Reviewed 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/4301)* status: all files reviewed, 1 unresolved discussion (waiting on @orizi)

___
*[`crates/cairo-lang-semantic/src/corelib.rs` line 42 at r2](https://reviewable.io/reviews/starkware-libs/cairo/4301#-NhfVXykE0dK-IkVY_oc:-NhgKh0t1uCP4yGfw5Jk:b-bu0lnr) ([raw file](https://github.com/starkware-libs/cairo/blob/759437ad2b130d971a27d1cca28180ba1e297e82/crates/cairo-lang-semantic/src/corelib.rs#L42)):*

Added in the doc and an expect.
Generally - whenever we know and can doc - i don't have an issue with unwraping.
here we know the only 2 callers of this function.
</blockquote></details>

You can't rely on the current callers. I agree that whenever we can - it's better to unwrap and doc rather than '?'. But in this case we can't (I think - would be happy to know we can and then would change in other places too).
What you added has a more elaborated error, but still panics. Since we anyway return an option here, I'd just change it to `db.module_submodules(base_module).ok()?` so it returns None in case the given base module doesn't contain the submodule, which is indeed the case if the given base module doesn't exist.

yuvalsw (changes requested):
> Reviewed 9 of 9 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/4301)* status: all files reviewed, 1 unresolved discussion (waiting on @orizi)

___
*[`corelib/src/felt_252.cairo` line 1 at r5](https://reviewable.io/reviews/starkware-libs/cairo/4301#-NhhFtaIDRrHG8I_FVpK:-NhhFtaJ-XjxC1PrQgTk:bunoo43) ([raw file](https://github.com/starkware-libs/cairo/blob/8e3a76887ca5146f099b065e42b6f1ef2a7ec023/corelib/src/felt_252.cairo#L1)):*
> ```Cairo
> impl Felt252Zero of core::num::traits::Zero<felt252> {
> ```

why did you have to add `core::` everywhere? Does it mean user code also needs it?
If so - it would break.

orizi (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/4301)* status: all files reviewed, 1 unresolved discussion (waiting on @yuvalsw)

___
*[`corelib/src/felt_252.cairo` line 1 at r5](https://reviewable.io/reviews/starkware-libs/cairo/4301#-NhhFtaIDRrHG8I_FVpK:-NhhHUJX8fcY3LkgPm61:b-69kekd) ([raw file](https://github.com/starkware-libs/cairo/blob/8e3a76887ca5146f099b065e42b6f1ef2a7ec023/corelib/src/felt_252.cairo#L1)):*

why did you have to add `core::` everywhere? Does it mean user code also needs it?
If so - it would break.
</blockquote></details>

No code was released with this.
The base prelude matches the last release.

mkaput (approved):
> Reviewed 2 of 6 files at r2, 2 of 3 files at r3, 9 of 9 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/4301)* status: all files reviewed, 1 unresolved discussion (waiting on @yuvalsw)

___
*[`corelib/src/lib.cairo` line 348 at r1](https://reviewable.io/reviews/starkware-libs/cairo/4301#-NhaTpfX3GcS_tWSQwn5:-NhkPFou6r2dG-BJRxqi:b4putsz) ([raw file](https://github.com/starkware-libs/cairo/blob/732938ed247e1f213582dd199bf6a0ff66ddf341/corelib/src/lib.cairo#L348)):*

gonna probably just have v0 and than go the v2023_10 (more date based)
</blockquote></details>

:ok_hand: 
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 25 at r1](https://reviewable.io/reviews/starkware-libs/cairo/4301#-NhaUi8_03vEYudlt9Tw:-NhkPQP156iIck_t3jIf:btcpymm) ([raw file](https://github.com/starkware-libs/cairo/blob/732938ed247e1f213582dd199bf6a0ff66ddf341/crates/cairo-lang-filesystem/src/db.rs#L25)):*

Hmm - changed to version - and made prelude a simple member function.
</blockquote></details>

I like this :100:

mkaput (approved):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/4301)* status: all files reviewed, 2 unresolved discussions (waiting on @orizi and @yuvalsw)

___
*[`crates/cairo-lang-filesystem/src/db.rs` line 38 at r5](https://reviewable.io/reviews/starkware-libs/cairo/4301#-NhkQ9uv3mTA_rc3t9Wq:-NhkQ9uv3mTA_rc3t9Wr:b-cg5829) ([raw file](https://github.com/starkware-libs/cairo/blob/8e3a76887ca5146f099b065e42b6f1ef2a7ec023/crates/cairo-lang-filesystem/src/db.rs#L38)):*
> ```XML
> /// major compiler changes without breaking existing crates. On major compiler updates, old
> /// compatibility versions may be removed, and default may be updated to a newer version.
> #[derive(Clone, Debug, Default, Hash, PartialEq, Eq)]
> ```

I honestly don't know why this couldn't be `Copy`.



_Suggestion:_
```XML
#[derive(Copy, Clone, Debug, Default, Hash, PartialEq, Eq)]
```

yuvalsw (changes requested):
> Reviewed 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/4301)* status: all files reviewed, 1 unresolved discussion (waiting on @orizi)

___
*[`corelib/src/felt_252.cairo` line 1 at r5](https://reviewable.io/reviews/starkware-libs/cairo/4301#-NhhFtaIDRrHG8I_FVpK:-NhuXmDVBV5gvHK3jbrF:byfphti) ([raw file](https://github.com/starkware-libs/cairo/blob/8e3a76887ca5146f099b065e42b6f1ef2a7ec023/corelib/src/felt_252.cairo#L1)):*

No code was released with this.
The base prelude matches the last release.
</blockquote></details>

I get it, I just wondered why you needed to add it, and if the same reason may be applicable to other user code.

orizi (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/4301)* status: all files reviewed, 1 unresolved discussion (waiting on @yuvalsw)

___
*[`corelib/src/felt_252.cairo` line 1 at r5](https://reviewable.io/reviews/starkware-libs/cairo/4301#-NhhFtaIDRrHG8I_FVpK:-Nhuip1V9KmYOEJd9mEi:be2creg) ([raw file](https://github.com/starkware-libs/cairo/blob/8e3a76887ca5146f099b065e42b6f1ef2a7ec023/corelib/src/felt_252.cairo#L1)):*

I get it, I just wondered why you needed to add it, and if the same reason may be applicable to other user code.
</blockquote></details>

the idea is that corelib is no longer "special" (in the lookup sense) only the prelude is "special" (and unavoidable) - so everything not in prelude must include `core::` - as this is the actual full name.

yuvalsw (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/4301)* status: all files reviewed, 2 unresolved discussions (waiting on @orizi)

___
*[`crates/cairo-lang-filesystem/src/db.rs` line 39 at r6](https://reviewable.io/reviews/starkware-libs/cairo/4301#-Nhvw9jK0dbK_MSNn8PP:-Nhvw9jK0dbK_MSNn8PQ:b-kidgrs) ([raw file](https://github.com/starkware-libs/cairo/blob/c63c03cbab76a23247d4b9c55cffdb9ebd04a301/crates/cairo-lang-filesystem/src/db.rs#L39)):*
> ```XML
> /// compatibility versions may be removed, and default may be updated to a newer version.
> #[derive(Clone, Copy, Debug, Default, Hash, PartialEq, Eq)]
> pub enum CompatibilityVersion {
> ```

Consider rename to CompilerVersion. This is used for compatibilty, but is not the only thing it can be used to. It just represents a compiler version. Also CompatibilityVersion is confusing to me.
___
*[`crates/cairo-lang-semantic/src/corelib.rs` line 55 at r6](https://reviewable.io/reviews/starkware-libs/cairo/4301#-NhvwdRC2D9qqUQnJB_E:-NhvwdRC2D9qqUQnJB_F:b-ovcg5c) ([raw file](https://github.com/starkware-libs/cairo/blob/c63c03cbab76a23247d4b9c55cffdb9ebd04a301/crates/cairo-lang-semantic/src/corelib.rs#L55)):*
> ```XML
> /// If no such submodule exists, panics.
> pub fn core_submodule(db: &dyn SemanticGroup, submodule_name: &str) -> ModuleId {
>     get_submodule(db, core_module(db), submodule_name).expect("Requested core submodule not found")
> ```

I know it wasn't here in the first place, but please add the submodule name to the panic message. Thanks



_Code quote:_
```XML
"Requested core submodule not found"
```

orizi (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/4301)* status: all files reviewed, 2 unresolved discussions (waiting on @yuvalsw)

___
*[`crates/cairo-lang-filesystem/src/db.rs` line 39 at r6](https://reviewable.io/reviews/starkware-libs/cairo/4301#-Nhvw9jK0dbK_MSNn8PP:-Nhw211_9GORvviHhygt:bjf4m4n) ([raw file](https://github.com/starkware-libs/cairo/blob/575db89941a5f5a890dc4891f1ac8a4bdabab391/crates/cairo-lang-filesystem/src/db.rs#L39)):*

Consider rename to CompilerVersion. This is used for compatibilty, but is not the only thing it can be used to. It just represents a compiler version. Also CompatibilityVersion is confusing to me.
</blockquote></details>

it isn't the compiler version - we already have a compiler version based on the actual crate config.
It should only be used for compatibility as this is the only sort of thing that makes sense to do per crate.
___
*[`crates/cairo-lang-semantic/src/corelib.rs` line 55 at r6](https://reviewable.io/reviews/starkware-libs/cairo/4301#-NhvwdRC2D9qqUQnJB_E:-Nhw2r_t4lYki8ZeUu3E:b-896fix) ([raw file](https://github.com/starkware-libs/cairo/blob/575db89941a5f5a890dc4891f1ac8a4bdabab391/crates/cairo-lang-semantic/src/corelib.rs#L55)):*

I know it wasn't here in the first place, but please add the submodule name to the panic message. Thanks
</blockquote></details>

Done.

mkaput (approved):
> Reviewed 1 of 1 files at r6, 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/4301)* status: all files reviewed, 2 unresolved discussions (waiting on @yuvalsw)

___
*[`crates/cairo-lang-filesystem/src/db.rs` line 39 at r6](https://reviewable.io/reviews/starkware-libs/cairo/4301#-Nhvw9jK0dbK_MSNn8PP:-NhzpkqE7eE90Se79aKt:b-12lwyv) ([raw file](https://github.com/starkware-libs/cairo/blob/575db89941a5f5a890dc4891f1ac8a4bdabab391/crates/cairo-lang-filesystem/src/db.rs#L39)):*

it isn't the compiler version - we already have a compiler version based on the actual crate config.
It should only be used for compatibility as this is the only sort of thing that makes sense to do per crate.
</blockquote></details>

Alternatively, what about stealing Rust's `edition` term? :upside_down_face:

yuvalsw (changes requested):
> Reviewed 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/4301)* status: all files reviewed, 1 unresolved discussion (waiting on @mkaput and @orizi)

___
*[`crates/cairo-lang-filesystem/src/db.rs` line 39 at r6](https://reviewable.io/reviews/starkware-libs/cairo/4301#-Nhvw9jK0dbK_MSNn8PP:-NhzsL91EjS5zTjVe-yC:b-dobzoq) ([raw file](https://github.com/starkware-libs/cairo/blob/575db89941a5f5a890dc4891f1ac8a4bdabab391/crates/cairo-lang-filesystem/src/db.rs#L39)):*

Alternatively, what about stealing Rust's `edition` term? :upside_down_face:
</blockquote></details>

The doc says "Allows configuring a crate to be compatible with an older version of the compiler". CompilerVersion seems right according to this. Maybe change to "with an older version of the compiler configuration" and CompilerConfigVersion?
Edition may also be OK, but it's not currently equivalent, so we would need to clarify this/ think this through as it might be understood to be the same by people who know it from Rust.
IMO we should have CompilerConfigVersion for now, and maybe later when we fully understand the details of how we leverage it, we can rename it if it makes sense.

mkaput (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/4301)* status: all files reviewed, 1 unresolved discussion (waiting on @yuvalsw)

___
*[`crates/cairo-lang-filesystem/src/db.rs` line 39 at r6](https://reviewable.io/reviews/starkware-libs/cairo/4301#-Nhvw9jK0dbK_MSNn8PP:-Ni-1PtO1liHe0FlQ1Th:b7ku15t) ([raw file](https://github.com/starkware-libs/cairo/blob/575db89941a5f5a890dc4891f1ac8a4bdabab391/crates/cairo-lang-filesystem/src/db.rs#L39)):*

The doc says "Allows configuring a crate to be compatible with an older version of the compiler". CompilerVersion seems right according to this. Maybe change to "with an older version of the compiler configuration" and CompilerConfigVersion?
Edition may also be OK, but it's not currently equivalent, so we would need to clarify this/ think this through as it might be understood to be the same by people who know it from Rust.
IMO we should have CompilerConfigVersion for now, and maybe later when we fully understand the details of how we leverage it, we can rename it if it makes sense.
</blockquote></details>

Oh, I just realised something. Scarb has the [`cairo-version`](https://docs.swmansion.com/scarb/docs/reference/manifest.html#cairo-version) field, which states what version of Cairo does this package support (it's semver range).

Go has recently started adopting a similar pattern to ours and Rust's, in that they introduce breaking changes per-module (package in our terms). They are simply using the statement from `go.mod` which tells what version of Go this module is written in. For reference: https://github.com/golang/go/wiki/LoopvarExperiment#if-the-proposal-is-accepted-how-will-the-change-be-deployed

**Perhaps we could re-use the existing mechanism in Go-fashion here: if a package specifies `cairo-version` which matches Cairo 2.4.0, but not 2.3.0, we enable the new compatibility version; otherwise we fall back to V0.**

Implementation-wise I think we could save this enum under whatever name we choose, but I'd be keen to keep the function which does version matching here, rather than in Scarb. Opinions?

orizi (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/4301)* status: all files reviewed, 1 unresolved discussion (waiting on @mkaput and @yuvalsw)

___
*[`crates/cairo-lang-filesystem/src/db.rs` line 39 at r6](https://reviewable.io/reviews/starkware-libs/cairo/4301#-Nhvw9jK0dbK_MSNn8PP:-NhzvFCdEZsI0h6nSVOM:b3n06ti) ([raw file](https://github.com/starkware-libs/cairo/blob/575db89941a5f5a890dc4891f1ac8a4bdabab391/crates/cairo-lang-filesystem/src/db.rs#L39)):*

Oh, I just realised something. Scarb has the [`cairo-version`](https://docs.swmansion.com/scarb/docs/reference/manifest.html#cairo-version) field, which states what version of Cairo does this package support (it's semver range).

Go has recently started adopting a similar pattern to ours and Rust's, in that they introduce breaking changes per-module (package in our terms). They are simply using the statement from `go.mod` which tells what version of Go this module is written in. For reference: https://github.com/golang/go/wiki/LoopvarExperiment#if-the-proposal-is-accepted-how-will-the-change-be-deployed

**Perhaps we could re-use the existing mechanism in Go-fashion here: if a package specifies `cairo-version` which matches Cairo 2.4.0, but not 2.3.0, we enable the new compatibility version; otherwise we fall back to V0.**

Implementation-wise I think we could save this enum under whatever name we choose, but I'd be keen to keep the function which does version matching here, rather than in Scarb. Opinions?
</blockquote></details>

i like the `edition` - and it is very similar to rust, so would be less confusing.
I like the configuration of the crate being only in the toml and not mixing it with the `mod.cairo` of the project.

mkaput (approved):
:lgtm:

> Reviewed 3 of 3 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/4301#-:-Ni0IVON5wPXhJk-LAq7:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @yuvalsw)

mkaput (approved):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/4301)* status: all files reviewed, 2 unresolved discussions (waiting on @orizi and @yuvalsw)

___
*[`crates/cairo-lang-filesystem/src/db.rs` line 42 at r8](https://reviewable.io/reviews/starkware-libs/cairo/4301#-Ni0IjQFCQ3X3YvbfGSC:-Ni0IjQFCQ3X3YvbfGSD:b-tkr3qy) ([raw file](https://github.com/starkware-libs/cairo/blob/952a80738bd43c14210dc8d5aba3eca9a4e3242d/crates/cairo-lang-filesystem/src/db.rs#L42)):*
> ```XML
>     /// The base edition.
>     #[default]
>     V0,
> ```

What do you think about about `2023_01` - date of first release? This way all values would be uniform here.

yuvalsw (changes requested):
> Reviewed 3 of 3 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/4301)* status: all files reviewed, 4 unresolved discussions (waiting on @mkaput and @orizi)

___
*[`crates/cairo-lang-filesystem/src/db.rs` line 24 at r8](https://reviewable.io/reviews/starkware-libs/cairo/4301#-Ni40Vd04QCQInNpRSn7:-Ni40Vd04QCQInNpRSn8:undefined) ([raw file](https://github.com/starkware-libs/cairo/blob/952a80738bd43c14210dc8d5aba3eca9a4e3242d/crates/cairo-lang-filesystem/src/db.rs#L24)):*
> ```XML
>     /// The root directry of the crate.
>     pub root: Directory,
>     /// The cairo edition of the trait.
> ```





_Suggestion:_
```XML
he crate.
```
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 34 at r8](https://reviewable.io/reviews/starkware-libs/cairo/4301#-Ni416XcCnF5GviTALz8:-Ni416XcCnF5GviTALz9:byzu2xe) ([raw file](https://github.com/starkware-libs/cairo/blob/952a80738bd43c14210dc8d5aba3eca9a4e3242d/crates/cairo-lang-filesystem/src/db.rs#L34)):*
> ```XML
> }
> 
> /// The Cairo edition of a crate.
> ```

The term Edition needs more clarification. Especially the diff and the relation between version (e.g. 2.3.0) and edition.
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 35 at r8](https://reviewable.io/reviews/starkware-libs/cairo/4301#-Ni40cooB9J0TX7PdQK_:-Ni40cooB9J0TX7PdQKa:undefined) ([raw file](https://github.com/starkware-libs/cairo/blob/952a80738bd43c14210dc8d5aba3eca9a4e3242d/crates/cairo-lang-filesystem/src/db.rs#L35)):*
> ```XML
> 
> /// The Cairo edition of a crate.
> /// Allows configuring a crate to be compatible with an older editions of the compiler, to enable
> ```





_Suggestion:_
```XML
e with older editions of the compiler
```
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 42 at r8](https://reviewable.io/reviews/starkware-libs/cairo/4301#-Ni0IjQFCQ3X3YvbfGSC:-Ni40u487w4qZhsJXEE4:b-k2qv4d) ([raw file](https://github.com/starkware-libs/cairo/blob/952a80738bd43c14210dc8d5aba3eca9a4e3242d/crates/cairo-lang-filesystem/src/db.rs#L42)):*

What do you think about about `2023_01` - date of first release? This way all values would be uniform here.
</blockquote></details>

+1
Let's also include the day, in case we have 2 future releases in the same month, and for future uniformity.

mkaput (approved):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/4301)* status: all files reviewed, 4 unresolved discussions (waiting on @orizi)

orizi (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/4301)* status: all files reviewed, 3 unresolved discussions (waiting on @mkaput and @yuvalsw)

___
*[`crates/cairo-lang-filesystem/src/db.rs` line 24 at r8](https://reviewable.io/reviews/starkware-libs/cairo/4301#-Ni40Vd04QCQInNpRSn7:-Ni8amAw0JlruX2HUsMF:b-896fix) ([raw file](https://github.com/starkware-libs/cairo/blob/952a80738bd43c14210dc8d5aba3eca9a4e3242d/crates/cairo-lang-filesystem/src/db.rs#L24)):*
> ```XML
>     /// The root directry of the crate.
>     pub root: Directory,
>     /// The cairo edition of the trait.
> ```

Done.
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 34 at r8](https://reviewable.io/reviews/starkware-libs/cairo/4301#-Ni416XcCnF5GviTALz8:-Ni8bPM27KwGfWtuuIfv:b-896fix) ([raw file](https://github.com/starkware-libs/cairo/blob/952a80738bd43c14210dc8d5aba3eca9a4e3242d/crates/cairo-lang-filesystem/src/db.rs#L34)):*

The term Edition needs more clarification. Especially the diff and the relation between version (e.g. 2.3.0) and edition.
</blockquote></details>

Done.
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 42 at r8](https://reviewable.io/reviews/starkware-libs/cairo/4301#-Ni0IjQFCQ3X3YvbfGSC:-Ni8aD_F9snAsp1FXK3V:bniy9t9) ([raw file](https://github.com/starkware-libs/cairo/blob/952a80738bd43c14210dc8d5aba3eca9a4e3242d/crates/cairo-lang-filesystem/src/db.rs#L42)):*

+1
Let's also include the day, in case we have 2 future releases in the same month, and for future uniformity.
</blockquote></details>

it is unlikely we have several editions in the same month, and i think i'd rather have the a little bit shorter name.

yuvalsw (approved):
:lgtm:

> Reviewed 3 of 3 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/4301#-:-Ni94uHY2i3Ms7P1WFjq:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @mkaput and @orizi)

___
*[`crates/cairo-lang-filesystem/src/db.rs` line 42 at r8](https://reviewable.io/reviews/starkware-libs/cairo/4301#-Ni0IjQFCQ3X3YvbfGSC:-Ni94Y3F66ButJ4_iriu:bsy9pbw) ([raw file](https://github.com/starkware-libs/cairo/blob/952a80738bd43c14210dc8d5aba3eca9a4e3242d/crates/cairo-lang-filesystem/src/db.rs#L42)):*

it is unlikely we have several editions in the same month, and i think i'd rather have the a little bit shorter name.
</blockquote></details>

I don't think it's unlikely. I'd actually bet we would have it sometime.
Anyway, we can then have an extra "_2" or something.

orizi (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/4301)* status: all files reviewed, 1 unresolved discussion (waiting on @mkaput and @yuvalsw)

___
*[`crates/cairo-lang-filesystem/src/db.rs` line 42 at r8](https://reviewable.io/reviews/starkware-libs/cairo/4301#-Ni0IjQFCQ3X3YvbfGSC:-Ni9SdL3BJacLAILA5dN:bgka21k) ([raw file](https://github.com/starkware-libs/cairo/blob/952a80738bd43c14210dc8d5aba3eca9a4e3242d/crates/cairo-lang-filesystem/src/db.rs#L42)):*

I don't think it's unlikely. I'd actually bet we would have it sometime.
Anyway, we can then have an extra "_2" or something.
</blockquote></details>

exactly - it would probably happen at some point, but i'd rather have this for a short name.

PR #4382, opened 2023-11-09
Make `IMPLCIT_PRECEDENCE` public

piotmag769:
The reason for this change is that we use this constant in Starknet Foundry for test compilation, this way it will be easier to stay up to date with compiler. Now we need to manually check if it did not change

- - -
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/4382)

### DIFF
--- a/crates/cairo-lang-starknet/src/plugin/consts.rs
+++ b/crates/cairo-lang-starknet/src/plugin/consts.rs
@@ -50,7 +50,7 @@ pub(super) const L1_HANDLER_FIRST_PARAM_NAME: &str = "from_address";
 pub(super) const CALLDATA_PARAM_NAME: &str = "__calldata__";
 
 /// Starknet OS required implicit precedence.
-pub(super) const IMPLICIT_PRECEDENCE: &[&str] = &[
+pub const IMPLICIT_PRECEDENCE: &[&str] = &[
     "Pedersen",
     "RangeCheck",
     "Bitwise",


orizi (commented):
how are you using this constant?
the main idea of this one is the ordering - it is caused by the definition of starknet.
Is it possible that one of the more starknet specific crates somhow contains this data?
`starknet_api` or similar?
(because i would have like if this wouldn't have been here as well)

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/4382#-:-Ninz2Q79voVnY5zO_HV:b-kg528d)* status: 0 of 1 files reviewed, all discussions resolved

piotmag769 (commented):
I didn't find it in `starknet_api` nor `starknet-rs` :/ We use it to filter test function arguments to get the ones provided by a user (useful for fuzzing)

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/4382#-:-Nio6qp18ghA44UU2Wee:bnkphyz)* status: 0 of 1 files reviewed, all discussions resolved

orizi (commented):
I'm still not entirly sure how exactly this is used - but if it is for a list of types - this is definately not the correct place, we just need to enforce the ordering required when using starknet, rather not support it for external usages.

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/4382#-:-NioZNzS--RGoe08L7PF:b-7wtgpt)* status: 0 of 1 files reviewed, all discussions resolved

PR #4671, opened 2023-12-28
Added *by_id queries

yuvalsw:
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/4671)


orizi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/4671)* status: 0 of 12 files reviewed, 1 unresolved discussion (waiting on @yuvalsw)

*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/4671#-Nmk0AqM95VSgFDqVenc:-Nmk0AqNAixPqe-eIe6G:b-x7zm86) (no related file):*
Can the addition of these queries be a bit much to continue contributing to how large the db mem signature actually is?

___

yuvalsw:

Can the addition of these queries be a bit much to continue contributing to how large the db mem signature actually is?
</blockquote></details>

WDYM? Is it affecting any performance?
Adding them caches more results, that should improve the performance.
If the mem signature affects something, which I wasn't aware is an issue so far - some of them can be probably removed as they are used only once anyway, but this would be nit picking, and assuming it will remain this way in the future

orizi (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/4671)* status: 0 of 12 files reviewed, 1 unresolved discussion (waiting on @yuvalsw)

*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/4671#-Nmk0AqM95VSgFDqVenc:-NmksfTh9CxjgBJ6emDz:b-iaph8q) (no related file):*

WDYM? Is it affecting any performance?
Adding them caches more results, that should improve the performance.
If the mem signature affects something, which I wasn't aware is an issue so far - some of them can be probably removed as they are used only once anyway, but this would be nit picking, and assuming it will remain this way in the future
</blockquote></details>

caching isn't free - we needed to do some work on the language server to reduce the memory footprint somewhat.
performance wise this should be not particularly helpful as this is already only a simple hash access.
Anyway unblocking.

___

orizi (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/4671#-:-Nmkt8sD9lTQU5qnqksa:bnfp4nl)* status: 0 of 12 files reviewed, 1 unresolved discussion (waiting on @yuvalsw)

PR #4690, opened 2023-12-29
Add standalone conditions

wangshouh:
In [dojo-starter](https://github.com/dojoengine/dojo-starter) template, dojo is configured as follows:

```json
{
    "cairo1.languageServerPath": "$HOME/.dojo/bin/dojo-language-server",
    "cairo1.enableLanguageServer": true,
    "cairo1.enableScarb": true
}
```

But this configuration will cause cairo-vscode to start the LSP inside the scarb. The following shows the Cairo extension log:

```
Using Scarb binary from: /root/.local/bin/scarb
Cairo language server running from Scarb at: /root/.local/bin/scarb
```

This is not in line with developer intuition. Generally speaking, the configuration specified by the developer should override the default configuration. So I modified the `getServerType` function to make vscode-cairo more intuitive to users. The following shows the modified cairo extension log:

```
Using Scarb binary from: /root/.local/bin/scarb
Cairo language server running from: /root/.dojo/bin/dojo-language-server
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/4690)

### DIFF
--- a/vscode-cairo/src/extension.ts
+++ b/vscode-cairo/src/extension.ts
@@ -245,6 +245,9 @@ async function getServerType(
   context: vscode.ExtensionContext
 ) {
   if (!isScarbEnabled) return ServerType.Standalone;
+  if (isScarbEnabled && !!configLanguageServerPath) {
+    return ServerType.Standalone
+  }
   if (!(await isScarbProject()) && !!configLanguageServerPath) {
     // If Scarb manifest is missing, and Cairo-LS path is explicit.
     return ServerType.Standalone;


orizi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/4690)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @wangshouh)

*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/4690#-Nn3f3IX2K90f1yQm3VK:-Nn3f3IX2K90f1yQm3VL:bt3yvuv) (no related file):*
@maciektr 

___

maciektr:
Why not simply toggle the `enableScarb` flag then? I don't think I follow why does your condition require `enableScarb` flag to be true? :thinking: 

IMHO the current logic makes sense. It runs from Scarb for Scarb based projects, and from standalone for non Scarb projects. This simplifies corelib management for end users (as user can rely on Scarb to provide corelib for Scarb based projects, but need to download and manage it manually for non-Scarb ones). If user wants to override this, it can be done with `enableScarb` flag. 
 
Let me know wdyt @wangshouh

mkaput (commented):
Yeah, I think the real issue here is that `enableScarb` name conveys a slightly different meaning than what actually happens under the hood. I think it should be named differently, like `preferScarbLanguageServer`.

So @wangshouh, I recommend you to change dojo template to set `enableScarb` to `false`, and on our side; we'll push efforts to clarify this config property. 

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/4690#-:-NnFFM1g3fWmNTJLHjsj:b-9p819h)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @orizi and @wangshouh)

wangshouh:
> Why not simply toggle the `enableScarb` flag then? I don't think I follow why does your condition require `enableScarb` flag to be true? 🤔
> 
> IMHO the current logic makes sense. It runs from Scarb for Scarb based projects, and from standalone for non Scarb projects. This simplifies corelib management for end users (as user can rely on Scarb to provide corelib for Scarb based projects, but need to download and manage it manually for non-Scarb ones). If user wants to override this, it can be done with `enableScarb` flag.
> 
> Let me know wdyt @wangshouh

Thank you very much for your comment. The important reason why I modified this part of the code is that I think it is correct for the dojo template to use `"cairo1.enableScarb": true`, so I think there is a problem with the expansion code of cairo vsocode, but I tried` "cairo1.enableScarb":false` configuration, this is valid.
To sum up, I think my modifications are invalid.

PR #5180, opened 2024-03-01
Add `ErrorCode` to `DiagnosticEntry`

mkaput:
**Stack**:
- #5196
- #5181
- #5180 ⬅

- - -
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/5180)

### DIFF
--- a/crates/cairo-lang-diagnostics/src/diagnostics.rs
+++ b/crates/cairo-lang-diagnostics/src/diagnostics.rs
@@ -1,8 +1,5 @@
-#[cfg(test)]
-#[path = "diagnostics_test.rs"]
-mod test;
-
-use core::fmt;
+use std::fmt;
+use std::hash::Hash;
 use std::sync::Arc;
 
 use cairo_lang_debug::debug::DebugWithDb;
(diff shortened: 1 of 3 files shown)


orizi (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/5180#-:-Ns-nx0E2ls0H0JKl-IH:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @mkaput)

LucasLvy:
WDYT of linking the error message to the diagnostic kind ?

mkaput:
> WDYT of linking the error message to the diagnostic kind ?

You mean linking the error code to the message, so that one code always outputs the same text? I thought about this, and I find this may end up quite limiting:
1. For example, in some cases, the compiler might be able to produce extra help that it can use to augment the message in particular scenarios.
2. Or, in the parser tests, there are multiple error kinds that boil down to the following error: _I expected this token/grammar production: XYZ_.
3. Finally, I guess there will be errors for which we will not assign error codes at all. Rust has such cases, for example, it has no codes for syntax errors.

I believe what could be much more beneficial, is to add an error codebook ([example](https://doc.rust-lang.org/error_codes/E0001.html)) and the `--explain` feature that Rust has.

mkaput:
I'll remove the `E` vs `W` differentiation and stick with `E` alone because in runtime, user might want to make all compiler warnings an error etc.

LucasLvy:
Linking in the sense having a function `to_error_code` on the diagnostic kind

orizi (approved):
> Reviewed 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/5180)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @mkaput)

mkaput:
> Linking in the sense having a function `to_error_code` on the diagnostic kind

Ahh okey, now I understand. Yep, I can do this; but diagnostic kinds do not have any particular trait, so this will be a local impl to be added in appropriate PRs.

PR #5440, opened 2024-04-17
Collect functions with executable attributes into a debug info field

maciektr:
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/5440)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -477,6 +477,7 @@ dependencies = [
  "cairo-lang-sierra-generator",
  "cairo-lang-syntax",
  "cairo-lang-utils",
+ "indoc",
  "salsa",
  "smol_str",
  "thiserror",
--- a/crates/cairo-lang-compiler/Cargo.toml
+++ b/crates/cairo-lang-compiler/Cargo.toml
@@ -14,11 +14,12 @@ cairo-lang-filesystem = { path = "../cairo-lang-filesystem", version = "~2.6.3" 
 cairo-lang-lowering = { path = "../cairo-lang-lowering", version = "~2.6.3" }
 cairo-lang-parser = { path = "../cairo-lang-parser", version = "~2.6.3" }
 cairo-lang-project = { path = "../cairo-lang-project", version = "~2.6.3" }
-cairo-lang-semantic = { path = "../cairo-lang-semantic", version = "~2.6.3" }
+cairo-lang-semantic = { path = "../cairo-lang-semantic", version = "~2.6.3", features = ["testing"] }
 cairo-lang-sierra = { path = "../cairo-lang-sierra", version = "~2.6.3" }
 cairo-lang-sierra-generator = { path = "../cairo-lang-sierra-generator", version = "~2.6.3" }
 cairo-lang-syntax = { path = "../cairo-lang-syntax", version = "~2.6.3" }
 cairo-lang-utils = { path = "../cairo-lang-utils", version = "~2.6.3" }
+indoc.workspace = true
 salsa.workspace = true
 smol_str.workspace = true
 thiserror.workspace = true
--- a/crates/cairo-lang-compiler/src/lib.rs
+++ b/crates/cairo-lang-compiler/src/lib.rs
@@ -11,6 +11,7 @@ use cairo_lang_filesystem::ids::CrateId;
 use cairo_lang_sierra::debug_info::{Annotations, DebugInfo};
 use cairo_lang_sierra::program::{Program, ProgramArtifact};
 use cairo_lang_sierra_generator::db::SierraGenGroup;
+use cairo_lang_sierra_generator::executables::{collect_executables, find_executable_function_ids};
 use cairo_lang_sierra_generator::program_generator::SierraProgramWithDebug;
 use cairo_lang_sierra_generator::replace_ids::replace_sierra_ids_in_program;
 
@@ -21,6 +22,9 @@ use crate::project::{get_main_crate_ids_from_project, setup_project, ProjectConf
 pub mod db;
 pub mod diagnostics;
 pub mod project;
+
+#[cfg(test)]
+mod test;
 
 /// Configuration for the compiler.
 #[derive(Default)]
@@ -148,27 +152,57 @@ pub fn compile_prepared_db(
 pub fn compile_prepared_db_program_artifact(
     db: &mut RootDatabase,
     main_crate_ids: Vec<CrateId>,
-    compiler_config: CompilerConfig<'_>,
+    mut compiler_config: CompilerConfig<'_>,
 ) -> Result<ProgramArtifact> {
     let add_statements_functions = compiler_config.add_statements_functions;
 
-    let sierra_program_with_debug = compile_prepared_db(db, main_crate_ids, compiler_config)?;
-    let mut program_artifact = ProgramArtifact::stripped(sierra_program_with_debug.program);
+    compiler_config.diagnostics_reporter.ensure(db)?;
 
+    let executable_functions = find_executable_function_ids(db, main_crate_ids.clone());
+
+    let mut sierra_program_with_debug = if executable_functions.is_empty() {
+        // No executables found - compile for all main crates.
+        // TODO(maciektr): Deprecate in future. This compilation is useless, without `replace_ids`.
+        Arc::unwrap_or_clone(
+            db.get_sierra_program(main_crate_ids)
+                .to_option()
+                .context("Compilation failed without any diagnostics")?,
+        )
+    } else {
+        // Compile for executable functions only.
+        Arc::unwrap_or_clone(
+            db.get_sierra_program_for_functions(executable_functions.clone().into_keys().collect())
+                .to_option()
+                .context("Compilation failed without any diagnostics")?,
+        )
+    };
+
+    if compiler_config.replace_ids {
+        sierra_program_with_debug.program =
+            replace_sierra_ids_in_program(db, &sierra_program_with_debug.program);
+    }
+
+    // Calculate debug info.
+    let mut debug_info = DebugInfo::default();
     if add_statements_functions {
         let statements_functions = sierra_program_with_debug
             .debug_info
             .statements_locations
             .extract_statements_functions(db);
 
-        let debug_info = DebugInfo {
+        debug_info = DebugInfo {
             type_names: Default::default(),
             libfunc_names: Default::default(),
             user_func_names: Default::default(),
             annotations: Annotations::from(statements_functions),
+            executables: Default::default(),
         };
-        program_artifact = program_artifact.with_debug_info(debug_info);
     }
 
-    Ok(program_artifact)
+    // Calculate executable function Sierra ids.
+    let executables =
+        collect_executables(db, executable_functions, &sierra_program_with_debug.program);
+
+    Ok(ProgramArtifact::stripped(sierra_program_with_debug.program)
+        .with_debug_info(DebugInfo { executables, ..debug_info }))
 }
--- a/crates/cairo-lang-compiler/src/test.rs
+++ b/crates/cairo-lang-compiler/src/test.rs
@@ -1,0 +1,59 @@
+use cairo_lang_defs::plugin::{MacroPlugin, MacroPluginMetadata, PluginResult};
+use cairo_lang_semantic::plugin::PluginSuite;
+use cairo_lang_semantic::test_utils::setup_test_crate;
+use cairo_lang_syntax::node::ast::ModuleItem;
+use cairo_lang_syntax::node::db::SyntaxGroup;
+use indoc::indoc;
+use smol_str::SmolStr;
+
+use crate::db::RootDatabase;
+use crate::{compile_prepared_db_program_artifact, CompilerConfig};
+
+#[derive(Debug, Default)]
+pub struct MockExecutablePlugin {}
+
+impl MacroPlugin for MockExecutablePlugin {
+    fn generate_code(
+        &self,
+        _db: &dyn SyntaxGroup,
+        _item_ast: ModuleItem,
+        _metadata: &MacroPluginMetadata<'_>,
+    ) -> PluginResult {
+        PluginResult { code: None, diagnostics: vec![], remove_original_item: false }
+    }
+
+    fn declared_attributes(&self) -> Vec<String> {
+        vec!["some".to_string()]
+    }
+
+    fn executable_attributes(&self) -> Vec<String> {
+        vec!["some".to_string()]
+    }
+}
+
+#[test]
+fn can_collect_executables() {
+    let content = indoc! {r#"
+        fn x() -> felt252 { 12 }
+
+        fn main() -> felt252 { x() }
+
+        #[some]
+        fn test() -> felt252 { x() }
+    "#};
+    let mut suite = PluginSuite::default();
+    suite.add_plugin::<MockExecutablePlugin>();
+    let mut db = RootDatabase::builder().detect_corelib().with_plugin_suite(suite).build().unwrap();
+    let crate_id = setup_test_crate(&db, content);
+    let config = CompilerConfig { replace_ids: true, ..CompilerConfig::default() };
+    let artefact = compile_prepared_db_program_artifact(&mut db, vec![crate_id], config).unwrap();
+    let executables = artefact.debug_info.unwrap().executables;
+    assert!(!executables.is_empty());
+    let f_ids = executables.get("some").unwrap();
+    assert_eq!(f_ids.len(), 1);
+    // Assert executable name
+    assert_eq!(f_ids[0].debug_name, Some(SmolStr::new("test::test")));
+    // Assert only executable functions are compiled.
+    assert_eq!(artefact.program.funcs.len(), 1);
+    assert_eq!(artefact.program.funcs[0].id.debug_name, Some(SmolStr::new("test::test")));
+}
--- a/crates/cairo-lang-defs/src/plugin.rs
+++ b/crates/cairo-lang-defs/src/plugin.rs
@@ -102,6 +102,15 @@ pub trait MacroPlugin: std::fmt::Debug + Sync + Send {
     /// Note: They may not cause a diagnostic if some other plugin declares such attribute, but
     /// plugin writers should not rely on that.
     fn declared_attributes(&self) -> Vec<String>;
+
+    /// Attributes that should mark the function as an executable.
+    /// Functions marked with executable attributes will be listed
+    /// in a dedicated field in the generated program.
+    /// Must return a subset of `declared_attributes`.
+    /// This mechanism is optional.
+    fn executable_attributes(&self) -> Vec<String> {
+        Vec::new()
+    }
 }
 
 /// Result of plugin code generation.
--- a/crates/cairo-lang-sierra-generator/src/executables.rs
+++ b/crates/cairo-lang-sierra-generator/src/executables.rs
@@ -1,0 +1,111 @@
+use std::collections::HashMap;
+
+use cairo_lang_diagnostics::ToOption;
+use cairo_lang_filesystem::ids::CrateId;
+use cairo_lang_lowering::ids::ConcreteFunctionWithBodyId;
+use cairo_lang_sierra::ids::FunctionId;
+use cairo_lang_sierra::program::Program;
+use cairo_lang_syntax::node::helpers::QueryAttrs;
+use cairo_lang_utils::LookupIntern;
+use smol_str::SmolStr;
+
+use crate::db::SierraGenGroup;
+
+/// Find all functions with executable attributes.
+///
+/// _Executable attributes_ are plugin-backed attributes that declare a function as executable.
+/// You can create an _executable attribute_ by implementing the `executable_attributes` function
+/// of `MacroPlugin` trait.
+/// This function finds all functions in the syntactic model marked with an _executable attribute_,
+/// and returns the attribute names, along the function id.
+/// Note, that a single function can be marked with more than one executable attribute.
+/// Only crates declared as _main_crate_ids_ are considered.
+pub fn find_executable_function_ids(
+    db: &dyn SierraGenGroup,
+    main_crate_ids: Vec<CrateId>,
+) -> HashMap<ConcreteFunctionWithBodyId, Vec<SmolStr>> {
+    let executable_attributes = db
+        .macro_plugins()
+        .iter()
+        .flat_map(|plugin| plugin.executable_attributes())
+        .map(SmolStr::new)
+        .collect::<Vec<_>>();
+    let mut executable_function_ids = HashMap::new();
+    if !executable_attributes.is_empty() {
+        for crate_id in main_crate_ids {
+            for module in db.crate_modules(crate_id).iter() {
+                if let Some(free_functions) = db.module_free_functions(*module).to_option() {
+                    for (free_func_id, body) in free_functions.iter() {
+                        let found_attrs = executable_attributes
+                            .clone()
+                            .iter()
+                            .filter(|attr| body.has_attr(db.upcast(), attr.as_str()))
+                            .cloned()
+                            .collect::<Vec<_>>();
+                        if found_attrs.is_empty() {
+                            // No executable attributes found.
+                            continue;
+                        }
+                        // Find function corresponding to the node by full path.
+                        let function_id = ConcreteFunctionWithBodyId::from_no_generics_free(
+                            db.upcast(),
+                            *free_func_id,
+                        );
+                        if let Some(function_id) = function_id {
+                            executable_function_ids.insert(function_id, found_attrs);
+                        }
+                    }
+                }
+            }
+        }
+    }
+    executable_function_ids
+}
+
+/// Extract Sierra function ids for executable functions based on syntactic function ids.
+///
+/// This functions accepts executable function ids, found with `find_executable_function_ids`,
+/// and finds their corresponding Sierra ids in a Sierra program.
+/// The returned function ids are grouped by the executable attribute name, and sorted by full path.
+pub fn collect_executables(
+    db: &dyn SierraGenGroup,
+    mut executable_function_ids: HashMap<ConcreteFunctionWithBodyId, Vec<SmolStr>>,
+    sierra_program: &Program,
+) -> HashMap<SmolStr, Vec<FunctionId>> {
+    if executable_function_ids.is_empty() {
+        Default::default()
+    } else {
+        let executable_function_ids = executable_function_ids
+            .drain()
+            .filter_map(|(function_id, attrs)| {
+                let function_id = function_id
+                    .function_id(db.upcast())
+                    .to_option()
+                    .map(|function_id| db.intern_sierra_function(function_id));
+                function_id.map(|function_id| (function_id, attrs))
+            })
+            .collect::<HashMap<FunctionId, Vec<SmolStr>>>();
+        let mut result: HashMap<SmolStr, Vec<(String, FunctionId)>> = Default::default();
+        for function in &sierra_program.funcs {
+            let Some(found_attrs) = executable_function_ids.get(&function.id) else {
+                continue;
+            };
+            let full_path = function.id.lookup_intern(db).semantic_full_path(db.upcast());
+            for attr in found_attrs {
+                result
+                    .entry(attr.clone())
+                    .or_default()
+                    .push((full_path.clone(), function.id.clone()));
+            }
+        }
+        // Sort by full path for stability.
+        result
+            .drain()
+            .map(|(key, functions)| {
+                let mut functions = functions.into_iter().collect::<Vec<_>>();
+                functions.sort_by_key(|(full_path, _)| full_path.clone());
+                (key, functions.into_iter().map(|(_, function_id)| function_id).collect::<Vec<_>>())
+            })
+            .collect::<HashMap<SmolStr, Vec<FunctionId>>>()
+    }
+}
--- a/crates/cairo-lang-sierra-generator/src/lib.rs
+++ b/crates/cairo-lang-sierra-generator/src/lib.rs
@@ -8,6 +8,7 @@ mod ap_tracking;
 mod block_generator;
 pub mod canonical_id_replacer;
 pub mod db;
+pub mod executables;
 mod expr_generator_context;
 mod extra_sierra_info;
 mod function_generator;
--- a/crates/cairo-lang-sierra/src/debug_info.rs
+++ b/crates/cairo-lang-sierra/src/debug_info.rs
@@ -14,7 +14,7 @@ use crate::program::{GenericArg, Program, Statement};
 mod test;
 
 /// Debug information for a Sierra program, to get readable names.
-#[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
+#[derive(Clone, Debug, Eq, PartialEq, Default, Serialize, Deserialize)]
 pub struct DebugInfo {
     #[serde(
         serialize_with = "serialize_map::<ConcreteTypeId, _>",
@@ -36,6 +36,9 @@ pub struct DebugInfo {
     /// See [`Annotations`] type documentation for more information about this field.
     #[serde(default, skip_serializing_if = "Annotations::is_empty")]
     pub annotations: Annotations,
+    /// List of functions marked as executable.
+    #[serde(default, skip_serializing_if = "HashMap::is_empty")]
+    pub executables: HashMap<SmolStr, Vec<FunctionId>>,
 }
 
 /// Store for non-crucial information about the program, for use by external libraries and tool.
@@ -90,6 +93,7 @@ impl DebugInfo {
                 })
                 .collect(),
             annotations: Default::default(),
+            executables: Default::default(),
         }
     }
 
--- a/crates/cairo-lang-sierra/src/debug_info_test.rs
+++ b/crates/cairo-lang-sierra/src/debug_info_test.rs
@@ -36,6 +36,7 @@ fn test_extract_names() {
                 ("Func2".into(), "Func2".into())
             ]),
             annotations: Default::default(),
+            executables: Default::default(),
         }
     );
 }
@@ -70,6 +71,7 @@ fn test_populate_names() {
         ]),
         user_func_names: HashMap::from([(0.into(), "Func1".into()), (1.into(), "Func2".into())]),
         annotations: Default::default(),
+        executables: Default::default(),
     }
     .populate(&mut program);
 


orizi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/5440)* status: 0 of 7 files reviewed, 1 unresolved discussion (waiting on @maciektr and @mkaput)

___
*[`crates/cairo-lang-syntax/src/attribute/consts.rs` line 39 at r1](https://reviewable.io/reviews/starkware-libs/cairo/5440#-Nvk9bhn8MvcK9EHawo_:-Nvk9bhn8MvcK9EHawoa:b-hqmlpm) ([raw file](https://github.com/starkware-libs/cairo/blob/da783ff56f7d21725755a8144cce171f843571fa/crates/cairo-lang-syntax/src/attribute/consts.rs#L39)):*
> ```XML
> 
> /// An attribute to mark function as an executable function, meaning a program execution entrypoint.
> pub const EXECUTABLE_ATTR: &str = "executable";
> ```

is a user expected to actually write this attr?
or will it be auto generated by something?

if auto generated - make long and more explicit name.



_Code quote:_
```XML
/// An attribute to mark function as an executable function, meaning a program execution entrypoint.
pub const EXECUTABLE_ATTR: &str = "executable";
```

mkaput (changes requested):
> Reviewed 7 of 7 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/5440)* status: all files reviewed, 3 unresolved discussions (waiting on @maciektr and @orizi)

___
*[`crates/cairo-lang-sierra/src/debug_info.rs` line 41 at r1](https://reviewable.io/reviews/starkware-libs/cairo/5440#-NvkPjOXE6rVBEO36P7U:-NvkPjOXE6rVBEO36P7V:b-b8gxvx) ([raw file](https://github.com/starkware-libs/cairo/blob/da783ff56f7d21725755a8144cce171f843571fa/crates/cairo-lang-sierra/src/debug_info.rs#L41)):*
> ```XML
>     /// List of functions marked as executable.
>     #[serde(default, skip_serializing_if = "Vec::is_empty")]
>     pub executables: Vec<(Function, SmolStr)>,
> ```

We're iterating the structure of this field internally/offline.
___
*[`crates/cairo-lang-sierra-generator/src/executables.rs` line 61 at r1](https://reviewable.io/reviews/starkware-libs/cairo/5440#-NvkPznlE57ycHoOwh86:-NvkPznlE57ycHoOwh87:bdiruet) ([raw file](https://github.com/starkware-libs/cairo/blob/da783ff56f7d21725755a8144cce171f843571fa/crates/cairo-lang-sierra-generator/src/executables.rs#L61)):*
> ```XML
> fn parse_arg(db: &dyn SierraGenGroup, args: Vec<AttributeArg>) -> Option<SmolStr> {
>     args.first().map(|arg| {
>         assert!(matches!(arg.variant, AttributeArgVariant::Unnamed { .. }));
> ```

Emit a diagnostic instead of panicking the compiler.
___
*[`crates/cairo-lang-syntax/src/attribute/consts.rs` line 39 at r1](https://reviewable.io/reviews/starkware-libs/cairo/5440#-Nvk9bhn8MvcK9EHawo_:-NvkPJxPChdhnEtuxmSr:b-novwc2) ([raw file](https://github.com/starkware-libs/cairo/blob/da783ff56f7d21725755a8144cce171f843571fa/crates/cairo-lang-syntax/src/attribute/consts.rs#L39)):*

is a user expected to actually write this attr?
or will it be auto generated by something?

if auto generated - make long and more explicit name.
</blockquote></details>

I imagine rare cases where the user might want to write this manually; but in general, we expect macros to generate these. Say the `#[test]` macro will emit `#[executable("test")]`.

mkaput (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/5440)* status: all files reviewed, 3 unresolved discussions (waiting on @maciektr and @orizi)

*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/5440#-NvkQdAQ3r9iu2waS9QY:-NvkQdAQ3r9iu2waS9QZ:bvvs3fi) (no related file):*
Please add a validation pass somewhere during compilation that will validate arguments of this attribute. You can leave the assertions in `collect_executables` then.

The reason behind this is that we want to emit diagnostics for this attribute even if the toolchain does not call `collect_excecutables` after all.

___
___
*[`crates/cairo-lang-sierra-generator/src/executables.rs` line 61 at r1](https://reviewable.io/reviews/starkware-libs/cairo/5440#-NvkPznlE57ycHoOwh86:-NvkQ_4y9HiYVtf3kjWy:bf6j443) ([raw file](https://github.com/starkware-libs/cairo/blob/da783ff56f7d21725755a8144cce171f843571fa/crates/cairo-lang-sierra-generator/src/executables.rs#L61)):*

Emit a diagnostic instead of panicking the compiler.
</blockquote></details>

Retracting because I have a better idea.

orizi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/5440)* status: all files reviewed, 3 unresolved discussions (waiting on @maciektr and @mkaput)

___
*[`crates/cairo-lang-syntax/src/attribute/consts.rs` line 39 at r1](https://reviewable.io/reviews/starkware-libs/cairo/5440#-Nvk9bhn8MvcK9EHawo_:-NvkW99l23VNJORfBTVi:bfqi4d9) ([raw file](https://github.com/starkware-libs/cairo/blob/da783ff56f7d21725755a8144cce171f843571fa/crates/cairo-lang-syntax/src/attribute/consts.rs#L39)):*

I imagine rare cases where the user might want to write this manually; but in general, we expect macros to generate these. Say the `#[test]` macro will emit `#[executable("test")]`.
</blockquote></details>

so rename to something longer and more cumbersome - for example:
`execution_entrypoint`

if a test will create such a thing - won't compilation recreate a lot of functions?

maciektr:
*[`crates/cairo-lang-syntax/src/attribute/consts.rs` line 39 at r1](https://reviewable.io/reviews/starkware-libs/cairo/5440#-Nvk9bhn8MvcK9EHawo_:-NvkkRDJ6n4dWQfx2e3X:b3gpjw1) ([raw file](https://github.com/starkware-libs/cairo/blob/da783ff56f7d21725755a8144cce171f843571fa/crates/cairo-lang-syntax/src/attribute/consts.rs#L39)):*

so rename to something longer and more cumbersome - for example:
`execution_entrypoint`

if a test will create such a thing - won't compilation recreate a lot of functions?
</blockquote></details>

I wonder whether it would make sense, to use it for marking a main function in `cairo-run` :thinking: In that case, user would need to use it directly.

orizi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/5440)* status: 4 of 7 files reviewed, 4 unresolved discussions (waiting on @maciektr and @mkaput)

*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/5440#-NvkQdAQ3r9iu2waS9QY:-NvktyCd0jdG13nUl0z2:bgrjw3v) (no related file):*

Please add a validation pass somewhere during compilation that will validate arguments of this attribute. You can leave the assertions in `collect_executables` then.

The reason behind this is that we want to emit diagnostics for this attribute even if the toolchain does not call `collect_excecutables` after all.
</blockquote></details>

maybe we can (and should) just add all the 

___
___
*[`crates/cairo-lang-compiler/src/lib.rs` line 150 at r2](https://reviewable.io/reviews/starkware-libs/cairo/5440#-NvkusDMCcQBECmm0JoV:-NvkusDMCcQBECmm0JoW:boa145e) ([raw file](https://github.com/starkware-libs/cairo/blob/ae04675d17452cb183bd74fb81f275f567ce0a44/crates/cairo-lang-compiler/src/lib.rs#L150)):*
> ```XML
> /// * `Err(anyhow::Error)` - Compilation failed.
> pub fn compile_prepared_db_program_artifact(
>     db: &mut RootDatabase,
> ```

i think maybe this function should just not exist.
and similarly to how we call `compile_contract_in_prepared_db` which is contract aware,
and `compile_test_prepared_db` in the test runner - we should call this with a list of the requested artifacts specifically - which would be in the executable list - being plugin based here feels like is going to be bad.



_Code quote:_
```XML
/// * `Ok(ProgramArtifact)` - The compiled program artifact with requested debug info.
/// * `Err(anyhow::Error)` - Compilation failed.
pub fn compile_prepared_db_program_artifact(
    db: &mut RootDatabase,
```

orizi (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/5440)* status: 4 of 7 files reviewed, 4 unresolved discussions (waiting on @maciektr and @mkaput)

*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/5440#-NvkQdAQ3r9iu2waS9QY:-Nvl9xSZ8q3y-oQvNUfN:b-z72pm5) (no related file):*

> maybe we can (and should) just add all the

could you elaborate please?
</blockquote></details>

the idea is - same as test plugin that knows what are the relevant functions (so it can fill the artifacts) and the starknet plugin, so can a run builtin - no need for new special attributes - as it logically per such build target in any case.

___

maciektr:
*[`crates/cairo-lang-compiler/src/lib.rs` line 150 at r2](https://reviewable.io/reviews/starkware-libs/cairo/5440#-NvkusDMCcQBECmm0JoV:-NvlN1SfDmmu91S-6rea:b-1lo195) ([raw file](https://github.com/starkware-libs/cairo/blob/ae04675d17452cb183bd74fb81f275f567ce0a44/crates/cairo-lang-compiler/src/lib.rs#L150)):*

i think maybe this function should just not exist.
and similarly to how we call `compile_contract_in_prepared_db` which is contract aware,
and `compile_test_prepared_db` in the test runner - we should call this with a list of the requested artifacts specifically - which would be in the executable list - being plugin based here feels like is going to be bad.
</blockquote></details>

Sorry, what do you mean by `list of the requested artifacts` :thinking: ? 
The artifact here is just `ProgramArtifact`.

maciektr:

the idea is - same as test plugin that knows what are the relevant functions (so it can fill the artifacts) and the starknet plugin, so can a run builtin - no need for new special attributes - as it logically per such build target in any case.
</blockquote></details>

Well, we would like to push for a generalization of the executable concept here, so that multiple tools (like cairo-run,  sncast-scripts, forge etc.) could rely on a single artifact format (sierra.json), without a need to define specialized targets for each tool in upstream (Scarb).

orizi (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/5440)* status: 4 of 7 files reviewed, 4 unresolved discussions (waiting on @maciektr and @mkaput)

*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/5440#-NvkQdAQ3r9iu2waS9QY:-NvlZIrsD230udAWyLX1:b8vmru0) (no related file):*
<details><summary><i>Previously, maciektr (maciektr) wrote…</i></summary><blockquote>

Well, we would like to push for a generalization of the executable concept here, so that multiple tools (like cairo-run,  sncast-scripts, forge etc.) could rely on a single artifact format (sierra.json), without a need to define specialized targets for each tool in upstream (Scarb).
</blockquote></details>

but this means that things like tests would need to rewrite their code - we probably don't want that as well

___

maciektr:

but this means that things like tests would need to rewrite their code - we probably don't want that as well
</blockquote></details>

Hmm, I am not saying we will push for cairo-test rewrite, but rather that we have tools that are bound to be rewritten already (forge test collection) and would benefit from this. With Scarb procedural macros + executable tag combo, we can run forge tests compiled through Scarb without any additional artefacts -> just `sierra.json`, thus fully decoupling Forge from the Cairo compilation level. Rewriting cairo-run / sncast scripts to this would be not a lot more than a one-liner, while also providing additional advantages (namely, ability to run code compiled in release mode, without debug ids).

orizi (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/5440)* status: 4 of 7 files reviewed, 4 unresolved discussions (waiting on @maciektr and @mkaput)

*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/5440#-NvkQdAQ3r9iu2waS9QY:-NvlsHoL-rPP5RQ7F-ZE:bbiszrw) (no related file):*
<details><summary><i>Previously, maciektr (maciektr) wrote…</i></summary><blockquote>

Hmm, I am not saying we will push for cairo-test rewrite, but rather that we have tools that are bound to be rewritten already (forge test collection) and would benefit from this. With Scarb procedural macros + executable tag combo, we can run forge tests compiled through Scarb without any additional artefacts -> just `sierra.json`, thus fully decoupling Forge from the Cairo compilation level. Rewriting cairo-run / sncast scripts to this would be not a lot more than a one-liner, while also providing additional advantages (namely, ability to run code compiled in release mode, without debug ids).
</blockquote></details>

i was not talking about rewriting the test-plugin - i was talking about all plugins requiring rewriting their code for this to capture it.
the artifacts are a good idea - but the artifacts should IMO be guided by the caller - instead of by annotation that will need to be added in all cases in the world, but would not make sense for any project that contains various such components.

in a project that contains both tests and run - should everything be compiled together and just the external runner should choose something from within? probably not - so there would be a compilation for testing, with its artifacts, and a different one for proof running, or starknet, or anything of that kind.

___

maciektr:
Ok, thanks so much for elaborating, it makes a lot more sense to me now!

> i was talking about all plugins requiring rewriting their code for this to capture it

Well, in case of Forge, which would be the biggest beneficiary of this flow,  we need to have a plugin that rewrites `test` attributes anyway, to move the compilation to Scarb (for full Cairo version <> Forge version decoupling). It would not be an issue there at all. 

> probably not - so there would be a compilation for testing, with its artifacts, and a different one for proof running, or starknet, or anything of that kind.

While it's of course true, that we should not push for a single artefact format for all use cases, I would still argue that there is a middle ground here. We would not want to have dedicated compilation targets, hardcoded into Scarb, for all possible tools that run Cairo (i.e. Forge/Cast/Dojo/etc) either. Some of them will naturally fall into +/- the same requirements category, i.e. finding a function in Sierra program and running it (which this proposition would enable).

maciektr:
> the artifacts should IMO be guided by the caller - instead of by annotation that will need to be added in all cases in the world

Do you think that we should have some per build input here, so we can only collect executables with specific keys? :thinking:

orizi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/5440)* status: 2 of 10 files reviewed, 4 unresolved discussions (waiting on @maciektr and @mkaput)

*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/5440#-NvkQdAQ3r9iu2waS9QY:-NwEPjEU5gxBl4t3j94S:b-plvccn) (no related file):*
<details><summary><i>Previously, maciektr (maciektr) wrote…</i></summary><blockquote>

> the artifacts should IMO be guided by the caller - instead of by annotation that will need to be added in all cases in the world

Do you think that we should have some per build input here, so we can only collect executables with specific keys? :thinking:
</blockquote></details>

i think every build should have some sort of basic plugin - so the build would have some information about possibly required data.
no need for a catch all allowing attr `executable` just for test - just grabbing what is marked with `test`.
for proof - validate extra things on the called function (for example, return type must be `Drop`able) and finding the set of these functions (for a weird project with more than one)

___
___
*[`crates/cairo-lang-compiler/src/lib.rs` line 150 at r2](https://reviewable.io/reviews/starkware-libs/cairo/5440#-NvkusDMCcQBECmm0JoV:-NwEP8Eg9BDYiDmxCRa7:b-3orywj) ([raw file](https://github.com/starkware-libs/cairo/blob/ae04675d17452cb183bd74fb81f275f567ce0a44/crates/cairo-lang-compiler/src/lib.rs#L150)):*
<details><summary><i>Previously, maciektr (maciektr) wrote…</i></summary><blockquote>

Sorry, what do you mean by `list of the requested artifacts` :thinking: ? 
The artifact here is just `ProgramArtifact`.
</blockquote></details>

this function by itself is already weird - it just tells us to drop *everything* into a file - which is not necessary in general.

if you want that i would assume you would want to actually extract the relevant data for testing or for running - so the set of test entrypoints to run, or the set of data required for running in proof mode.

orizi (changes requested):
> Reviewed 1 of 6 files at r3, 5 of 8 files at r4.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/5440)* status: 7 of 11 files reviewed, 3 unresolved discussions (waiting on @maciektr and @mkaput)

___
*[`crates/cairo-lang-compiler/src/lib.rs` line 156 at r4](https://reviewable.io/reviews/starkware-libs/cairo/5440#-NwL7uTYBnC98ay2P3MC:-NwL7uTZ43JIqsSBT_PX:be9n1l5) ([raw file](https://github.com/starkware-libs/cairo/blob/77c499d466459c6df5200da99fe7e0bb90cb5d80/crates/cairo-lang-compiler/src/lib.rs#L156)):*
> ```XML
>     let add_statements_functions = compiler_config.add_statements_functions;
>     let sierra_program_with_debug =
>         compile_prepared_db(db, main_crate_ids.clone(), compiler_config)?;
> ```

This can allow us to also collect only the relevant executable function and dependencies, probably vastly reducing the size of the file

maciektr:
*[`crates/cairo-lang-sierra-generator/src/executables.rs` line 61 at r1](https://reviewable.io/reviews/starkware-libs/cairo/5440#-NvkPznlE57ycHoOwh86:-NwOaNXNBHmhEJ6G1gW2:bpgwvvw) ([raw file](https://github.com/starkware-libs/cairo/blob/da783ff56f7d21725755a8144cce171f843571fa/crates/cairo-lang-sierra-generator/src/executables.rs#L61)):*

Retracting because I have a better idea.
</blockquote></details>

No longer relevant.

maciektr:
*[`crates/cairo-lang-sierra/src/debug_info.rs` line 41 at r1](https://reviewable.io/reviews/starkware-libs/cairo/5440#-NvkPjOXE6rVBEO36P7U:-NwOaTON3fA4hagL1PTU:b-896fix) ([raw file](https://github.com/starkware-libs/cairo/blob/da783ff56f7d21725755a8144cce171f843571fa/crates/cairo-lang-sierra/src/debug_info.rs#L41)):*

We're iterating the structure of this field internally/offline.
</blockquote></details>

Done.

maciektr (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/5440)* status: 7 of 11 files reviewed, 3 unresolved discussions (waiting on @mkaput and @orizi)

___
*[`crates/cairo-lang-compiler/src/lib.rs` line 156 at r4](https://reviewable.io/reviews/starkware-libs/cairo/5440#-NwL7uTYBnC98ay2P3MC:-NwPm-LHB0PJZY1dZmq9:b-896fix) ([raw file](https://github.com/starkware-libs/cairo/blob/77c499d466459c6df5200da99fe7e0bb90cb5d80/crates/cairo-lang-compiler/src/lib.rs#L156)):*

This can allow us to also collect only the relevant executable function and dependencies, probably vastly reducing the size of the file
</blockquote></details>

Done.

orizi (changes requested):
> Reviewed 1 of 8 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/5440)* status: 8 of 11 files reviewed, 6 unresolved discussions (waiting on @maciektr and @mkaput)

___
*[`crates/cairo-lang-sierra-generator/src/executables.rs` line 14 at r5](https://reviewable.io/reviews/starkware-libs/cairo/5440#-NwYeEHjCCX2t1i-JX7c:-NwYeEHjCCX2t1i-JX7d:b2ni1ar) ([raw file](https://github.com/starkware-libs/cairo/blob/3094e8513104b1216328fd8eee112ca90321aa33/crates/cairo-lang-sierra-generator/src/executables.rs#L14)):*
> ```XML
> use crate::db::SierraGenGroup;
> 
> pub fn find_executable_function_ids(
> ```

doc
___
*[`crates/cairo-lang-sierra-generator/src/executables.rs` line 60 at r5](https://reviewable.io/reviews/starkware-libs/cairo/5440#-NwYeMQ6Bzo9k4WB2-Yp:-NwYeMQ6Bzo9k4WB2-Yq:b2ni1ar) ([raw file](https://github.com/starkware-libs/cairo/blob/3094e8513104b1216328fd8eee112ca90321aa33/crates/cairo-lang-sierra-generator/src/executables.rs#L60)):*
> ```XML
> }
> 
> pub fn collect_executables(
> ```

doc
___
*[`crates/cairo-lang-sierra-generator/src/executables.rs` line 92 at r5](https://reviewable.io/reviews/starkware-libs/cairo/5440#-NwYexEi5fQKNP6I7Rba:-NwYexEi5fQKNP6I7Rbb:undefined) ([raw file](https://github.com/starkware-libs/cairo/blob/3094e8513104b1216328fd8eee112ca90321aa33/crates/cairo-lang-sierra-generator/src/executables.rs#L92)):*
> ```XML
>                 }
>             }
>         }
> ```





_Suggestion:_
```XML
        for function in &sierra_program.funcs {
            let function_id = function.id.lookup_intern(db);
            let Some(found_attrs) = executable_function_ids.get(function_id) else {
                continue;
            }
            let full_path = function_id
                .function_with_body_id(db.upcast())
                .base_semantic_function(db.upcast())
                .full_path(db.upcast());

            for attr in found_attrs {
                result
                    .entry(attr.clone())
                    .or_default()
                    .push((full_path.clone(), function.id.clone()));
            }
        }
```

maciektr (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/5440)* status: 8 of 11 files reviewed, 6 unresolved discussions (waiting on @mkaput and @orizi)

___
*[`crates/cairo-lang-sierra-generator/src/executables.rs` line 14 at r5](https://reviewable.io/reviews/starkware-libs/cairo/5440#-NwYeEHjCCX2t1i-JX7c:-Nwcqz9A2-dX6ieJ9wKA:b-896fix) ([raw file](https://github.com/starkware-libs/cairo/blob/3094e8513104b1216328fd8eee112ca90321aa33/crates/cairo-lang-sierra-generator/src/executables.rs#L14)):*

doc
</blockquote></details>

Done.
___
*[`crates/cairo-lang-sierra-generator/src/executables.rs` line 60 at r5](https://reviewable.io/reviews/starkware-libs/cairo/5440#-NwYeMQ6Bzo9k4WB2-Yp:-Nwcqyq28XaVfD2OXvqf:b-896fix) ([raw file](https://github.com/starkware-libs/cairo/blob/3094e8513104b1216328fd8eee112ca90321aa33/crates/cairo-lang-sierra-generator/src/executables.rs#L60)):*

doc
</blockquote></details>

Done.
___
*[`crates/cairo-lang-sierra-generator/src/executables.rs` line 92 at r5](https://reviewable.io/reviews/starkware-libs/cairo/5440#-NwYexEi5fQKNP6I7Rba:-NwcqyRU5vdgI9Wj1STB:b-896fix) ([raw file](https://github.com/starkware-libs/cairo/blob/3094e8513104b1216328fd8eee112ca90321aa33/crates/cairo-lang-sierra-generator/src/executables.rs#L92)):*
> ```XML
>                 }
>             }
>         }
> ```

Done.

mkaput (changes requested):
> Reviewed 1 of 3 files at r2, 1 of 6 files at r3, 6 of 8 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/5440)* status: 9 of 11 files reviewed, 5 unresolved discussions (waiting on @maciektr and @orizi)

___
*[`crates/cairo-lang-defs/src/plugin.rs` line 112 at r6](https://reviewable.io/reviews/starkware-libs/cairo/5440#-NwdTVFG9BxIq3Gv_m9F:-NwdTVFG9BxIq3Gv_m9G:b-kyljk3) ([raw file](https://github.com/starkware-libs/cairo/blob/2704f82389968232d884b91bfe575a50a0c55bcd/crates/cairo-lang-defs/src/plugin.rs#L112)):*
> ```XML
>     /// Must return a subset of `declared_attributes`.
>     /// This mechanism is optional.
>     fn executable_attributes(&self) -> IntoIter<String> {
> ```

It doesn't make sense to return `vec::IntoIter` here. Return an owned `Vec` (you'll boil down to the same thing). You could also return `impl Iterator<Item=String>`, but this trait is using in `dyn` context so this is impossible.

maciektr:
*[`crates/cairo-lang-defs/src/plugin.rs` line 112 at r6](https://reviewable.io/reviews/starkware-libs/cairo/5440#-NwdTVFG9BxIq3Gv_m9F:-NwdXFNZAdah_r2_gunr:b-896fix) ([raw file](https://github.com/starkware-libs/cairo/blob/2704f82389968232d884b91bfe575a50a0c55bcd/crates/cairo-lang-defs/src/plugin.rs#L112)):*

It doesn't make sense to return `vec::IntoIter` here. Return an owned `Vec` (you'll boil down to the same thing). You could also return `impl Iterator<Item=String>`, but this trait is using in `dyn` context so this is impossible.
</blockquote></details>

Done.

mkaput (changes requested):
> Reviewed 2 of 2 files at r6, 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/5440)* status: all files reviewed, 5 unresolved discussions (waiting on @maciektr and @orizi)

*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/5440#-NwdiUPY40BLEQGqW_Ab:-NwdiUPZ5MCHlzpSJUz7:bavgo84) (no related file):*
I am missing tests for this feature. Primarily, I'm interested in verifying the following aspects:
1. Executables are indeed collected and reflected in the artifact.
2. If they are found, dead code optimization & inlining do their job.

___
___
*[`crates/cairo-lang-sierra-generator/src/executables.rs` line 26 at r7](https://reviewable.io/reviews/starkware-libs/cairo/5440#-NwdjSyI2skJVOSOnOfV:-NwdjSyJ0vsHdiPOVwmX:blot2ab) ([raw file](https://github.com/starkware-libs/cairo/blob/45d312121dde92dbf646f9d5185a0569c11627bc/crates/cairo-lang-sierra-generator/src/executables.rs#L26)):*
> ```XML
>     db: &dyn SierraGenGroup,
>     main_crate_ids: Vec<CrateId>,
> ) -> HashMap<ConcreteFunctionWithBodyId, Vec<SmolStr>> {
> ```

FYI I'm afraid of the computational complexity of this function and the amount of memory allocations that are happening here (mostly via `Vec::new`), but it's good for the first iteration.

orizi (changes requested):
> Reviewed 3 of 5 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/5440)* status: 12 of 14 files reviewed, 2 unresolved discussions (waiting on @maciektr and @mkaput)

___
*[`crates/cairo-lang-compiler/src/lib.rs` line 163 at r8](https://reviewable.io/reviews/starkware-libs/cairo/5440#-NwjTRIA5tdm3GSZlV-g:-NwjTRIA5tdm3GSZlV-h:b-h4ifyk) ([raw file](https://github.com/starkware-libs/cairo/blob/762b52f427c373fffecc60f5a26a74fa477813ca/crates/cairo-lang-compiler/src/lib.rs#L163)):*
> ```XML
>     let executable_functions = find_executable_function_ids(db, main_crate_ids.clone());
> 
>     let mut sierra_program_with_debug = if executable_functions.is_empty() {
> ```

does this actually make sense? isn't failing more logical?

maciektr:
*[`crates/cairo-lang-compiler/src/lib.rs` line 163 at r8](https://reviewable.io/reviews/starkware-libs/cairo/5440#-NwjTRIA5tdm3GSZlV-g:-NwjYOV0Dkos5DgF5I99:ba1ga76) ([raw file](https://github.com/starkware-libs/cairo/blob/762b52f427c373fffecc60f5a26a74fa477813ca/crates/cairo-lang-compiler/src/lib.rs#L163)):*

does this actually make sense? isn't failing more logical?
</blockquote></details>

Hmm, why? You can have a simple Cairo build without any plugins.

orizi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/5440)* status: 12 of 14 files reviewed, 2 unresolved discussions (waiting on @maciektr and @mkaput)

___
*[`crates/cairo-lang-compiler/src/lib.rs` line 163 at r8](https://reviewable.io/reviews/starkware-libs/cairo/5440#-NwjTRIA5tdm3GSZlV-g:-NwjZxcA6EEnYN-3fCJ-:bgw9ut6) ([raw file](https://github.com/starkware-libs/cairo/blob/762b52f427c373fffecc60f5a26a74fa477813ca/crates/cairo-lang-compiler/src/lib.rs#L163)):*
<details><summary><i>Previously, maciektr (maciektr) wrote…</i></summary><blockquote>

Hmm, why? You can have a simple Cairo build without any plugins.
</blockquote></details>

You can - but it is meaning less building without a particular target.

maciektr:
*[`crates/cairo-lang-compiler/src/lib.rs` line 163 at r8](https://reviewable.io/reviews/starkware-libs/cairo/5440#-NwjTRIA5tdm3GSZlV-g:-Nwj_RQsCvu37qljIP11:bd229hl) ([raw file](https://github.com/starkware-libs/cairo/blob/762b52f427c373fffecc60f5a26a74fa477813ca/crates/cairo-lang-compiler/src/lib.rs#L163)):*

You can - but it is meaning less building without a particular target.
</blockquote></details>

I am not sure if I would call it meaning less. It's what we have now with the scarb lib target - it compiles Cairo to Sierra. It can be used for instance to run functions by name, it's what some tooling uses already, or even simply for debugging purposes. I don't think we are ready to abandon it completely now.

orizi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/5440)* status: 12 of 14 files reviewed, 2 unresolved discussions (waiting on @maciektr and @mkaput)

___
*[`crates/cairo-lang-compiler/src/lib.rs` line 163 at r8](https://reviewable.io/reviews/starkware-libs/cairo/5440#-NwjTRIA5tdm3GSZlV-g:-NwjgQyT8wv1cRiHP6Dd:bbfq9lm) ([raw file](https://github.com/starkware-libs/cairo/blob/762b52f427c373fffecc60f5a26a74fa477813ca/crates/cairo-lang-compiler/src/lib.rs#L163)):*
<details><summary><i>Previously, maciektr (maciektr) wrote…</i></summary><blockquote>

I am not sure if I would call it meaning less. It's what we have now with the scarb lib target - it compiles Cairo to Sierra. It can be used for instance to run functions by name, it's what some tooling uses already, or even simply for debugging purposes. I don't think we are ready to abandon it completely now.
</blockquote></details>

so at the very least it makes no sense without --replace-ids.

and should probably have a todo to deprecate.

maciektr:
*[`crates/cairo-lang-compiler/src/lib.rs` line 163 at r8](https://reviewable.io/reviews/starkware-libs/cairo/5440#-NwjTRIA5tdm3GSZlV-g:-Nwk3Zl94r51_p5qLu7_:b-tsh0f1) ([raw file](https://github.com/starkware-libs/cairo/blob/762b52f427c373fffecc60f5a26a74fa477813ca/crates/cairo-lang-compiler/src/lib.rs#L163)):*

so at the very least it makes no sense without --replace-ids.

and should probably have a todo to deprecate.
</blockquote></details>

Yes, logically you are probably right. I don't think it's worth to ban this specific configuration though, especially since we would really have to verify with community if there is any use case we are yet not aware of.

orizi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/5440)* status: 12 of 14 files reviewed, 2 unresolved discussions (waiting on @maciektr and @mkaput)

___
*[`crates/cairo-lang-compiler/src/lib.rs` line 163 at r8](https://reviewable.io/reviews/starkware-libs/cairo/5440#-NwjTRIA5tdm3GSZlV-g:-NwmRMms9mPKszq_Rjnz:byi88rx) ([raw file](https://github.com/starkware-libs/cairo/blob/762b52f427c373fffecc60f5a26a74fa477813ca/crates/cairo-lang-compiler/src/lib.rs#L163)):*
<details><summary><i>Previously, maciektr (maciektr) wrote…</i></summary><blockquote>

Yes, logically you are probably right. I don't think it's worth to ban this specific configuration though, especially since we would really have to verify with community if there is any use case we are yet not aware of.
</blockquote></details>

i don't see any way for this to be usable.
i'd rather add a failure anyway (as replace-ids is would keep the same meaning - and is always possible), so it is low risk as long as you give a proper error.

mkaput (approved):
:lgtm:

> Reviewed 5 of 5 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/5440#-:-NxBaQAR4f_9VQNp35PF:bnfp4nl)* status: all files reviewed, 2 unresolved discussions (waiting on @maciektr)

orizi (commented):
> Reviewed 2 of 2 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/5440)* status: all files reviewed, 2 unresolved discussions (waiting on @maciektr)

mkaput (commented):
> Reviewed 2 of 2 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/5440)* status: all files reviewed, 2 unresolved discussions (waiting on @maciektr)

maciektr (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/5440)* status: all files reviewed, 2 unresolved discussions (waiting on @mkaput)

*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/5440#-NvkQdAQ3r9iu2waS9QY:-NxDInR9Egn8KWb9c2j0:bel2era) (no related file):*

i think every build should have some sort of basic plugin - so the build would have some information about possibly required data.
no need for a catch all allowing attr `executable` just for test - just grabbing what is marked with `test`.
for proof - validate extra things on the called function (for example, return type must be `Drop`able) and finding the set of these functions (for a weird project with more than one)
</blockquote></details>

@orizi What would you say, if for now we make it a warning on the Scarb side? 

___
*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/5440#-NwdiUPY40BLEQGqW_Ab:-NxDIlIKC7qzCuOLgt98:b-896fix) (no related file):*

I am missing tests for this feature. Primarily, I'm interested in verifying the following aspects:
1. Executables are indeed collected and reflected in the artifact.
2. If they are found, dead code optimization & inlining do their job.
</blockquote></details>

Done.

___

orizi (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/5440)* status: all files reviewed, 2 unresolved discussions (waiting on @maciektr and @mkaput)

*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/5440#-NvkQdAQ3r9iu2waS9QY:-NxDWTO-EpvQ628kdW6f:b-6o5nlz) (no related file):*
<details><summary><i>Previously, maciektr (maciektr) wrote…</i></summary><blockquote>

@orizi What would you say, if for now we make it a warning on the Scarb side?
</blockquote></details>

sounds great - just add a TODO here.

___

maciektr:

sounds great - just add a TODO here.
</blockquote></details>

Great, added!

maciektr (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/5440)* status: 13 of 14 files reviewed, 2 unresolved discussions (waiting on @mkaput and @orizi)

___
*[`crates/cairo-lang-compiler/src/lib.rs` line 163 at r8](https://reviewable.io/reviews/starkware-libs/cairo/5440#-NwjTRIA5tdm3GSZlV-g:-NxDaK8SBy8wP-qQmBdp:b-896fix) ([raw file](https://github.com/starkware-libs/cairo/blob/762b52f427c373fffecc60f5a26a74fa477813ca/crates/cairo-lang-compiler/src/lib.rs#L163)):*

i don't see any way for this to be usable.
i'd rather add a failure anyway (as replace-ids is would keep the same meaning - and is always possible), so it is low risk as long as you give a proper error.
</blockquote></details>

Done.

orizi (approved):
:lgtm:

> Reviewed 1 of 5 files at r8, 1 of 1 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/5440#-:-NxDcV0e-FOLicS2YrjT:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @mkaput)

PR #5515, opened 2024-05-07
Add `Debug` implementation with clippy

tcoratger:
In circumstances like [cairo_native](https://github.com/lambdaclass/cairo_native), and other projects working on Cairo and its compilation, it is extremely useful to have a `Debug` implementation for each public structure and enum. 

Therefore, the lint [`MISSING_DEBUG_IMPLEMENTATIONS`](https://doc.rust-lang.org/stable/nightly-rustc/rustc_lint/builtin/static.MISSING_DEBUG_IMPLEMENTATIONS.html) is added with a derivation of `Debug` wherever possible, along with an `allow` attribute when the implementation is more complex. 

This allows for the implementation of `Debug` for a large number of new objects that did not previously have this implementation, with the potential for further refinement in the future.


orizi (changes requested):
> Reviewed 26 of 129 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/5515)* status: 26 of 129 files reviewed, 1 unresolved discussion (waiting on @tcoratger)

*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/5515#-NxWcx4NEy3UiE4S0dsL:-NxWcx4NEy3UiE4S0dsM:b-be15yn) (no related file):*
i'm not sure at all we actually want this - a lot of the crates are mostly internal crates that are not expected to be used by tools like `cairo_native` and such - unless we have a lot of such cases where we have something missing - i'd rather avoid the additional costs of having it everywhere (in compile time).

___

tcoratger (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/5515)* status: 26 of 129 files reviewed, 1 unresolved discussion (waiting on @orizi)

*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/5515#-NxWcx4NEy3UiE4S0dsL:-NxWeCUXCOBXtp_iAEEn:b-eaau7s) (no related file):*

i'm not sure at all we actually want this - a lot of the crates are mostly internal crates that are not expected to be used by tools like `cairo_native` and such - unless we have a lot of such cases where we have something missing - i'd rather avoid the additional costs of having it everywhere (in compile time).
</blockquote></details>

Would you accept other PRs on a case by case basis? I would first think about `ProgramRegistry` where a debug derivation would be more than useful for debugging in various tools.

___

orizi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/5515)* status: 26 of 129 files reviewed, 1 unresolved discussion (waiting on @tcoratger)

*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/5515#-NxWcx4NEy3UiE4S0dsL:-NxWfVpY98ulg57jQmxn:bsttegq) (no related file):*
<details><summary><i>Previously, tcoratger (Thomas Coratger) wrote…</i></summary><blockquote>

But on the other hand, I think that the `Debug` derivation can be useful also inside the compiler itself for various debugging scenarios. And I don't have the impression that it increases the compilation time much. What do you think?
</blockquote></details>

a case by case basic would totally be fine - but particularly - the `ProgramRegistry` sounds to me like an example of something that would be barely usable - as the Sierra `Program` itself would be more useful to begin with.

We need to analyse it as a case by case to see if we need the debug - for a lot of existing structure adding such a debug might be more confusing (just because there's the entire `DebugWithDb` structure all around the project, for salsa based stuff)

___

PR #5539, opened 2024-05-09
Made feedback set algorithm not ignore the nodes post cycle.

### LINKED ISSUE #5524: bug: gas computation / unexpected cycle in computation issue with recursive type definition
# Bug Report

**Cairo version:**

At least on `2.5.4` and `2.6.3`

**Current behavior:**

For the Dojo project, we use the following types to define the layout of our data that will be stored in a dedicated smart contract:

```rust
#[derive(Copy, Drop, Serde, Debug, PartialEq)]
struct FieldLayout {
    selector: felt252,
    layout: Layout
}

#[derive(Copy, Drop, Serde, Debug, PartialEq)]
enum Layout {
    Fixed: Span<u8>,
    Struct: Span<FieldLayout>,
    Tuple: Span<FieldLayout>,
    Array: Span<FieldLayout>,
    ByteArray,
    Enum: Span<FieldLayout>,
}
```

Then, we have 3 functions to read/write/delete an object based on its layout.
This code leads to: 
- `Failed calculating gas usage, it is likely a call for `gas::withdraw_gas` is missing.` error when we try to run our test suite.
- `Compiling Sierra class to CASM with compiler version 2.5.4...
Error: found an unexpected cycle during cost computation` when we try to declare the compiled contract class on Katana.

I built the following tiny project outside of Dojo to reproduce the issue : https://github.com/remybar/scarb-recursion

In this project, there is only these 2 types and the `entity()` function which reads an object thanks to the `_read_layout()` function.

This project has exactly the same issues than our Dojo project.
If I remove the `Struct: Span<FieldLayout>` variant from `Layout`, I can run `scarb test` and declare the contract class on Katana. 
Also, if the `_read_layout()` function does not have any loop inside, everything works well.

So, I guess the issue is due to the fact that 2 variants from `Layout` refers, directly on indirectly through an intermediate type, to itself (the `Layout` type). And if we loop on this type, the gas computation failed.

**Expected behavior:**

At a first glance, this Cairo code seems legit to me and should work fine.

**Steps to reproduce:**

Just run `scarb test` and/or declare the contract class on `Katana`.

**Related code:**

https://github.com/remybar/scarb-recursion

orizi:
Fixes #5524

- - -
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/5539)

### DIFF
--- a/crates/cairo-lang-lowering/src/graph_algorithms/feedback_set.rs
+++ b/crates/cairo-lang-lowering/src/graph_algorithms/feedback_set.rs
@@ -41,7 +41,7 @@ pub fn priv_function_with_body_feedback_set_of_representative(
     function: ConcreteSCCRepresentative,
 ) -> Maybe<OrderedHashSet<ConcreteFunctionWithBodyId>> {
     Ok(calc_feedback_set(
-        &ConcreteFunctionWithBodyNode {
+        ConcreteFunctionWithBodyNode {
             function_id: function.0,
             db,
             dependency_type: DependencyType::Cost,
--- a/crates/cairo-lang-utils/src/graph_algos/feedback_set.rs
+++ b/crates/cairo-lang-utils/src/graph_algos/feedback_set.rs
@@ -7,6 +7,8 @@
 //! so here we implement some straight-forward algorithm that guarantees to cover all the cycles in
 //! the graph, but doesn't necessarily produce the minimum size of such a set.
 
+use std::collections::VecDeque;
+
 use super::graph_node::GraphNode;
 use super::scc_graph_node::SccGraphNode;
 use super::strongly_connected_components::ComputeScc;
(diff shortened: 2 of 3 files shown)


gilbens-starkware (commented):
> Reviewed 2 of 3 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/5539)* status: all files reviewed (commit messages unreviewed), 1 unresolved discussion (waiting on @orizi and @yuvalsw)

___
*[`crates/cairo-lang-utils/src/graph_algos/feedback_set.rs` line 33 at r2](https://reviewable.io/reviews/starkware-libs/cairo/5539#-NxdeLrZ1Ge2Rc1hrBR_:-NxdeLr_DV7yA2BSvuMf:b45h) ([raw file](https://github.com/starkware-libs/cairo/blob/588a8b0c9582c347cd8a1f2e5d3b46fd4d24191c/crates/cairo-lang-utils/src/graph_algos/feedback_set.rs#L33)):*
> ```XML
>     /// reached, it indicates it's in some cycle that was not "resolved" yet.
>     in_flight: UnorderedHashSet<Node::NodeId>,
>     /// Node already visited in the current run.
> ```





_Suggestion:_
```XML
/// Already visited nodes in the current run.
```

gilbens-starkware (approved):
:lgtm:

> Reviewed 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/5539#-:-NxfXO5F6eq7Tz6SDpqI:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @yuvalsw)

gilbens-starkware (approved):
> Reviewed 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/5539)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @yuvalsw)

yuvalsw (commented):
:lgtm:

> Reviewed 2 of 3 files at r1, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/5539#-:-NxfsxeJ-RvDAL9aWYuJ:bnfp4nl)* status: all files reviewed, 2 unresolved discussions

___
*[`crates/cairo-lang-utils/src/graph_algos/feedback_set.rs` line 25 at r4](https://reviewable.io/reviews/starkware-libs/cairo/5539#-NxfovRVB8gO3sj5WFSG:-NxfovRVB8gO3sj5WFSH:b-krx9pw) ([raw file](https://github.com/starkware-libs/cairo/blob/2f97689d8e3155f7de62b31c173b0bba34672423/crates/cairo-lang-utils/src/graph_algos/feedback_set.rs#L25)):*
> ```XML
> struct FeedbackSetAlgoContext<Node: ComputeScc> {
>     /// Nodes that were discovered as reachable from the current run, but possibly were not yet
>     /// visited.
> ```

please elaborate about the "possibly" here - what are the cases where they were visited and still pending? (IIUC, if visited after being added to pending)



_Code quote:_
```XML
    /// Nodes that were discovered as reachable from the current run, but possibly were not yet
    /// visited.
```
___
*[`crates/cairo-lang-utils/src/graph_algos/feedback_set_test.rs` line 144 at r4](https://reviewable.io/reviews/starkware-libs/cairo/5539#-NxfsJ9I-HkK3dvjao8K:-NxfsJ9I-HkK3dvjao8L:biawbwe) ([raw file](https://github.com/starkware-libs/cairo/blob/2f97689d8e3155f7de62b31c173b0bba34672423/crates/cairo-lang-utils/src/graph_algos/feedback_set_test.rs#L144)):*
> ```XML
>         vec![1, 2],
>         // 1:
>         vec![1, 0],
> ```

Did this fail before the fix?
I think the bug was only relevant in the case of [0, 1]. Here you first find the size-1 cycle [1], and then don't add 0 to the fset and don't break. If it was [0, 1] (before the fix) you would not add 1, add 0 and break, and skip 2, missing the size-1 cycle [2].



_Code quote:_
```XML
vec![1, 0],
```

orizi (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/5539)* status: all files reviewed, 2 unresolved discussions

___
*[`crates/cairo-lang-utils/src/graph_algos/feedback_set_test.rs` line 144 at r4](https://reviewable.io/reviews/starkware-libs/cairo/5539#-NxfsJ9I-HkK3dvjao8K:-Nxg3Iqr4LMd9wLZk8CD:b-5ynihf) ([raw file](https://github.com/starkware-libs/cairo/blob/2f97689d8e3155f7de62b31c173b0bba34672423/crates/cairo-lang-utils/src/graph_algos/feedback_set_test.rs#L144)):*

Did this fail before the fix?
I think the bug was only relevant in the case of [0, 1]. Here you first find the size-1 cycle [1], and then don't add 0 to the fset and don't break. If it was [0, 1] (before the fix) you would not add 1, add 0 and break, and skip 2, missing the size-1 cycle [2].
</blockquote></details>

it failed only on the case starting from 0.
a similar case failed for more. (where the edges were sorted)

PR #5632, opened 2024-05-24
LS: Make all paths at input absolute

mkaput:
Language server's Salsa **must** use absolute paths for `FileLongId`s
**exclusively**, as panics would happen when trying to generate
`file://` URLs for LSP use.

This is something LS-specific. It doesn't matter for the compiler itself
because none of its interfaces are sensitive to this problem, hence the
fix is applied at call sites instead of enforcing this within
`FileLongId`. And probably such enforcement would not be desired, as
diagnostics output etc. would have to be able to relativize such
absolute paths for clean look.

In the ideal world, a proper VFS abstraction in
`cairo-lang-filesystem` should provide a mean to enforce absolute
paths if needed, but this is very much TBD yet.

---

**Stack**:
- #5671
- #5670
- #5632 ⬅

- - -
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/5632)

### DIFF
--- a/crates/cairo-lang-language-server/src/lib.rs
+++ b/crates/cairo-lang-language-server/src/lib.rs
@@ -581,6 +581,10 @@ impl Backend {
             }
 
             Some(ProjectManifestPath::CairoProject(config_path)) => {
+                // The base path of ProjectConfig must be absolute to ensure that all paths in Salsa
+                // DB will also be absolute.
+                assert!(config_path.is_absolute());
+
                 try_to_init_unmanaged_core(&*self.config.read().await, db);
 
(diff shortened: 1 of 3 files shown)


mkaput (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/5632)* status: 0 of 2 files reviewed, all discussions resolved

*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/5632#-NyfxTef9jFiHniQwqqV:-NyzRGsB-6ItcLplPaWa:bb6un1j) (no related file):*

Working: I want to write tests that are fixed by these changes.
</blockquote></details>

Done

___

PR #5993, opened 2024-07-10
Remove `__validate_declare__` func

FilipLaurentiu:
The function is not required according to https://docs.starknet.io/architecture-and-concepts/accounts/account-functions/ and this is missleading.

It makes more sense to have the interface containing only the required functions that need to be implemented.

### DIFF
--- a/corelib/src/starknet/account.cairo
+++ b/corelib/src/starknet/account.cairo
@@ -9,7 +9,6 @@ pub struct Call {
 
 #[starknet::interface]
 pub trait AccountContract<TContractState> {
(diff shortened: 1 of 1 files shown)


orizi (commented):
Don't think we can actually remove it - as it will break all implementers.
@ArielElp  - any suggestions?
maybe just a doc that this is not depracated.

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/5993#-:-O1QYcV03w03pU2gx1gq:blj5msn)* status: 0 of 1 files reviewed, all discussions resolved

PR #6023, opened 2024-07-15
feat: allow to access ByteArray fields

Th0rgal:
This makes `ByteArray` fields accessible outside of the crate.

A ByteArray is stored in memory at two locations: one storage slot receives its length, while other slots store the multiple elements (data + pending_word).

I use ByteArray to store Strings. In my use case, most of the time the stored String would fit in a single felt and my users end up paying twice the price. I would like to impl an alternative _storage optimized_ trait which only stores the data and pending_word and retrieves the String size by iterating until it finds an empty string. To do so efficiently I would need to be able to read the content of the struct. I could also add the alternative Store trait to a pr if you think that could be useful.

### DIFF
--- a/corelib/src/byte_array.cairo
+++ b/corelib/src/byte_array.cairo
@@ -24,14 +24,14 @@ const BYTES_IN_BYTES31_MINUS_ONE: usize = BYTES_IN_BYTES31 - 1;
 pub struct ByteArray {
     // Full "words" of 31 bytes each. The first byte of each word in the byte array
     // is the most significant byte in the word.
-    pub(crate) data: Array<bytes31>,
+    pub data: Array<bytes31>,
     // This felt252 actually represents a bytes31, with < 31 bytes.
     // It is represented as a felt252 to improve performance of building the byte array.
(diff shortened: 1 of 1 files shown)


orizi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/6023)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @Th0rgal)

*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/6023#-O1vPXA73x7T1pPiey8F:-O1vPXA73x7T1pPiey8G:bxbc32n) (no related file):*
This cannot be approved - as it breaks encapsulation (which would break any attempts for general future optimizations).

You may consider allowing for a conversion into `felt252` - so you'd easily find and convert into short strings.
Also - just having an enum `Short: felt252` | `Long: ByteArray` would probably meet most of your basic demands.

in general - how do you expect to save your strings in the storage?
do note that having objects with varying size is not actually ok (if you where planning on saving the str continusly).

___

PR #6232, opened 2024-08-19
feat: publicly expose some EthAddress trait impl

tdelabro:
I personally need `EthAddressIntoFelt252`, and I don't see any reason for the other ones not to be exposed to the lib users. How am I supposed to construct an `EthAdress` value when it has a private field, no new method, and all the `Into`/`TryInto` methods are private?
If I cannot construct an `EthAdress`, I cannot use the methods in `corelib/src/starknet/eth_signature.cairo`
Correct me if I'm wrong

- - -
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/6232)

### DIFF
--- a/corelib/src/starknet/eth_address.cairo
+++ b/corelib/src/starknet/eth_address.cairo
@@ -20,7 +20,7 @@ impl EthAddressStorePacking of starknet::StorePacking<EthAddress, felt252> {
     }
 }
 
-pub(crate) impl Felt252TryIntoEthAddress of TryInto<felt252, EthAddress> {
+pub impl Felt252TryIntoEthAddress of TryInto<felt252, EthAddress> {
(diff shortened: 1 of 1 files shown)


orizi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/6232)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @tdelabro)

*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/6232#-O4jLd-u4mySNKhQVAKv:-O4jLd-u4mySNKhQVAKw:b-7p574a) (no related file):*
use the `Into` / `TryInto` trait directly.
no need for using the impls.
close the PR.

___

PR #6496, opened 2024-10-17
LS: search scarb in env var `PATH`

j1mbo64:
Find Scarb executable without having to set `SCARB` environment variable if `Scarb` is installed.
This change makes it easier to integrate LSP with other code editors.

### DIFF
--- a/crates/cairo-lang-language-server/src/toolchain/scarb.rs
+++ b/crates/cairo-lang-language-server/src/toolchain/scarb.rs
@@ -1,5 +1,6 @@
 use std::path::{Path, PathBuf};
 use std::sync::{Arc, OnceLock};
+use std::{env, fs};
 
 use anyhow::{Context, Result, bail};
 use lsp_types::notification::Notification;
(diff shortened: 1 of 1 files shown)


lotem-starkware:
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/6496)

mkaput:
This approach will not work nicely, unfortunately, because CairoLS demands a very, very specific version of Scarb: both have to source the same revision of the Cairo compiler.

We are solving this problem by bundling matching CairoLS with Scarb. It is available as `scarb cairo-language-server`. You can look for it via the `scarb commands` command, and this is what [VSCode extension does](https://github.com/starkware-libs/cairo/blob/40a7b60687682238f7f71ef7c59c986cc5733915/vscode-cairo/src/scarb.ts#L87), but you could just assume its existence as well if your editor's environment is more limited. The UX in case of users having a stripped down Scarb installation will be just worse, but the happy path will stay the same.

The `SCARB` env var is a way of notifying that bundled CairoLS about the matching Scarb installation. In Scarb-less setups (we call them _standalone_) it should always be unset.

I'm closing this PR because there's no way forward here. Please let me know if you have any questions 😃

PR #6570, opened 2024-11-01
Allow to debug print ascii escape sequence

m-kus:
This change enables ASCII escape sequences in `println!`, which allows to change output text color and style.  

Why:
This is fun and can be used to promote Cairo as a generic programming language.
Also enables some legit use cases e.g. https://github.com/m-kus/cairo-pretty-assertions

Security considerations:
If the source Cairo file is provided by an untrusted third party it can potentially mess with the terminal window title and style; also if the output is logged, there are ways to confuse the log reader (see https://nvd.nist.gov/vuln/detail/CVE-2009-4487)

### DIFF
--- a/crates/cairo-lang-runner/src/short_string.rs
+++ b/crates/cairo-lang-runner/src/short_string.rs
@@ -42,7 +42,7 @@ pub fn as_cairo_short_string_ex(value: &Felt252, length: usize) -> Option<String
     for byte in bytes {
         if byte == 0 {
             as_string.push_str(r"\0");
-        } else if byte.is_ascii_graphic() || byte.is_ascii_whitespace() {
+        } else if byte.is_ascii_graphic() || byte.is_ascii_whitespace() || is_ascii_escape_sequence(byte) {
             as_string.push(byte as char);
         } else {
             as_string.push_str(format!(r"\x{:02x}", byte).as_str());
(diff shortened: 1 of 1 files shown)


orizi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/6570)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @m-kus)

___
*[`crates/cairo-lang-runner/src/short_string.rs` line 45 at r1](https://reviewable.io/reviews/starkware-libs/cairo/6570#-OAsYuarCu9zCqszvMDK:-OAsYuarCu9zCqszvMDL:b-21kto7) ([raw file](https://github.com/starkware-libs/cairo/blob/64487f4a81e80b086daffe91309da5fef4346a7a/crates/cairo-lang-runner/src/short_string.rs#L45)):*
> ```XML
>         if byte == 0 {
>             as_string.push_str(r"\0");
>         } else if byte.is_ascii_graphic() || byte.is_ascii_whitespace() || is_ascii_escape_sequence(byte) {
> ```

run `./scripts/cairo_fmt.sh`.
___
*[`crates/cairo-lang-runner/src/short_string.rs` line 61 at r1](https://reviewable.io/reviews/starkware-libs/cairo/6570#-OAsYUhZ93VZsFLpYmf5:-OAsYUhZ93VZsFLpYmf6:b-dblhjh) ([raw file](https://github.com/starkware-libs/cairo/blob/64487f4a81e80b086daffe91309da5fef4346a7a/crates/cairo-lang-runner/src/short_string.rs#L61)):*
> ```XML
> /// Checks if given byte is a beginning of an ASCII escape sequence
> /// https://gist.github.com/ConnerWill/d4b6c776b509add763e17f9f113fd25b
> fn is_ascii_escape_sequence(byte: u8) -> bool {
> ```

i don't really understand what this does - add some test that actually uses this sort of escaping.

PR #6917, opened 2024-12-24
fix: Improve clarity of comment in the script Update sierra_update_ch…

brawncode:
I've made a small improvement in the comment that explains when the code change does not break previously compilable code. The original phrasing was a bit unclear, so I updated it to make the message more precise.

The updated comment is now:

```bash
echo "If there is a change, but the code that was previously compilable remains compilable:"
```

This should help to avoid any confusion when reading the script and make the intention behind the condition more understandable.


orizi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/6917)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @brawncode)

___
*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/6917#-OEsjCDu5q7xyy_V9xfM:-OEsjCDu5q7xyy_V9xfN:bf07ina) (no related file):*
the new comment is plain wrong.

please close PR.

PR #6924, opened 2024-12-25
fix(corelib): migrate deprecated-index-traits

julio4:
- Migrate all deprecated `core::traits::{IndexView, Index}` to `core::ops::{IndexView, Index}` (`Array`, `ByteArray`, `bytes31`, `Felt252Dict`)
- Add missing documentation
- Rename `IndexView` implementation of `ArrayIndex` and `SpanIndex` to `ArrayIndexView` and `SpanIndexView`

### DIFF
--- a/corelib/src/array.cairo
+++ b/corelib/src/array.cairo
@@ -65,12 +65,12 @@ use crate::box::BoxTrait;
 use crate::iter::Iterator;
 use crate::metaprogramming::TypeEqual;
 use crate::serde::Serde;
+use crate::ops::IndexView;
 use crate::RangeCheck;
 #[allow(unused_imports)]
 use crate::gas::withdraw_gas;
 #[allow(unused_imports)]
 use crate::option::OptionTrait;
-use crate::ops::IndexView;
 /// A collection of elements of the same type continuous in memory.
 #[derive(Drop)]
 pub extern type Array<T>;
--- a/corelib/src/array.cairo
+++ b/corelib/src/array.cairo
@@ -70,8 +70,7 @@ use crate::RangeCheck;
 use crate::gas::withdraw_gas;
 #[allow(unused_imports)]
 use crate::option::OptionTrait;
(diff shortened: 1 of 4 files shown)


orizi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/6924)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @julio4)

___
*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/6924#-OEz04mf7-2SkKIyK-SS:-OEz04mf7-2SkKIyK-ST:bae6gps) (no related file):*
can't do this migration - as existing code relying on the old traits still needs to work.
this will only be done in 3.* sort of breaking change.

julio4:

can't do this migration - as existing code relying on the old traits still needs to work.
this will only be done in 3.* sort of breaking change.
</blockquote></details>

I'm not entirely sure how existing code break as it seems the behaviors is the same, but I'm sure I don't fully understand the side effects. You can close this PR then, (or change base to cairo-major-update?)

orizi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/6924)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @julio4)

___
*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/6924#-OEz04mf7-2SkKIyK-SS:-OEz3nPuDC6uLkK_8QFt:b-rbmhe6) (no related file):*

I'm not entirely sure how existing code break as it seems the behaviors is the same, but I'm sure I don't fully understand the side effects. You can close this PR then, (or change base to cairo-major-update?)
</blockquote></details>

If someone already had code depending on implementation of `traits::Index` - now we would be missing an implementation for it.

the current way did 2 things -
1. encouraged people to use only the new trait.
2. Provided the new trait assuming the old trait exists.

PR #6941, opened 2024-12-27
feat(corelib): storage vectors iterators

julio4:
Add new iterator trait `IntoIterRange<T>` in the vec module to turn a collection of values into an iterator over a specific range, with `into_iter_range(range: Range<u64>)` or `into_iter_full_range()`.

Add `IntoIterRange` implementations for storage vectors to enable iterator over vectors in storage.

1. `VecIntoIterRange` and `MutableVecIntoIterRange` for direct iteration over `StoragePath<Vec<T>>` and `StoragePath<Mutable<Vec<T>>>`
2. `PathableVecIntoIterRange` and `PathableMutableVecIntoIterRange` for iteration over types implementing `StorageAsPath` into storage vectors (such as Vecs in Starknet contract storage structs)

You need to bring the trait in the current scope:
```cairo
use starknet::storage::IntoIterRange;
```

Before:
```cairo
for i in 0..vec.len() {
    let entry = vec.at(i);
}
```

After:
```cairo
for entry in vec.into_iter_full_range() {
    // entry: StoragePath<T> or StoragePath<Mutable<T>>
}
```

#### Examples

```cairo
use starknet::storage::{
    IntoIterRange, MutableVecTrait, StoragePathEntry, StoragePointerReadAccess,
    StoragePointerWriteAccess,
};

#[starknet::contract]
mod contract_with_vec {
    use starknet::storage::Vec;
    #[storage]
    pub struct Storage {
        pub simple: Vec<u32>,
    }
}

fn vec_iter() {
    let mut mut_state = contract_with_vec::contract_state_for_testing();
    for i in 0..9_usize {
        mut_state.simple.append().write(i);
    };

    let state = @contract_with_vec::contract_state_for_testing();
    let mut i = 0;
    for entry in state.simple.into_iter_full_range() {
        assert_eq!(entry.read(), i);
        i += 1;
    };
    assert_eq!(i, 9);

    let mut i = 2;
    for entry in state.simple.into_iter_range(2..5) {
        assert_eq!(entry.read(), i);
        i += 1;
    };
    assert_eq!(i, 5);
}
```

### DIFF
--- a/corelib/src/iter.cairo
+++ b/corelib/src/iter.cairo
@@ -230,4 +230,4 @@
 //! [`map`]: Iterator::map
 mod adapters;
 mod traits;
-pub use traits::{IntoIterRange, IntoIterator, Iterator};
+pub use traits::{IntoIterator, Iterator};
--- a/corelib/src/iter.cairo
+++ b/corelib/src/iter.cairo
@@ -230,4 +230,4 @@
 //! [`map`]: Iterator::map
 mod adapters;
 mod traits;
-pub use traits::{FromIterator, IntoIterator, Iterator};
+pub use traits::{IntoIterRange, IntoIterator, Iterator};
--- a/corelib/src/iter.cairo
+++ b/corelib/src/iter.cairo
@@ -230,4 +230,4 @@
 //! [`map`]: Iterator::map
 mod adapters;
 mod traits;
-pub use traits::{IntoIterator, Iterator};
+pub use traits::{FromIterator, IntoIterator, Iterator};
--- a/corelib/src/iter/traits/iterator.cairo
+++ b/corelib/src/iter/traits/iterator.cairo
@@ -1,5 +1,4 @@
 use crate::iter::adapters::{Map, mapped_iterator};
-use crate::ops::Range;
 
 /// A trait for dealing with iterators.
 ///
@@ -103,16 +102,6 @@ pub trait IntoIterator<T> {
     impl Iterator: Iterator<Self::IntoIter>;
     /// Creates an iterator from a collection.
     fn into_iter(self: T) -> Self::IntoIter;
-}
-
-/// Turn a collection of values into an iterator over a specific range.
-pub trait IntoIterRange<T, I> {
-    type IntoIter;
-    impl Iterator: Iterator<Self::IntoIter>;
-    /// Creates an iterator over a range from a collection.
-    fn into_iter_range(self: T, range: Range<I>) -> Self::IntoIter;
-    /// Creates an iterator over the full range of a collection.
-    fn into_iter_full_range(self: T) -> Self::IntoIter;
 }
 
 impl IteratorIntoIterator<T, +Iterator<T>> of IntoIterator<T> {
--- a/corelib/src/iter/traits/iterator.cairo
+++ b/corelib/src/iter/traits/iterator.cairo
@@ -1,4 +1,5 @@
 use crate::iter::adapters::{Map, mapped_iterator};
+use crate::ops::Range;
 
 /// A trait for dealing with iterators.
 ///
@@ -92,5 +93,31 @@ pub trait Iterator<T> {
         self: T, f: F,
     ) -> Map<T, F> {
         mapped_iterator(self, f)
+    }
+}
+
+/// Turn a collection of values into an iterator.
+pub trait IntoIterator<T> {
+    /// The iterator type that will be created.
+    type IntoIter;
+    impl Iterator: Iterator<Self::IntoIter>;
+    /// Creates an iterator from a collection.
+    fn into_iter(self: T) -> Self::IntoIter;
+}
+
+/// Turn a collection of values into an iterator over a specific range.
+pub trait IntoIterRange<T, I> {
+    type IntoIter;
+    impl Iterator: Iterator<Self::IntoIter>;
+    /// Creates an iterator over a range from a collection.
+    fn into_iter_range(self: T, range: Range<I>) -> Self::IntoIter;
+    /// Creates an iterator over the full range of a collection.
+    fn into_iter_full_range(self: T) -> Self::IntoIter;
+}
+
+impl IteratorIntoIterator<T, +Iterator<T>> of IntoIterator<T> {
+    type IntoIter = T;
+    fn into_iter(self: T) -> T {
+        self
     }
 }
--- a/corelib/src/iter/traits/iterator.cairo
+++ b/corelib/src/iter/traits/iterator.cairo
@@ -94,19 +94,3 @@ pub trait Iterator<T> {
         mapped_iterator(self, f)
     }
 }
-
-/// Turn a collection of values into an iterator.
-pub trait IntoIterator<T> {
-    /// The iterator type that will be created.
-    type IntoIter;
-    impl Iterator: Iterator<Self::IntoIter>;
-    /// Creates an iterator from a collection.
-    fn into_iter(self: T) -> Self::IntoIter;
-}
-
-impl IteratorIntoIterator<T, +Iterator<T>> of IntoIterator<T> {
-    type IntoIter = T;
-    fn into_iter(self: T) -> T {
-        self
-    }
-}
--- a/corelib/src/starknet/storage.cairo
+++ b/corelib/src/starknet/storage.cairo
@@ -139,7 +139,6 @@
 //! address (the `sn_keccak` hash of the variable name) combined with the mapping keys or vector
 //! indices.
 //! See their respective module documentation for more details.
-
 use core::hash::HashStateTrait;
 #[allow(unused_imports)]
 use core::pedersen::HashState;
@@ -161,8 +160,11 @@ mod sub_pointers;
 pub use sub_pointers::{SubPointers, SubPointersForward, SubPointersMut, SubPointersMutForward};
 
 mod vec;
-pub use vec::{IntoIterRange, MutableVecTrait, Vec, VecTrait};
-use vec::{MutableVecIndexView, VecIndexView};
+use vec::{
+    MutableVecIndexView, MutableVecIntoIterRange, PathableMutableVecIntoIterRange,
+    PathableVecIntoIterRange, VecIndexView, VecIntoIterRange,
+};
+pub use vec::{MutableVecTrait, Vec, VecTrait};
 
 /// A pointer to an address in storage, can be used to read and write values, if the generic type
 /// supports it (e.g. basic types like `felt252`).
@@ -538,4 +540,14 @@ trait MutableTrait<T> {
 
 impl MutableImpl<T> of MutableTrait<Mutable<T>> {
     type InnerType = T;
+}
+
+/// Trait for turning collection of values into an iterator over a specific range.
+pub trait IntoIterRange<T> {
+    type IntoIter;
+    impl Iterator: Iterator<Self::IntoIter>;
+    /// Creates an iterator over a range from a collection.
+    fn into_iter_range(self: T, range: core::ops::Range<u64>) -> Self::IntoIter;
+    /// Creates an iterator over the full range of a collection.
+    fn into_iter_full_range(self: T) -> Self::IntoIter;
 }
--- a/corelib/src/starknet/storage.cairo
+++ b/corelib/src/starknet/storage.cairo
@@ -161,11 +161,8 @@ mod sub_pointers;
 pub use sub_pointers::{SubPointers, SubPointersForward, SubPointersMut, SubPointersMutForward};
 
 mod vec;
+pub use vec::{IntoIterRange, MutableVecTrait, Vec, VecTrait};
 use vec::{MutableVecIndexView, VecIndexView};
-pub use vec::{
-    MutableVecIntoIterRange, MutableVecTrait, PathableMutableVecIntoIterRange,
-    PathableVecIntoIterRange, Vec, VecIntoIterRange, VecTrait,
-};
 
 /// A pointer to an address in storage, can be used to read and write values, if the generic type
 /// supports it (e.g. basic types like `felt252`).
--- a/corelib/src/starknet/storage.cairo
+++ b/corelib/src/starknet/storage.cairo
@@ -162,7 +162,10 @@ pub use sub_pointers::{SubPointers, SubPointersForward, SubPointersMut, SubPoint
 
 mod vec;
 use vec::{MutableVecIndexView, VecIndexView};
-pub use vec::{MutableVecTrait, Vec, VecTrait};
+pub use vec::{
+    MutableVecTrait, PathableMutableVecIntoIterator, PathableVecIntoIterator, Vec, VecIntoIterator,
+    VecTrait,
+};
 
 /// A pointer to an address in storage, can be used to read and write values, if the generic type
 /// supports it (e.g. basic types like `felt252`).
--- a/corelib/src/starknet/storage.cairo
+++ b/corelib/src/starknet/storage.cairo
@@ -163,8 +163,8 @@ pub use sub_pointers::{SubPointers, SubPointersForward, SubPointersMut, SubPoint
 mod vec;
 use vec::{MutableVecIndexView, VecIndexView};
 pub use vec::{
-    MutableVecTrait, PathableMutableVecIntoIterator, PathableVecIntoIterator, Vec, VecIntoIterator,
-    VecTrait,
+    MutableVecIntoIterRange, MutableVecTrait, PathableMutableVecIntoIterRange,
+    PathableVecIntoIterRange, Vec, VecIntoIterRange, VecTrait,
 };
 
 /// A pointer to an address in storage, can be used to read and write values, if the generic type
--- a/corelib/src/starknet/storage/vec.cairo
+++ b/corelib/src/starknet/storage/vec.cairo
@@ -422,7 +422,7 @@ pub impl VecIntoIterRange<
     }
     #[inline]
     fn into_iter_full_range(self: StoragePath<Vec<T>>) -> Self::IntoIter {
-        VecIter { current_index: (0..VecTraitImpl::len(self)).into_iter(), vec: self }
+        VecIter { current_index: (0..core::num::traits::Bounded::MAX).into_iter(), vec: self }
     }
 }
 
@@ -442,7 +442,7 @@ pub impl PathableVecIntoIterRange<
     #[inline]
     fn into_iter_full_range(self: T) -> Self::IntoIter {
         let vec = self.as_path();
-        VecIter { current_index: (0..vec.len()).into_iter(), vec }
+        VecIter { current_index: (0..core::num::traits::Bounded::MAX).into_iter(), vec }
     }
 }
 
@@ -473,7 +473,9 @@ pub impl MutableVecIntoIterRange<
     }
     #[inline]
     fn into_iter_full_range(self: StoragePath<Mutable<Vec<T>>>) -> Self::IntoIter {
-        MutableVecIter { current_index: (0..MutVecTraitImpl::len(self)).into_iter(), vec: self }
+        MutableVecIter {
+            current_index: (0..core::num::traits::Bounded::MAX).into_iter(), vec: self,
+        }
     }
 }
 
@@ -493,6 +495,6 @@ pub impl PathableMutableVecIntoIterRange<
     #[inline]
     fn into_iter_full_range(self: T) -> Self::IntoIter {
         let vec = self.as_path();
-        MutableVecIter { current_index: (0..vec.len()).into_iter(), vec }
+        MutableVecIter { current_index: (0..core::num::traits::Bounded::MAX).into_iter(), vec }
     }
 }
--- a/corelib/src/starknet/storage/vec.cairo
+++ b/corelib/src/starknet/storage/vec.cairo
@@ -396,3 +396,83 @@ pub impl MutableVecIndexView<
         (*self).at(index)
     }
 }
+
+/// An iterator struct over a Vec in storage.
+#[derive(Drop)]
+pub struct VecIter<T, impl VecImpl: VecTrait<T>> {
+    vec: T,
+    alive: crate::ops::range::RangeIterator<u64>,
+}
+
+impl VecIterator<
+    T, impl VecImpl: VecTrait<T>, +Drop<T>, +Copy<T>,
+> of core::iter::Iterator<VecIter<T>> {
+    type Item = StoragePath<VecImpl::ElementType>;
+    fn next(ref self: VecIter<T>) -> Option<Self::Item> {
+        self.vec.get(self.alive.next()?)
+    }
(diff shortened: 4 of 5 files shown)


orizi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/6941)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @julio4)

___
*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/6941#-OFCYeGOBWdvR0bmq_TM:-OFCYeGOBWdvR0bmq_TN:bjaw6e4) (no related file):*
i'm not sure this is actually a good idea for the time being - as this sort of hides the fact that the result vector may be very very large.

the original code at least very specifically needs to specifiy the fact that iteration over a whole section of a vector is required.
Maybe iterating over a specific range of the storage is a bit more sensible? - so both (mostly) easy but still very explicit?

julio4:

i'm not sure this is actually a good idea for the time being - as this sort of hides the fact that the result vector may be very very large.

the original code at least very specifically needs to specifiy the fact that iteration over a whole section of a vector is required.
Maybe iterating over a specific range of the storage is a bit more sensible? - so both (mostly) easy but still very explicit?
</blockquote></details>

True. But I would argue that the cognitive difference between `0..vec.len()` and `vec.into_iter()` is not that big.
Also iterators provide benefits not only loop syntax sugar, they enable interesting functional patterns and composition.

Maybe we could make size-bounded operations more clear and explicit in the code, for example by forcing iterator instanciation with explicit bound with new patterns like `vec.into_iter().take(vec.len())`/`vec.into_bounded_iter(vec.len())`?

Thoughts?

orizi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/6941)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @julio4)

___
*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/6941#-OFCYeGOBWdvR0bmq_TM:-OFCnPuF8Ust-xeGA1Rj:b-utkbot) (no related file):*

True. But I would argue that the cognitive difference between `0..vec.len()` and `vec.into_iter()` is not that big.
Also iterators provide benefits not only loop syntax sugar, they enable interesting functional patterns and composition.

Maybe we could make size-bounded operations more clear and explicit in the code, for example by forcing iterator instanciation with explicit bound with new patterns like `vec.into_iter().take(vec.len())`/`vec.into_bounded_iter(vec.len())`?

Thoughts?
</blockquote></details>

do note that `for x in vec` would work as well - whish isn't as clear as `.into_vec` to most people probably.

the `vec.into_bounded_iter(vec.len())` sound pretty close to the idea i suggested - `vec.iter_range(4..89)` and `vec.into_full_iter_range()` style thing.

So - not providing `IntoIterator` just specific functions providing iterators sound much better to my mind.

julio4:

do note that `for x in vec` would work as well - whish isn't as clear as `.into_vec` to most people probably.

the `vec.into_bounded_iter(vec.len())` sound pretty close to the idea i suggested - `vec.iter_range(4..89)` and `vec.into_full_iter_range()` style thing.

So - not providing `IntoIterator` just specific functions providing iterators sound much better to my mind.
</blockquote></details>

Will do, this makes sense.

orizi (changes requested):
> Reviewed 2 of 6 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/6941)* status: 2 of 6 files reviewed, 3 unresolved discussions (waiting on @julio4)

___
*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/6941#-OFDPBKS2uNFirOgXRgF:-OFDPBKS2uNFirOgXRgG:b-rfpwr) (no related file):*
@gilbens-starkware 2nd eye.
___
*[`corelib/src/iter/traits/iterator.cairo` line 22 at r2](https://reviewable.io/reviews/starkware-libs/cairo/6941#-OFDPjyy1axVZeE5nzxN:-OFDPjyy1axVZeE5nzxO:bpbnrsx) ([raw file](https://github.com/starkware-libs/cairo/blob/b57049593a7b9e8fb47d83c9ce339b82d5a00de3/corelib/src/iter/traits/iterator.cairo#L22)):*
> ```Cairo
> 
> /// Turn a collection of values into an iterator over a specific range.
> pub trait IntoIterRange<T, I> {
> ```

not sure this trait is required - why not add directly to the `VecTrait`.



_Code quote:_
```Cairo
/// Turn a collection of values into an iterator over a specific range.
pub trait IntoIterRange<T, I> {
```

julio4:
*[`corelib/src/iter/traits/iterator.cairo` line 22 at r2](https://reviewable.io/reviews/starkware-libs/cairo/6941#-OFDPjyy1axVZeE5nzxN:-OFDVxj-6cwuZjtwJaIY:buh01lr) ([raw file](https://github.com/starkware-libs/cairo/blob/b57049593a7b9e8fb47d83c9ce339b82d5a00de3/corelib/src/iter/traits/iterator.cairo#L22)):*

not sure this trait is required - why not add directly to the `VecTrait`.
</blockquote></details>

I thought it could be potentially implemented by other types. If you think it's useless I can move it in `VecTrait`, or a separate `VecIterTrait`

orizi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/6941)* status: 2 of 6 files reviewed, 3 unresolved discussions (waiting on @julio4)

___
*[`corelib/src/iter/traits/iterator.cairo` line 22 at r2](https://reviewable.io/reviews/starkware-libs/cairo/6941#-OFDPjyy1axVZeE5nzxN:-OFGy0cREbdu6ca5Lql5:b-em8a3i) ([raw file](https://github.com/starkware-libs/cairo/blob/b57049593a7b9e8fb47d83c9ce339b82d5a00de3/corelib/src/iter/traits/iterator.cairo#L22)):*

I thought it could be potentially implemented by other types. If you think it's useless I can move it in `VecTrait`, or a separate `VecIterTrait`
</blockquote></details>

i think this is more relvant specifically for storage stuff - since in the usual case probably some combination of `into_iter().skip(10).take(20)` makes more sense.

julio4:
*[`corelib/src/iter/traits/iterator.cairo` line 22 at r2](https://reviewable.io/reviews/starkware-libs/cairo/6941#-OFDPjyy1axVZeE5nzxN:-OFI0tx64r_fPgo337X_:bhhf4v9) ([raw file](https://github.com/starkware-libs/cairo/blob/b57049593a7b9e8fb47d83c9ce339b82d5a00de3/corelib/src/iter/traits/iterator.cairo#L22)):*

i think this is more relvant specifically for storage stuff - since in the usual case probably some combination of `into_iter().skip(10).take(20)` makes more sense.
</blockquote></details>

Should I move the IntoIterRange trait to the vec module and rename it to VecIntoIterRange, or should I extend VecTrait and MutableVecTrait directly with the new iterators methods?

orizi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/6941)* status: 2 of 6 files reviewed, 3 unresolved discussions (waiting on @julio4)

___
*[`corelib/src/iter/traits/iterator.cairo` line 22 at r2](https://reviewable.io/reviews/starkware-libs/cairo/6941#-OFDPjyy1axVZeE5nzxN:-OFKdpdz0QH4P9fkuzwo:b-hr2tdk) ([raw file](https://github.com/starkware-libs/cairo/blob/b57049593a7b9e8fb47d83c9ce339b82d5a00de3/corelib/src/iter/traits/iterator.cairo#L22)):*

Should I move the IntoIterRange trait to the vec module and rename it to VecIntoIterRange, or should I extend VecTrait and MutableVecTrait directly with the new iterators methods?
</blockquote></details>

oh - adding one trait and using it for both is probably nicer and opens more options later - so i like it all and all.

julio4:
*[`corelib/src/iter/traits/iterator.cairo` line 22 at r2](https://reviewable.io/reviews/starkware-libs/cairo/6941#-OFDPjyy1axVZeE5nzxN:-OFREtaU9ZLKRmDE936o:b55hrje) ([raw file](https://github.com/starkware-libs/cairo/blob/b57049593a7b9e8fb47d83c9ce339b82d5a00de3/corelib/src/iter/traits/iterator.cairo#L22)):*
<details><summary><i>Previously, gilbens-starkware (Gil Ben-Shachar) wrote…</i></summary><blockquote>

Notice that this means that the impls will not be automatically found (is not near the trait nor the generic type) and thus it needs to be whereever it is used.
</blockquote></details>

I also think it is nicer but yes the trait need to be used. Is there any pattern to "bring" automatically the trait in the scope of VecTrait?

gilbens-starkware (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/6941)* status: 2 of 6 files reviewed, 4 unresolved discussions (waiting on @julio4 and @orizi)

___
*[`corelib/src/iter/traits/iterator.cairo` line 22 at r2](https://reviewable.io/reviews/starkware-libs/cairo/6941#-OFDPjyy1axVZeE5nzxN:-OFY7KU40RXJUDnJJoEg:b-d0wdxr) ([raw file](https://github.com/starkware-libs/cairo/blob/b57049593a7b9e8fb47d83c9ce339b82d5a00de3/corelib/src/iter/traits/iterator.cairo#L22)):*

I also think it is nicer but yes the trait need to be used. Is there any pattern to "bring" automatically the trait in the scope of VecTrait?
</blockquote></details>

Importing another trait is fine in this case. The problem is if we need to import the impls as well, but your code seems fine in this regard, does it work if you remove the imported impls from the tests (marked in another comment)?
___
*[`crates/cairo-lang-starknet/cairo_level_tests/collections_test.cairo` line 4 at r2](https://reviewable.io/reviews/starkware-libs/cairo/6941#-OFY9Jny4DUeIvJ7nZSs:-OFY9Jny4DUeIvJ7nZSt:undefined) ([raw file](https://github.com/starkware-libs/cairo/blob/b57049593a7b9e8fb47d83c9ce339b82d5a00de3/crates/cairo-lang-starknet/cairo_level_tests/collections_test.cairo#L4)):*
> ```Cairo
>     MutableVecTrait, PathableMutableVecIntoIterRange, PathableVecIntoIterRange, StoragePathEntry,
>     StoragePointerReadAccess, StoragePointerWriteAccess,
> };
> ```





_Suggestion:_
```Cairo
use starknet::storage::{
    MutableVecTrait, StoragePathEntry,
    StoragePointerReadAccess, StoragePointerWriteAccess,
};
use core::iter::IntoIterRange;
```

julio4 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/6941)* status: 2 of 6 files reviewed, 4 unresolved discussions (waiting on @gilbens-starkware and @orizi)

___
*[`crates/cairo-lang-starknet/cairo_level_tests/collections_test.cairo` line 4 at r2](https://reviewable.io/reviews/starkware-libs/cairo/6941#-OFY9Jny4DUeIvJ7nZSs:-OFgHkIy-HCPQYIKXYyC:bb6tru7) ([raw file](https://github.com/starkware-libs/cairo/blob/d27581f22bc6e013c26549f1ea824f97d2d7cd0b/crates/cairo-lang-starknet/cairo_level_tests/collections_test.cairo#L4)):*
> ```Cairo
>     MutableVecTrait, PathableMutableVecIntoIterRange, PathableVecIntoIterRange, StoragePathEntry,
>     StoragePointerReadAccess, StoragePointerWriteAccess,
> };
> ```

See other comment, :done:

orizi (commented):
> Reviewed 3 of 5 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/6941)* status: 4 of 6 files reviewed, 3 unresolved discussions (waiting on @gilbens-starkware and @julio4)

gilbens-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/6941)* status: 2 of 6 files reviewed, 2 unresolved discussions (waiting on @julio4 and @orizi)

___
*[`corelib/src/iter/traits/iterator.cairo` line 22 at r2](https://reviewable.io/reviews/starkware-libs/cairo/6941#-OFDPjyy1axVZeE5nzxN:-OFv3ZgG-pRL8irr4r3d:bpl33sq) ([raw file](https://github.com/starkware-libs/cairo/blob/d27581f22bc6e013c26549f1ea824f97d2d7cd0b/corelib/src/iter/traits/iterator.cairo#L22)):*

When not explicitly importing impls there was an error of duplicated impl.
By making `VecIntoIter` impl less generic it works correctly as it can resolved the correct impl:
```cairo
impl VecIntoIterRange<
    T, impl VecTraitImpl: VecTrait<StoragePath<Vec<T>>>,
> of IntoIterRange<StoragePath<Vec<T>>> { //... }
```

But then I have to edit `PathableVecIntoIterRange` to be able to delegate, and I didn't found a way to correctly bound `Vec<T>` to the `StorageAsPath<Vec<T>>::Value`. I tried something like this but it's not working:
```cairo
impl PathableVecIntoIterRange<
    T,
    impl PathImpl: StorageAsPath<Vec<T>>,
    impl VecTraitImpl: VecTrait<StoragePath<PathImpl::Value>>,
> of IntoIterRange<Vec<T>> {
    type IntoIter = VecIter<StoragePath<PathImpl::Value>, VecTraitImpl>;
    #[inline]
    fn into_iter_range(self: Vec<T>, range: Range<u64>) -> Self::IntoIter {
        VecIntoIterRange::<T, VecTraitImpl>::into_iter_range(self.as_path(), range)
    }
    #[inline]
    fn into_iter_full_range(self: Vec<T>) -> Self::IntoIter {
        VecIntoIterRange::<T, VecTraitImpl>::into_iter_full_range(self.as_path())
    }
}
```

In the end, it was more easy to just duplicate the implementation and not delegate, that's what I did and updated the PR. But if there's any way to correctly do trait bounds I would be curious to know!
</blockquote></details>

That's good.
But in order to implement this trait for future storage object i'd move this trait to `storage.cairo` and import the impls there too (in order for them to be automatically found).

julio4 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/6941)* status: 0 of 6 files reviewed, 2 unresolved discussions (waiting on @gilbens-starkware and @orizi)

___
*[`corelib/src/iter/traits/iterator.cairo` line 22 at r2](https://reviewable.io/reviews/starkware-libs/cairo/6941#-OFDPjyy1axVZeE5nzxN:-OGA-_Z00rLCdXUQQw7M:b-i0xh62) ([raw file](https://github.com/starkware-libs/cairo/blob/d27581f22bc6e013c26549f1ea824f97d2d7cd0b/corelib/src/iter/traits/iterator.cairo#L22)):*
<details><summary><i>Previously, gilbens-starkware (Gil Ben-Shachar) wrote…</i></summary><blockquote>

That's good.
But in order to implement this trait for future storage object i'd move this trait to `storage.cairo` and import the impls there too (in order for them to be automatically found).
</blockquote></details>

I moved IntoIterRange to starknet::storage

gilbens-starkware (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/6941#-:-OGACGJ87lYvkvyk74VJ:bnfp4nl)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @julio4 and @orizi)

orizi (changes requested):
> Reviewed 3 of 5 files at r5, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/6941)* status: 4 of 6 files reviewed, 1 unresolved discussion (waiting on @gilbens-starkware and @julio4)

___
*[`corelib/src/starknet/storage/vec.cairo` line 445 at r6](https://reviewable.io/reviews/starkware-libs/cairo/6941#-OGOjnix34skFp8yZbTx:-OGOjnix34skFp8yZbTy:b-gx2d5r) ([raw file](https://github.com/starkware-libs/cairo/blob/23c32ab10fe9849a5d48d30c14b5a41e46f34616/corelib/src/starknet/storage/vec.cairo#L445)):*
> ```Cairo
>     fn into_iter_full_range(self: T) -> Self::IntoIter {
>         let vec = self.as_path();
>         VecIter { current_index: (0..vec.len()).into_iter(), vec }
> ```

we currently range check every access - so no need for the length read.

alternatively  - check ranges here - and use `update` instead of `get` in the iterator implementation.
(probably better)



_Suggestion:_
```Cairo
        VecIter { current_index: (0..core::num::traits::Bounded::MAX).into_iter(), vec }
```

julio4 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/6941)* status: 4 of 6 files reviewed, 1 unresolved discussion (waiting on @gilbens-starkware and @orizi)

___
*[`corelib/src/starknet/storage/vec.cairo` line 445 at r6](https://reviewable.io/reviews/starkware-libs/cairo/6941#-OGOjnix34skFp8yZbTx:-OGRcdYy3yWcHH6lDMeF:b-9wp9rb) ([raw file](https://github.com/starkware-libs/cairo/blob/23c32ab10fe9849a5d48d30c14b5a41e46f34616/corelib/src/starknet/storage/vec.cairo#L445)):*

we currently range check every access - so no need for the length read.

alternatively  - check ranges here - and use `update` instead of `get` in the iterator implementation.
(probably better)
</blockquote></details>

I'm not sure I understand what you mean by using `update`?
`get` return `None` on out of bound, so I agree we don't need to actually call `len` there.

Then, is there even a need for a range?  We could simply use `cur: u64`, to not have to create a `RangeIterator` struct.

Also range end are exclusive, so technically it can only iterate until `core::num::traits::Bounded::MAX - 1` if not mistaken, not a big issue but still maybe not exactly the specification.

orizi (changes requested):
> Reviewed 1 of 2 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/6941)* status: 4 of 6 files reviewed, 1 unresolved discussion (waiting on @gilbens-starkware and @julio4)

___
*[`corelib/src/starknet/storage/vec.cairo` line 445 at r6](https://reviewable.io/reviews/starkware-libs/cairo/6941#-OGOjnix34skFp8yZbTx:-OGUgkPA-963gmq_BpMX:b-i5t2hp) ([raw file](https://github.com/starkware-libs/cairo/blob/23c32ab10fe9849a5d48d30c14b5a41e46f34616/corelib/src/starknet/storage/vec.cairo#L445)):*

I'm not sure I understand what you mean by using `update`?
`get` return `None` on out of bound, so I agree we don't need to actually call `len` there.

Then, is there even a need for a range?  We could simply use `cur: u64`, to not have to create a `RangeIterator` struct.

Also range end are exclusive, so technically it can only iterate until `core::num::traits::Bounded::MAX - 1` if not mistaken, not a big issue but still maybe not exactly the specification.
</blockquote></details>

testing the len in construction allows to not check the length on iteration.
instead of using `get` to get the entry - skipping the length check, by calling `update` instead of `get` (which just updates the hash-state without checking ranges).

about the overflow thing - i don;t see this becoming an actual issue.

julio4 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/6941)* status: 4 of 6 files reviewed, 1 unresolved discussion (waiting on @gilbens-starkware and @orizi)

___
*[`corelib/src/starknet/storage/vec.cairo` line 445 at r6](https://reviewable.io/reviews/starkware-libs/cairo/6941#-OGOjnix34skFp8yZbTx:-OGVq3KCDt8P8mybJq60:boyzr7t) ([raw file](https://github.com/starkware-libs/cairo/blob/23c32ab10fe9849a5d48d30c14b5a41e46f34616/corelib/src/starknet/storage/vec.cairo#L445)):*

testing the len in construction allows to not check the length on iteration.
instead of using `get` to get the entry - skipping the length check, by calling `update` instead of `get` (which just updates the hash-state without checking ranges).

about the overflow thing - i don;t see this becoming an actual issue.
</blockquote></details>

I see, so keeping `(0..vec.len())` in the iterator construction and directly using `update` from StoragePath. 

But do we have guarantee that the range check holds because it's an append only vec?

Consider this:
```cairo
let mut vec = [];
let mut iter = vec.into_iter_full_range();
vec.append().write(1);
assert!(iter.next().is_some());
```
This should work with `(0..core::num::traits::Bounded::MAX)`/`cur: u64` and  `get` range check at every access, but with `(0..vec.len())` and `update` it would not as the len would be 0 when the iterator is instantiated.

orizi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/6941)* status: 4 of 6 files reviewed, 1 unresolved discussion (waiting on @gilbens-starkware and @julio4)

___
*[`corelib/src/starknet/storage/vec.cairo` line 445 at r6](https://reviewable.io/reviews/starkware-libs/cairo/6941#-OGOjnix34skFp8yZbTx:-OGWEa_X2CEbrpbO3zzx:bizgki0) ([raw file](https://github.com/starkware-libs/cairo/blob/23c32ab10fe9849a5d48d30c14b5a41e46f34616/corelib/src/starknet/storage/vec.cairo#L445)):*

I see, so keeping `(0..vec.len())` in the iterator construction and directly using `update` from StoragePath. 

But do we have guarantee that the range check holds because it's an append only vec?

Consider this:
```cairo
let mut vec = [];
let mut iter = vec.into_iter_full_range();
vec.append().write(1);
assert!(iter.next().is_some());
```
This should work with `(0..core::num::traits::Bounded::MAX)`/`cur: u64` and  `get` range check at every access, but with `(0..vec.len())` and `update` it would not as the len would be 0 when the iterator is instantiated.
</blockquote></details>

so the way it is now is probably better.

in any case, i would expect the `into_iter...` to be consuming, or at least borrowing (which doesn't exist yet) - so that you won't be able to append during the iteration - but maybe sometime in the future.

orizi (approved):
:lgtm:

> Reviewed 1 of 5 files at r3, 1 of 2 files at r7.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/6941#-:-OG_klIc91jRqSoS7avK:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @julio4)

gilbens-starkware (approved):
> Reviewed 1 of 5 files at r3, 2 of 5 files at r5, 1 of 2 files at r7, 2 of 2 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/6941)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @julio4)

PR #6945, opened 2024-12-28
fix rewrites of ordered_hash_map so the key is also rewritten

TomerStarkware:


orizi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/6945)* status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on @TomerStarkware)

___
*[`crates/cairo-lang-semantic/src/expr/inference.rs` line 387 at r1](https://reviewable.io/reviews/starkware-libs/cairo/6945#-OFBpKdn31gmvUe7zqX3:-OFBpKdn31gmvUe7zqX4:bsgo77q) ([raw file](https://github.com/starkware-libs/cairo/blob/225dab9b4ff3b918fc18bb3f68c560e80508b337/crates/cairo-lang-semantic/src/expr/inference.rs#L387)):*
> ```XML
>                 .type_assignment
>                 .iter()
>                 .map(|(k, v)| (*k, inference_id_replacer.rewrite(*v).no_err()))
> ```

this sort of rewrite sounds like the same thing.



_Code quote:_
```XML
            type_assignment: self
                .type_assignment
                .iter()
                .map(|(k, v)| (*k, inference_id_replacer.rewrite(*v).no_err()))
```
___
*[`crates/cairo-lang-semantic/src/substitution.rs` line 213 at r1](https://reviewable.io/reviews/starkware-libs/cairo/6945#-OFBo_fv0bxy6ELlpGfF:-OFBo_fw8apyEPLvUj34:b9ctbdh) ([raw file](https://github.com/starkware-libs/cairo/blob/225dab9b4ff3b918fc18bb3f68c560e80508b337/crates/cairo-lang-semantic/src/substitution.rs#L213)):*
> ```XML
>         *value = temp_value
>             .into_iter()
>             .map(|(mut k, mut v)| {
> ```

This makes map rewrite much heavier - as it actually has to create a new map.

this wasn't this way on purpose, we need to either makes sure this doesn't hurt performance (note the explicit calling for this sort of rewrites in other areas, like the inference rewrite), or just call this heavy version when required.



_Code quote:_
```XML
        let mut result = RewriteResult::NoChange;
        let temp_value = std::mem::take(value);
        *value = temp_value
            .into_iter()
            .map(|(mut k, mut v)| {
```

TomerStarkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/6945)* status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on @orizi)

___
*[`crates/cairo-lang-semantic/src/substitution.rs` line 213 at r1](https://reviewable.io/reviews/starkware-libs/cairo/6945#-OFBo_fv0bxy6ELlpGfF:-OFGqxuf2vZdPZAfBZsY:bebj1ul) ([raw file](https://github.com/starkware-libs/cairo/blob/225dab9b4ff3b918fc18bb3f68c560e80508b337/crates/cairo-lang-semantic/src/substitution.rs#L213)):*

This makes map rewrite much heavier - as it actually has to create a new map.

this wasn't this way on purpose, we need to either makes sure this doesn't hurt performance (note the explicit calling for this sort of rewrites in other areas, like the inference rewrite), or just call this heavy version when required.
</blockquote></details>

Now I am only cloning the keys when there is no rewrite, 
also we only use this method with copiable keys so we can change T to be copiable
___
*[`crates/cairo-lang-semantic/src/expr/inference.rs` line 387 at r1](https://reviewable.io/reviews/starkware-libs/cairo/6945#-OFBpKdn31gmvUe7zqX3:-OFGkL8z63abWurRDG_K:b1as7q3) ([raw file](https://github.com/starkware-libs/cairo/blob/225dab9b4ff3b918fc18bb3f68c560e80508b337/crates/cairo-lang-semantic/src/expr/inference.rs#L387)):*

this sort of rewrite sounds like the same thing.
</blockquote></details>

here the key is not rewritable

orizi (approved):
:lgtm:

> Reviewed 2 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/6945#-:-OFGxvep0JHTxwRG9OJ-:bnfp4nl)* status: 2 of 3 files reviewed, 1 unresolved discussion (waiting on @TomerStarkware)

___
*[`crates/cairo-lang-semantic/src/substitution.rs` line 213 at r1](https://reviewable.io/reviews/starkware-libs/cairo/6945#-OFBo_fv0bxy6ELlpGfF:-OFGvMPGDlfsejgetsC-:ba6z04j) ([raw file](https://github.com/starkware-libs/cairo/blob/225dab9b4ff3b918fc18bb3f68c560e80508b337/crates/cairo-lang-semantic/src/substitution.rs#L213)):*
<details><summary><i>Previously, TomerStarkware wrote…</i></summary><blockquote>

Now I am only cloning the keys when there is no rewrite, 
also we only use this method with copiable keys so we can change T to be copiable
</blockquote></details>

do note that you should know this from the type of the keys (if they have 'identity-rewrite') - but maybe that's enough.

orizi (approved):
> Reviewed 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/6945)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @TomerStarkware)

PR #6949, opened 2024-12-28
feat(corelib): iter::adapters::Map

julio4:
This is mostly an exploratory PR to test implementation of iterator adapters, i.e. functions that take an `Iterator` to return another `Iterator`. It's really useful as any type that implement `IntoIterator` can benefits from these adapters, and it also enables function chaining on iterators.

Added a first implementation of `iter::adapters::Map`:

```cairo
let mut i = 1;
for elem in (1..4_u8).into_iter().map(|x| 2 * x).map(|x| x + 1) {
    assert_eq!(elem, i * 2 + 1);
    i += 1;
}
```

---

I'm curious to know if an implementation of `ops::FnMut` would be possible, to be able to perform function calls with a reference? With it I believe the efficiency of these adapters can be greatly improved to avoid excessive copy/drop.

Feedback welcome on both the current implementation and potential optimization

### DIFF
--- a/corelib/src/iter.cairo
+++ b/corelib/src/iter.cairo
@@ -1,2 +1,3 @@
 mod traits;
+mod adapters;
 pub use traits::{IntoIterator, Iterator};
--- a/corelib/src/iter.cairo
+++ b/corelib/src/iter.cairo
@@ -54,9 +54,9 @@
 //! [`Option::Some(Item)`]: Option::Some
 //! [`next`]: Iterator::next
 //!
-//! # Iterators in Cairo
+//! # Forms of iteration
 //!
-//! There are currently one common method which can create iterators from a collection:
+//! There is currently only one common method which can create iterators from a collection:
 //!
 //! * `into_iter()`, which iterates over `T`.
 //!
@@ -138,19 +138,19 @@
 //! ```
 //!
 //! This will print the numbers one through five, each on their own line. But
-//! you'll notice something here: we never called anything on our vector to
+//! you'll notice something here: we never called anything on our array to
 //! produce an iterator. What gives?
 //!
-//! There's a trait in the standard library for converting something into an
+//! There's a trait in the core library for converting something into an
 //! iterator: [`IntoIterator`]. This trait has one method, [`into_iter`],
 //! which converts the thing implementing [`IntoIterator`] into an iterator.
 //! Let's take a look at that `for` loop again, and what the compiler converts
 //! it into:
 //!
-//! [`into_iter`]: IntoIterator::into_iter
+//! [`into_iter`]: core::iter::IntoIterator::into_iter
 //!
 //! ```
-//! let values = vec![1, 2, 3, 4, 5];
+//! let values = array![1, 2, 3, 4, 5];
 //!
 //! for x in values {
 //!     println!("{x}");
@@ -182,7 +182,7 @@
 //! that returns, calling [`next`] over and over until we see a `Option::None`. At
 //! that point, we `break` out of the loop, and we're done iterating.
 //!
-//! There's one more subtle bit here: the standard library contains an
+//! There's one more subtle bit here: the core library contains an
 //! interesting implementation of [`IntoIterator`]:
 //!
 //! ```ignore (only-for-syntax-highlight)
@@ -204,9 +204,6 @@
 //!
 //! The only adapter for now is [`map`].
 //! For more, see the [`map`] documentation.
-//!
-//! If an iterator adapter panics, the iterator will be in an unspecified (but
-//! memory safe) state.
 //!
 //! [`map`]: Iterator::map
 //!
--- a/corelib/src/iter.cairo
+++ b/corelib/src/iter.cairo
@@ -216,9 +216,8 @@
 //! method calls a closure on each element it iterates over:
 //!
 //! ```
-//! # #![allow(unused_must_use)]
 //! let v = array![1, 2, 3, 4, 5];
-//! v.into_iter().map(|x| println!("{x}"));
+//! let _ = v.into_iter().map(|x| println!("{x}"));
 //! ```
 //!
 //! This will not print any values, as we only created an iterator, rather than
--- a/corelib/src/iter.cairo
+++ b/corelib/src/iter.cairo
@@ -1,3 +1,237 @@
+//! Composable external iteration.
+//!
+//! If you've found yourself with a collection of some kind, and needed to
+//! perform an operation on the elements of said collection, you'll quickly run
+//! into 'iterators'. Iterators are heavily used in idiomatic code, so
+//! it's worth becoming familiar with them.
+//!
+//! Before explaining more, let's talk about how this module is structured:
+//!
+//! # Organization
+//!
+//! This module is largely organized by type:
+//!
+//! * [Traits] are the core portion: these traits define what kind of iterators
+//!   exist and what you can do with them. The methods of these traits are worth
+//!   putting some extra study time into.
+//! * [Functions] provide some helpful ways to create some basic iterators.
+//! * [Structs] are often the return types of the various methods on this
+//!   module's traits. You'll usually want to look at the method that creates
+//!   the `struct`, rather than the `struct` itself. For more detail about why,
+//!   see '[Implementing Iterator](#implementing-iterator)'.
+//!
+//! [Traits]: #traits
+//! [Functions]: #functions
+//! [Structs]: #structs
+//!
+//! That's it! Let's dig into iterators.
+//!
+//! # Iterator
+//!
+//! The heart and soul of this module is the [`Iterator`] trait. The core of
+//! [`Iterator`] looks like this:
+//!
+//! ```
+//! trait Iterator {
+//!     type Item;
+//!     fn next(ref self) -> Option<Self::Item>;
+//! }
+//! ```
+//!
+//! An iterator has a method, [`next`], which when called, returns an
+//! <code>[Option]\<Item></code>. Calling [`next`] will return [`Option::Some(Item)`] as long as
+//! there are elements, and once they've all been exhausted, will return `Option::None` to
+//! indicate that iteration is finished.
+//!
+//! [`Iterator`]'s full definition includes a number of other methods as well,
+//! but they are default methods, built on top of [`next`], and so you get
+//! them for free.
+//!
+//! Iterators are also composable, and it's common to chain them together to do
+//! more complex forms of processing. See the [Adapters](#adapters) section
+//! below for more details.
+//!
+//! [`Option::Some(Item)`]: Option::Some
+//! [`next`]: Iterator::next
+//!
+//! # Iterators in Cairo
+//!
+//! There are currently one common method which can create iterators from a collection:
+//!
+//! * `into_iter()`, which iterates over `T`.
+//!
+//! # Implementing Iterator
+//!
+//! Creating an iterator of your own involves two steps: creating a `struct` to
+//! hold the iterator's state, and then implementing [`Iterator`] for that `struct`.
+//! This is why there are so many `struct`s in this module: there is one for
+//! each iterator and iterator adapter.
+//!
+//! Let's make an iterator named `Counter` which counts from `1` to `5`:
+//!
+//! ```
+//! // First, the struct:
+//!
+//! /// An iterator which counts from one to five
+//! #[derive(Drop)]
+//! struct Counter {
+//!     count: usize,
+//! }
+//!
+//! // we want our count to start at one, so let's add a new() method to help.
+//! // This isn't strictly necessary, but is convenient. Note that we start
+//! // `count` at zero, we'll see why in `next()`'s implementation below.
+//! #[generate_trait]
+//! impl CounterImpl of CounterTrait {
+//!     fn new() -> Counter {
+//!         Counter { count: 0 }
+//!     }
+//! }
+//!
+//! // Then, we implement `Iterator` for our `Counter`:
+//!
+//! impl CounterIter of core::iter::Iterator<Counter> {
+//!     // we will be counting with usize
+//!     type Item = usize;
+//!
+//!     // next() is the only required method
+//!     fn next(ref self: Counter) -> Option<Self::Item> {
+//!         // Increment our count. This is why we started at zero.
+//!         self.count += 1;
+//!
+//!         // Check to see if we've finished counting or not.
+//!         if self.count < 6 {
+//!             Option::Some(self.count)
+//!         } else {
+//!             Option::None
+//!         }
+//!     }
+//! }
+//!
+//! // And now we can use it!
+//!
+//! let mut counter = CounterTrait::new();
+//!
+//! assert!(counter.next() == Option::Some(1));
+//! assert!(counter.next() == Option::Some(2));
+//! assert!(counter.next() == Option::Some(3));
+//! assert!(counter.next() == Option::Some(4));
+//! assert!(counter.next() == Option::Some(5));
+//! assert!(counter.next() == Option::None);
+//! ```
+//!
+//! Calling [`next`] this way gets repetitive. Cairo has a construct which can
+//! call [`next`] on your iterator, until it reaches `Option::None`. Let's go over that
+//! next.
+//!
+//! # `for` loops and `IntoIterator`
+//!
+//! Cairo's `for` loop syntax is actually sugar for iterators. Here's a basic
+//! example of `for`:
+//!
+//! ```
+//! let values = array![1, 2, 3, 4, 5];
+//!
+//! for x in values {
+//!     println!("{x}");
+//! }
+//! ```
+//!
+//! This will print the numbers one through five, each on their own line. But
+//! you'll notice something here: we never called anything on our vector to
+//! produce an iterator. What gives?
+//!
+//! There's a trait in the standard library for converting something into an
+//! iterator: [`IntoIterator`]. This trait has one method, [`into_iter`],
+//! which converts the thing implementing [`IntoIterator`] into an iterator.
+//! Let's take a look at that `for` loop again, and what the compiler converts
+//! it into:
+//!
+//! [`into_iter`]: IntoIterator::into_iter
+//!
+//! ```
+//! let values = vec![1, 2, 3, 4, 5];
+//!
+//! for x in values {
+//!     println!("{x}");
+//! }
+//! ```
+//!
+//! Cairo de-sugars this into:
+//!
+//! ```
+//! let values = array![1, 2, 3, 4, 5];
+//! {
+//!     let mut iter = IntoIterator::into_iter(values);
+//!     let result = loop {
+//!             let mut next = 0;
+//!             match iter.next() {
+//!                 Option::Some(val) => next = val,
+//!                 Option::None => {
+//!                     break;
+//!                 },
+//!             };
+//!             let x = next;
+//!             let () = { println!("{x}"); };
+//!         };
+//!     result
+//! }
+//! ```
+//!
+//! First, we call `into_iter()` on the value. Then, we match on the iterator
+//! that returns, calling [`next`] over and over until we see a `Option::None`. At
+//! that point, we `break` out of the loop, and we're done iterating.
+//!
+//! There's one more subtle bit here: the standard library contains an
+//! interesting implementation of [`IntoIterator`]:
+//!
+//! ```ignore (only-for-syntax-highlight)
+//! impl IteratorIntoIterator<T, +Iterator<T>> of IntoIterator<T>
+//! ```
+//!
+//! In other words, all [`Iterator`]s implement [`IntoIterator`], by just
+//! returning themselves. This means two things:
+//!
+//! 1. If you're writing an [`Iterator`], you can use it with a `for` loop.
+//! 2. If you're creating a collection, implementing [`IntoIterator`] for it
+//!    will allow your collection to be used with the `for` loop.
+//!
+//! # Adapters
+//!
+//! Functions which take an [`Iterator`] and return another [`Iterator`] are
+//! often called 'iterator adapters', as they're a form of the 'adapter
+//! pattern'.
+//!
+//! The only adapter for now is [`map`].
+//! For more, see the [`map`] documentation.
+//!
+//! If an iterator adapter panics, the iterator will be in an unspecified (but
+//! memory safe) state.
+//!
+//! [`map`]: Iterator::map
+//!
+//! # Laziness
+//!
+//! Iterators (and iterator [adapters](#adapters)) are *lazy*. This means that
+//! just creating an iterator doesn't _do_ a whole lot. Nothing really happens
+//! until you call [`next`]. This is sometimes a source of confusion when
+//! creating an iterator solely for its side effects. For example, the [`map`]
+//! method calls a closure on each element it iterates over:
+//!
+//! ```
+//! # #![allow(unused_must_use)]
+//! let v = array![1, 2, 3, 4, 5];
+//! v.into_iter().map(|x| println!("{x}"));
+//! ```
+//!
+//! This will not print any values, as we only created an iterator, rather than
+//! using it. The compiler will warn us about this kind of behavior:
+//!
+//! ```text
+//! Unhandled `#[must_use]` type
+//! ```
+//!
+//! [`map`]: Iterator::map
 mod traits;
 mod adapters;
 pub use traits::{IntoIterator, Iterator};
--- a/corelib/src/iter.cairo
+++ b/corelib/src/iter.cairo
@@ -228,6 +228,6 @@
 //! ```
 //!
 //! [`map`]: Iterator::map
-mod traits;
 mod adapters;
+mod traits;
 pub use traits::{IntoIterator, Iterator};
--- a/corelib/src/iter/adapters.cairo
+++ b/corelib/src/iter/adapters.cairo
@@ -1,2 +1,2 @@
 mod map;
-pub use map::{Map, MapTrait};
+pub use map::{Map, mapped_iterator};
--- a/corelib/src/iter/adapters.cairo
+++ b/corelib/src/iter/adapters.cairo
@@ -1,2 +1,4 @@
 mod map;
-pub use map::{Map, mapped_iterator};
+pub use map::Map;
+#[allow(unused_imports)]
+pub(crate) use map::mapped_iterator;
--- a/corelib/src/iter/adapters.cairo
+++ b/corelib/src/iter/adapters.cairo
@@ -1,0 +1,2 @@
+mod map;
+pub use map::{Map, MapTrait};
--- a/corelib/src/iter/adapters/map.cairo
+++ b/corelib/src/iter/adapters/map.cairo
@@ -24,16 +24,14 @@ pub(crate) impl MapImpl<I, F> of MapTrait<I, F> {
 
 impl MapIterator<
     I,
-    B,
     F,
     impl TIter: Iterator<I>,
-    +core::ops::FnOnce<F, (TIter::Item,)>[Output: B],
+    impl func: core::ops::Fn<F, (TIter::Item,)>,
     +Destruct<I>,
     +Destruct<F>,
-    +Copy<F>,
 > of Iterator<Map<I, F>> {
-    type Item = B;
-    fn next(ref self: Map<I, F>) -> Option<B> {
-        self.iter.next().map(self.f)
+    type Item = func::Output;
+    fn next(ref self: Map<I, F>) -> Option<func::Output> {
+        self.iter.next().map(@self.f)
     }
 }
--- a/corelib/src/iter/adapters/map.cairo
+++ b/corelib/src/iter/adapters/map.cairo
@@ -28,8 +28,8 @@ impl MapIterator<
     F,
     impl TIter: Iterator<I>,
     +core::ops::FnOnce<F, (TIter::Item,)>[Output: B],
-    +Drop<I>,
-    +Drop<F>,
+    +Destruct<I>,
+    +Destruct<F>,
     +Copy<F>,
 > of Iterator<Map<I, F>> {
     type Item = B;
--- a/corelib/src/iter/adapters/map.cairo
+++ b/corelib/src/iter/adapters/map.cairo
@@ -1,7 +1,7 @@
 use crate::iter::Iterator;
 
 #[must_use]
-#[derive(Clone)]
+#[derive(Drop, Clone)]
 pub struct Map<I, F> {
     iter: I,
     f: F,
@@ -15,8 +15,8 @@ pub(crate) impl MapImpl<I, F> of MapTrait<I, F> {
 }
 
 impl MapIterator<
-    B,
     I,
+    B,
     F,
     impl TIter: Iterator<I>,
     +core::ops::FnOnce<F, (TIter::Item,)>[Output: B],
--- a/corelib/src/iter/adapters/map.cairo
+++ b/corelib/src/iter/adapters/map.cairo
@@ -15,11 +15,8 @@ pub struct Map<I, F> {
     f: F,
 }
 
-#[generate_trait]
-pub(crate) impl MapImpl<I, F> of MapTrait<I, F> {
-    fn new(iter: I, f: F) -> Map<I, F> {
-        Map { iter, f }
-    }
+pub(crate) fn mapped_iterator<I, F>(iter: I, f: F) -> Map<I, F> {
+    Map { iter, f }
 }
 
 impl MapIterator<
--- a/corelib/src/iter/adapters/map.cairo
+++ b/corelib/src/iter/adapters/map.cairo
@@ -1,5 +1,13 @@
 use crate::iter::Iterator;
 
+/// An iterator that maps the values of `iter` with `f`.
+///
+/// This `struct` is created by the [`map`] method on [`Iterator`]. See its
+/// documentation for more.
+///
+/// [`map`]: Iterator::map
+/// [`Iterator`]: core::iter::Iterator
+///
 #[must_use]
 #[derive(Drop, Clone)]
 pub struct Map<I, F> {
--- a/corelib/src/iter/adapters/map.cairo
+++ b/corelib/src/iter/adapters/map.cairo
@@ -1,0 +1,31 @@
+use crate::iter::Iterator;
+
+#[must_use]
+#[derive(Clone)]
+pub struct Map<I, F> {
+    iter: I,
+    f: F,
+}
+
+#[generate_trait]
+pub(crate) impl MapImpl<I, F> of MapTrait<I, F> {
+    fn new(iter: I, f: F) -> Map<I, F> {
+        Map { iter, f }
+    }
+}
+
+impl MapIterator<
+    B,
+    I,
+    F,
+    impl TIter: Iterator<I>,
+    +core::ops::FnOnce<F, (TIter::Item,)>[Output: B],
+    +Drop<I>,
+    +Drop<F>,
+    +Copy<F>,
+> of Iterator<Map<I, F>> {
+    type Item = B;
+    fn next(ref self: Map<I, F>) -> Option<B> {
+        self.iter.next().map(self.f)
+    }
+}
--- a/corelib/src/iter/adapters/map.cairo
+++ b/corelib/src/iter/adapters/map.cairo
@@ -1,5 +1,3 @@
-use crate::iter::Iterator;
-
 /// An iterator that maps the values of `iter` with `f`.
 ///
 /// This `struct` is created by the [`map`] method on [`Iterator`]. See its
@@ -15,7 +13,7 @@ pub struct Map<I, F> {
     f: F,
 }
 
-pub(crate) fn mapped_iterator<I, F>(iter: I, f: F) -> Map<I, F> {
+pub fn mapped_iterator<I, F>(iter: I, f: F) -> Map<I, F> {
     Map { iter, f }
 }
 
--- a/corelib/src/iter/traits/iterator.cairo
+++ b/corelib/src/iter/traits/iterator.cairo
@@ -76,9 +76,8 @@ pub trait Iterator<T> {
     /// If you're doing some sort of side effect, prefer `for` to `map()`:
     ///
     /// ```
-    /// # #![allow(unused_must_use)]
     /// // don't do this:
-    /// (0..5_usize).into_iter().map(|x| println!("{x}"));
+    /// let _ = (0..5_usize).into_iter().map(|x| println!("{x}"));
     ///
     /// // it won't even execute, as it is lazy. Cairo will warn you about this.
     ///
--- a/corelib/src/iter/traits/iterator.cairo
+++ b/corelib/src/iter/traits/iterator.cairo
@@ -1,4 +1,4 @@
-use crate::iter::adapters::{Map, MapTrait};
+use crate::iter::adapters::{Map, mapped_iterator};
 
 /// A trait for dealing with iterators.
 ///
@@ -98,6 +98,6 @@ pub trait Iterator<T> {
     >(
         self: T, f: F,
     ) -> Map<T, F> {
-        MapTrait::new(self, f)
+        mapped_iterator(self, f)
     }
 }
--- a/corelib/src/iter/traits/iterator.cairo
+++ b/corelib/src/iter/traits/iterator.cairo
@@ -12,7 +12,8 @@ pub trait Iterator<T> {
     fn map<
         B,
         F,
-        +core::ops::FnOnce<F, (Self::Item,)>[Output: B],
+        impl TIter: Self,
+        +core::ops::FnOnce<F, (TIter::Item,)>[Output: B],
         +Drop<T>,
         +Drop<F>,
         +Copy<F>,
--- a/corelib/src/iter/traits/iterator.cairo
+++ b/corelib/src/iter/traits/iterator.cairo
@@ -1,3 +1,5 @@
+use crate::iter::adapters::{Map, MapTrait};
+
 /// An iterator over a collection of values.
 pub trait Iterator<T> {
     /// The type of the elements being iterated over.
@@ -5,4 +7,18 @@ pub trait Iterator<T> {
 
     /// Advance the iterator and return the next value.
     fn next(ref self: T) -> Option<Self::Item>;
+
+    #[inline]
+    fn map<
+        B,
+        F,
+        +core::ops::FnOnce<F, (Self::Item,)>[Output: B],
+        +Drop<T>,
+        +Drop<F>,
+        +Copy<F>,
+    >(
+        self: T, f: F,
+    ) -> Map<T, F> {
+        MapTrait::new(self, f)
+    }
 }
--- a/corelib/src/iter/traits/iterator.cairo
+++ b/corelib/src/iter/traits/iterator.cairo
@@ -79,7 +79,8 @@ pub trait Iterator<T> {
     /// // don't do this:
     /// let _ = (0..5_usize).into_iter().map(|x| println!("{x}"));
     ///
-    /// // it won't even execute, as it is lazy. Cairo will warn you about this.
+    /// // it won't even execute, as it is lazy. Cairo will warn you about this if not specifically
+    /// ignored, as is done here.
     ///
     /// // Instead, use for:
     /// for x in 0..5_usize {
--- a/corelib/src/iter/traits/iterator.cairo
+++ b/corelib/src/iter/traits/iterator.cairo
@@ -89,13 +89,7 @@ pub trait Iterator<T> {
     /// ```
     #[inline]
     fn map<
-        B,
-        F,
-        impl TIter: Self,
-        +core::ops::FnOnce<F, (TIter::Item,)>[Output: B],
-        +Drop<T>,
-        +Drop<F>,
-        +Copy<F>,
+        B, F, impl TIter: Self, +core::ops::Fn<F, (TIter::Item,)>[Output: B], +Drop<T>, +Drop<F>,
     >(
         self: T, f: F,
     ) -> Map<T, F> {
--- a/corelib/src/iter/traits/iterator.cairo
+++ b/corelib/src/iter/traits/iterator.cairo
@@ -1,13 +1,92 @@
 use crate::iter::adapters::{Map, MapTrait};
 
-/// An iterator over a collection of values.
+/// A trait for dealing with iterators.
+///
+/// This is the main iterator trait. For more about the concept of iterators
+/// generally, please see the [module-level documentation]. In particular, you
+/// may want to know how to [implement `Iterator`][impl].
+///
+/// [module-level documentation]: crate::iter
+/// [impl]: crate::iter#implementing-iterator
 pub trait Iterator<T> {
     /// The type of the elements being iterated over.
     type Item;
 
-    /// Advance the iterator and return the next value.
+
+    /// Advances the iterator and returns the next value.
+    ///
+    /// Returns [`Option::None`] when iteration is finished.
+    ///
+    /// [`Option::Some(Item)`]: Option::Some
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// let a = array![1, 2, 3];
+    ///
+    /// let mut iter = a.into_iter();
+    ///
+    /// // A call to next() returns the next value...
+    /// assert!(Option::Some(1) == iter.next());
+    /// assert!(Option::Some(2) == iter.next());
+    /// assert!(Option::Some(3) == iter.next());
+    ///
+    /// // ... and then None once it's over.
+    /// assert!(Option::None == iter.next());
+    ///
+    /// // More calls may or may not return `Option::None`. Here, they always will.
+    /// assert!(Option::None == iter.next());
+    /// assert!(Option::None == iter.next());
+    /// ```
     fn next(ref self: T) -> Option<Self::Item>;
 
+
+    /// Takes a closure and creates an iterator which calls that closure on each
+    /// element.
+    ///
+    /// `map()` transforms one iterator into another, by means of its argument:
+    /// something that implements [`FnOnce`]. It produces a new iterator which
+    /// calls this closure on each element of the original iterator.
+    ///
+    /// If you are good at thinking in types, you can think of `map()` like this:
+    /// If you have an iterator that gives you elements of some type `A`, and
+    /// you want an iterator of some other type `B`, you can use `map()`,
+    /// passing a closure that takes an `A` and returns a `B`.
+    ///
+    /// `map()` is conceptually similar to a `for` loop. However, as `map()` is
+    /// lazy, it is best used when you're already working with other iterators.
+    /// If you're doing some sort of looping for a side effect, it's considered
+    /// more idiomatic to use `for` than `map()`.
+    ///
+    /// # Examples
+    ///
+    /// Basic usage:
+    ///
+    /// ```
+    /// let a = array![1, 2, 3];
+    ///
+    /// let mut iter = a.into_iter().map(|x| 2 * x);
+    ///
+    /// assert!(iter.next() == Option::Some(2));
+    /// assert!(iter.next() == Option::Some(4));
+    /// assert!(iter.next() == Option::Some(6));
+    /// assert!(iter.next() == Option::None);
+    /// ```
+    ///
+    /// If you're doing some sort of side effect, prefer `for` to `map()`:
+    ///
+    /// ```
+    /// # #![allow(unused_must_use)]
+    /// // don't do this:
+    /// (0..5_usize).into_iter().map(|x| println!("{x}"));
+    ///
+    /// // it won't even execute, as it is lazy. Cairo will warn you about this.
+    ///
+    /// // Instead, use for:
+    /// for x in 0..5_usize {
+    ///     println!("{x}");
+    /// }
+    /// ```
     #[inline]
     fn map<
         B,
--- a/corelib/src/ops/function.cairo
+++ b/corelib/src/ops/function.cairo
@@ -23,6 +23,15 @@ impl FnOnceImpl<T, Args, +Destruct<T>, +Fn<T, Args>> of FnOnce<T, Args> {
     }
 }
 
+/// An implementation of `FnOnce` when `Fn` is implemented and the receiver is by-reference.
+/// Makes sure we can always pass an `Fn` to a function that expects an `FnOnce`.
+impl FnOnceSnapImpl<T, Args, +Destruct<T>, +Fn<T, Args>> of FnOnce<@T, Args> {
+    type Output = Fn::<T, Args>::Output;
+    fn call(self: @T, args: Args) -> Self::Output {
+        Fn::call(self, args)
+    }
+}
+
 /// The version of the call operator that takes a by-snapshot receiver.
 ///
 /// Instances of `Fn` can be called multiple times.
--- a/corelib/src/ops/function.cairo
+++ b/corelib/src/ops/function.cairo
@@ -23,15 +23,6 @@ impl FnOnceImpl<T, Args, +Destruct<T>, +Fn<T, Args>> of FnOnce<T, Args> {
     }
 }
 
-/// An implementation of `FnOnce` when `Fn` is implemented and the receiver is by-reference.
-/// Makes sure we can always pass an `Fn` to a function that expects an `FnOnce`.
-impl FnOnceSnapImpl<T, Args, +Destruct<T>, +Fn<T, Args>> of FnOnce<@T, Args> {
-    type Output = Fn::<T, Args>::Output;
-    fn call(self: @T, args: Args) -> Self::Output {
-        Fn::call(self, args)
-    }
-}
-
 /// The version of the call operator that takes a by-snapshot receiver.
 ///
 /// Instances of `Fn` can be called multiple times.
--- a/corelib/src/option.cairo
+++ b/corelib/src/option.cairo
@@ -508,6 +508,10 @@ pub trait OptionTrait<T> {
         self: Option<T>, f: F,
     ) -> T;
 
+    /////////////////////////////////////////////////////////////////////////
+    // Transforming contained values
+    /////////////////////////////////////////////////////////////////////////
+
     /// Maps an `Option<T>` to `Option<U>` by applying a function to a contained value (if `Some`)
     /// or returns `Option::None` (if `None`).
     ///
@@ -718,7 +722,7 @@ pub impl OptionTraitImpl<T> of OptionTrait<T> {
     }
 
     #[inline]
-    fn map<U, F, +Drop<F>, +core::ops::FnOnce<F, (T,)>[Output: U]>(
+    fn map<U, F, +Destruct<F>, +core::ops::FnOnce<F, (T,)>[Output: U]>(
         self: Option<T>, f: F,
     ) -> Option<U> {
         match self {
--- a/corelib/src/option.cairo
+++ b/corelib/src/option.cairo
@@ -508,12 +508,8 @@ pub trait OptionTrait<T> {
         self: Option<T>, f: F,
     ) -> T;
 
-    /////////////////////////////////////////////////////////////////////////
-    // Transforming contained values
-    /////////////////////////////////////////////////////////////////////////
-
     /// Maps an `Option<T>` to `Option<U>` by applying a function to a contained value (if `Some`)
-    /// or returns `None` (if `None`).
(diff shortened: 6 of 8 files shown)


orizi (changes requested):
> Reviewed 2 of 6 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/6949)* status: 2 of 6 files reviewed, 4 unresolved discussions (waiting on @julio4)

___
*[`corelib/src/iter/traits/iterator.cairo` line 12 at r1](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFbpruiD02kb68ct6Ov:-OFbpruiD02kb68ct6Ow:bb7f2bf) ([raw file](https://github.com/starkware-libs/cairo/blob/f64ccf6bbf3e643c06f55ef392632d771ec975ea/corelib/src/iter/traits/iterator.cairo#L12)):*
> ```Cairo
> 
>     #[inline]
>     fn map<
> ```

doc.
___
*[`corelib/src/iter/adapters/map.cairo` line 2 at r1](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFbpmpzErkp0MYeQSLB:-OFbpmpzErkp0MYeQSLC:bvi7ldh) ([raw file](https://github.com/starkware-libs/cairo/blob/f64ccf6bbf3e643c06f55ef392632d771ec975ea/corelib/src/iter/adapters/map.cairo#L2)):*
> ```Cairo
> use crate::iter::Iterator;
> ```

doc all around.
___
*[`corelib/src/iter/adapters/map.cairo` line 24 at r1](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFbqO2Z1BSISYNjjijO:-OFbqO2_65T47uFIcOOQ:bi8k983) ([raw file](https://github.com/starkware-libs/cairo/blob/f64ccf6bbf3e643c06f55ef392632d771ec975ea/corelib/src/iter/adapters/map.cairo#L24)):*
> ```Cairo
>     +core::ops::FnOnce<F, (TIter::Item,)>[Output: B],
>     +Drop<I>,
>     +Drop<F>,
> ```

why Drop and not Destruct all around?



_Code quote:_
```Cairo
    +Drop<I>,
    +Drop<F>,
```
___
*[`corelib/src/iter/adapters/map.cairo` line 25 at r1](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFbq0nJ9xqajlHXNBGI:-OFbq0nJ9xqajlHXNBGJ:b-3fwwcy) ([raw file](https://github.com/starkware-libs/cairo/blob/f64ccf6bbf3e643c06f55ef392632d771ec975ea/corelib/src/iter/adapters/map.cairo#L25)):*
> ```Cairo
>     +Drop<I>,
>     +Drop<F>,
>     +Copy<F>,
> ```

why is Copy required?



_Code quote:_
```Cairo
    +Copy<F>,
```

julio4:
*[`corelib/src/iter/adapters/map.cairo` line 2 at r1](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFbpmpzErkp0MYeQSLB:-OFgSXaU6Ji8kvrsz-X9:b-896fix) ([raw file](https://github.com/starkware-libs/cairo/blob/f64ccf6bbf3e643c06f55ef392632d771ec975ea/corelib/src/iter/adapters/map.cairo#L2)):*

doc all around.
</blockquote></details>

Done.

julio4:
*[`corelib/src/iter/adapters/map.cairo` line 25 at r1](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFbq0nJ9xqajlHXNBGI:-OFgUylX9mSRCefH123l:b6sstkt) ([raw file](https://github.com/starkware-libs/cairo/blob/f64ccf6bbf3e643c06f55ef392632d771ec975ea/corelib/src/iter/adapters/map.cairo#L25)):*

why is Copy required?
</blockquote></details>

When calling `map(self.f)`, self.f is of type `FnOnce` so it is moved:
```cairo
    fn map<U, F, +Destruct<F>, +core::ops::FnOnce<F, (T,)>[Output: U]>(
        self: Option<T>, f: F,
    ) -> Option<U>;
```
Maybe we could use `Fn` instead of `FnOnce` to not have to copy one new instance of f at every iteration

julio4:
*[`corelib/src/iter/adapters/map.cairo` line 24 at r1](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFbqO2Z1BSISYNjjijO:-OFgXFdt6gFnxsh2OpHY:bml962c) ([raw file](https://github.com/starkware-libs/cairo/blob/f64ccf6bbf3e643c06f55ef392632d771ec975ea/corelib/src/iter/adapters/map.cairo#L24)):*

why Drop and not Destruct all around?
</blockquote></details>

I am unsure what is the best. But I changed to `Destruct`

julio4 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/6949)* status: 1 of 7 files reviewed, 4 unresolved discussions (waiting on @orizi)

___
*[`corelib/src/iter/traits/iterator.cairo` line 12 at r1](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFbpruiD02kb68ct6Ov:-OFgXEOVB7CiHg53Nsnj:b-896fix) ([raw file](https://github.com/starkware-libs/cairo/blob/f64ccf6bbf3e643c06f55ef392632d771ec975ea/corelib/src/iter/traits/iterator.cairo#L12)):*

doc.
</blockquote></details>

Done.

orizi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/6949)* status: 1 of 7 files reviewed, 2 unresolved discussions (waiting on @julio4)

___
*[`corelib/src/iter/adapters/map.cairo` line 25 at r1](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFbq0nJ9xqajlHXNBGI:-OFhH_tQBkjEmpV2MdK7:b4mx4o2) ([raw file](https://github.com/starkware-libs/cairo/blob/f64ccf6bbf3e643c06f55ef392632d771ec975ea/corelib/src/iter/adapters/map.cairo#L25)):*

When calling `map(self.f)`, self.f is of type `FnOnce` so it is moved:
```cairo
    fn map<U, F, +Destruct<F>, +core::ops::FnOnce<F, (T,)>[Output: U]>(
        self: Option<T>, f: F,
    ) -> Option<U>;
```
Maybe we could use `Fn` instead of `FnOnce` to not have to copy one new instance of f at every iteration
</blockquote></details>

oh - you should definitely use `Fn` and not `FnOnce` in the iterator map case.
___
*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFhHvZo95O-Pb103Wsr:-OFhHvZo95O-Pb103Wss:b-2p48eo) (no related file):*
@enitrat for doc 2nd eye

enitrat (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/6949)* status: 1 of 7 files reviewed, 13 unresolved discussions (waiting on @julio4 and @orizi)

___
*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFhHvZo95O-Pb103Wsr:-OFhWvHC0ALLPJhoRsa7:b6vu3bq) (no related file):*

@enitrat for doc 2nd eye
</blockquote></details>

"Adapters" are  "Functions which take an Iterator and return another Iterator". Thus `Map` is not an adapter as it returns
___
*[`corelib/src/iter.cairo` line 25 at r2](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFhRtPw3J-gp5lcYkgi:-OFhRtPw3J-gp5lcYkgj:b-p7j7he) ([raw file](https://github.com/starkware-libs/cairo/blob/a08c28f787b2e1441389530ab8c0307f6108c28b/corelib/src/iter.cairo#L25)):*
> ```Cairo
> //! [Traits]: #traits
> //! [Functions]: #functions
> //! [Structs]: #structs
> ```

we'll figure this out better once scarb doc updates but I think this might need to refer the full path to the modules
___
*[`corelib/src/iter.cairo` line 57 at r2](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFhSf7dAXuu9B7TpZ4u:-OFhSf7dAXuu9B7TpZ4v:undefined) ([raw file](https://github.com/starkware-libs/cairo/blob/a08c28f787b2e1441389530ab8c0307f6108c28b/corelib/src/iter.cairo#L57)):*
> ```Cairo
> //! [`next`]: Iterator::next
> //!
> //! # Iterators in Cairo
> ```





_Suggestion:_
```Cairo
//! # Forms of iteration
```
___
*[`corelib/src/iter.cairo` line 59 at r2](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFhSjty0_H7mRnOWsqj:-OFhSjty0_H7mRnOWsqk:undefined) ([raw file](https://github.com/starkware-libs/cairo/blob/a08c28f787b2e1441389530ab8c0307f6108c28b/corelib/src/iter.cairo#L59)):*
> ```Cairo
> //! # Iterators in Cairo
> //!
> //! There are currently one common method which can create iterators from a collection:
> ```





_Suggestion:_
```Cairo
//! There is currently only one common method which can create iterators from a collection:
```
___
*[`corelib/src/iter.cairo` line 141 at r2](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFhTmEM5_i1e2JJcxeW:-OFhTmEM5_i1e2JJcxeX:undefined) ([raw file](https://github.com/starkware-libs/cairo/blob/a08c28f787b2e1441389530ab8c0307f6108c28b/corelib/src/iter.cairo#L141)):*
> ```Cairo
> //!
> //! This will print the numbers one through five, each on their own line. But
> //! you'll notice something here: we never called anything on our vector to
> ```





_Suggestion:_
```Cairo
//! you'll notice something here: we never called anything on our array to
```
___
*[`corelib/src/iter.cairo` line 144 at r2](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFhTrKT3bKWOyfVqbC7:-OFhTpBl2RHRLENNIGyT:undefined) ([raw file](https://github.com/starkware-libs/cairo/blob/a08c28f787b2e1441389530ab8c0307f6108c28b/corelib/src/iter.cairo#L144)):*
> ```Cairo
> //! produce an iterator. What gives?
> //!
> //! There's a trait in the standard library for converting something into an
> ```





_Suggestion:_
```Cairo
//! There's a trait in the core library for converting something into an
```
___
*[`corelib/src/iter.cairo` line 150 at r2](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFhTuhy2DKddkHVNsAh:-OFhTuhy2DKddkHVNsAi:bu9hrj9) ([raw file](https://github.com/starkware-libs/cairo/blob/a08c28f787b2e1441389530ab8c0307f6108c28b/corelib/src/iter.cairo#L150)):*
> ```Cairo
> //! it into:
> //!
> //! [`into_iter`]: IntoIterator::into_iter
> ```

might need the full path here
___
*[`corelib/src/iter.cairo` line 179 at r2](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFhTzxF5nWTABFVfpCR:-OFhTzxF5nWTABFVfpCS:b-n3e5o4) ([raw file](https://github.com/starkware-libs/cairo/blob/a08c28f787b2e1441389530ab8c0307f6108c28b/corelib/src/iter.cairo#L179)):*
> ```Cairo
> //!     result
> //! }
> //! ```
> ```

@orizi is this accurate for Cairo as well?



_Code quote:_
```Cairo
//! ```
//! let values = array![1, 2, 3, 4, 5];
//! {
//!     let mut iter = IntoIterator::into_iter(values);
//!     let result = loop {
//!             let mut next = 0;
//!             match iter.next() {
//!                 Option::Some(val) => next = val,
//!                 Option::None => {
//!                     break;
//!                 },
//!             };
//!             let x = next;
//!             let () = { println!("{x}"); };
//!         };
//!     result
//! }
//! ```
```
___
*[`corelib/src/iter.cairo` line 185 at r2](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFhU7fyDW0_fALcsHr7:-OFhU7fyDW0_fALcsHr8:undefined) ([raw file](https://github.com/starkware-libs/cairo/blob/a08c28f787b2e1441389530ab8c0307f6108c28b/corelib/src/iter.cairo#L185)):*
> ```Cairo
> //! that point, we `break` out of the loop, and we're done iterating.
> //!
> //! There's one more subtle bit here: the standard library contains an
> ```





_Suggestion:_
```Cairo
//! There's one more subtle bit here: the core library contains an
```
___
*[`corelib/src/iter.cairo` line 209 at r2](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFhUeAWAb8HTo21N6Kq:-OFhUeAWAb8HTo21N6Kr:bxsynfn) ([raw file](https://github.com/starkware-libs/cairo/blob/a08c28f787b2e1441389530ab8c0307f6108c28b/corelib/src/iter.cairo#L209)):*
> ```Cairo
> //!
> //! If an iterator adapter panics, the iterator will be in an unspecified (but
> //! memory safe) state.
> ```

Not sure this applies



_Code quote:_
```Cairo
//! If an iterator adapter panics, the iterator will be in an unspecified (but
//! memory safe) state.
```
___
*[`corelib/src/option.cairo` line 525 at r2](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFhWS5_0SEJFA0Vk51g:-OFhWS5_0SEJFA0Vk51h:b-1blfvs) ([raw file](https://github.com/starkware-libs/cairo/blob/a08c28f787b2e1441389530ab8c0307f6108c28b/corelib/src/option.cairo#L525)):*
> ```Cairo
>     /// assert!(x.map(|s: ByteArray| s.len()) == Option::None);
>     /// ```
>     fn map<U, F, +Destruct<F>, +core::ops::FnOnce<F, (T,)>[Output: U]>(
> ```

update trait as well
___
*[`corelib/src/option.cairo` line 513 at r2](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFhV1vNDPzVQeuxaP_J:-OFhV1vNDPzVQeuxaP_K:b-z825pk) ([raw file](https://github.com/starkware-libs/cairo/blob/825b42d05399ba8d38d27cf3f30496c4acafa1fe/corelib/src/option.cairo#L513)):*
> ```Cairo
>     /////////////////////////////////////////////////////////////////////////
>     // Transforming contained values
>     /////////////////////////////////////////////////////////////////////////
> ```

restore this header



_Code quote:_
```Cairo
    /////////////////////////////////////////////////////////////////////////
    // Transforming contained values
    /////////////////////////////////////////////////////////////////////////
```

enitrat (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/6949)* status: 1 of 7 files reviewed, 13 unresolved discussions (waiting on @julio4 and @orizi)

___
*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFhHvZo95O-Pb103Wsr:-OFhYfbJD7g1hHG6jgMp:bad4xp6) (no related file):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

"Adapters" are  "Functions which take an Iterator and return another Iterator". Thus `Map` is not an adapter as it returns
</blockquote></details>

disregard this comment, I didn't finish to type it, it's fine

julio4:
*[`corelib/src/iter.cairo` line 25 at r2](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFhRtPw3J-gp5lcYkgi:-OFkytg98eM-FgPzAVjI:b-7jlbed) ([raw file](https://github.com/starkware-libs/cairo/blob/a08c28f787b2e1441389530ab8c0307f6108c28b/corelib/src/iter.cairo#L25)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

we'll figure this out better once scarb doc updates but I think this might need to refer the full path to the modules
</blockquote></details>

What format should I use for now?

julio4 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/6949)* status: 1 of 7 files reviewed, 13 unresolved discussions (waiting on @enitrat and @orizi)

___
*[`corelib/src/iter.cairo` line 150 at r2](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFhTuhy2DKddkHVNsAh:-OFkzsJPBpLv0Dfx5trc:b-896fix) ([raw file](https://github.com/starkware-libs/cairo/blob/a08c28f787b2e1441389530ab8c0307f6108c28b/corelib/src/iter.cairo#L150)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

might need the full path here
</blockquote></details>

Done.
___
*[`corelib/src/iter.cairo` line 179 at r2](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFhTzxF5nWTABFVfpCR:-OFkzwv1Cf47lIJG2Jq1:bao6hr4) ([raw file](https://github.com/starkware-libs/cairo/blob/a08c28f787b2e1441389530ab8c0307f6108c28b/corelib/src/iter.cairo#L179)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

@orizi is this accurate for Cairo as well?
</blockquote></details>

I did a quick translation of the rust documentation and have no idea how it's done in Cairo. We can remove this part maybe.
___
*[`corelib/src/iter.cairo` line 209 at r2](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFhUeAWAb8HTo21N6Kq:-OFl-EJN3i5xYaPJ4hR4:b-f4wmlk) ([raw file](https://github.com/starkware-libs/cairo/blob/a08c28f787b2e1441389530ab8c0307f6108c28b/corelib/src/iter.cairo#L209)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

Not sure this applies
</blockquote></details>

Removed as I'm not sure as well
___
*[`corelib/src/option.cairo` line 513 at r2](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFhV1vNDPzVQeuxaP_J:-OFl-VOG8X7YOt0y_K6p:b-lgficv) ([raw file](https://github.com/starkware-libs/cairo/blob/825b42d05399ba8d38d27cf3f30496c4acafa1fe/corelib/src/option.cairo#L513)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

restore this header
</blockquote></details>

Done.
I don't like the way it's show up in the documentation hover by the language server so I removed it and forgot to revert.
It's more a lsp issue

___
*[`corelib/src/option.cairo` line 525 at r2](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFhWS5_0SEJFA0Vk51g:-OFl-R5r7XQuV0xLu9PF:b-896fix) ([raw file](https://github.com/starkware-libs/cairo/blob/a08c28f787b2e1441389530ab8c0307f6108c28b/corelib/src/option.cairo#L525)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

update trait as well
</blockquote></details>

Done.
___
*[`corelib/src/iter.cairo` line 57 at r2](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFhSf7dAXuu9B7TpZ4u:-OFkzXPTBnaEcGuFYawS:b-896fix) ([raw file](https://github.com/starkware-libs/cairo/blob/a08c28f787b2e1441389530ab8c0307f6108c28b/corelib/src/iter.cairo#L57)):*
> ```Cairo
> //! [`next`]: Iterator::next
> //!
> //! # Iterators in Cairo
> ```

Done.
___
*[`corelib/src/iter.cairo` line 59 at r2](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFhSjty0_H7mRnOWsqj:-OFkzXld8-IzgJain_MN:b-896fix) ([raw file](https://github.com/starkware-libs/cairo/blob/a08c28f787b2e1441389530ab8c0307f6108c28b/corelib/src/iter.cairo#L59)):*
> ```Cairo
> //! # Iterators in Cairo
> //!
> //! There are currently one common method which can create iterators from a collection:
> ```

Done.
___
*[`corelib/src/iter.cairo` line 141 at r2](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFhTmEM5_i1e2JJcxeW:-OFkzbB936rQHbrGqrOT:b-896fix) ([raw file](https://github.com/starkware-libs/cairo/blob/a08c28f787b2e1441389530ab8c0307f6108c28b/corelib/src/iter.cairo#L141)):*
> ```Cairo
> //!
> //! This will print the numbers one through five, each on their own line. But
> //! you'll notice something here: we never called anything on our vector to
> ```

Done.
___
*[`corelib/src/iter.cairo` line 144 at r2](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFhTrKT3bKWOyfVqbC7:-OFkzdrB3TK6OUQ6djtL:b-896fix) ([raw file](https://github.com/starkware-libs/cairo/blob/a08c28f787b2e1441389530ab8c0307f6108c28b/corelib/src/iter.cairo#L144)):*
> ```Cairo
> //! produce an iterator. What gives?
> //!
> //! There's a trait in the standard library for converting something into an
> ```

Done.
___
*[`corelib/src/iter.cairo` line 185 at r2](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFhU7fyDW0_fALcsHr7:-OFl-9ck0PMCzX9uMeLr:b-896fix) ([raw file](https://github.com/starkware-libs/cairo/blob/a08c28f787b2e1441389530ab8c0307f6108c28b/corelib/src/iter.cairo#L185)):*
> ```Cairo
> //! that point, we `break` out of the loop, and we're done iterating.
> //!
> //! There's one more subtle bit here: the standard library contains an
> ```

Done.

orizi (commented):
FnMut will be possible in the future, but won't actually affect the efficiency, as Cairo has an immutable memory model, so all mutability is simulates, so even given FnMut, it would not be "reference based" as there are no references.

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/6949#-:-OFl9HaW77bYRvAAlSc1:b-ycajr5)* status: 1 of 7 files reviewed, 13 unresolved discussions (waiting on @enitrat and @julio4)

___
*[`corelib/src/iter.cairo` line 179 at r2](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFhTzxF5nWTABFVfpCR:-OFlA5vUASlb-3j5hSpL:ba6mv6m) ([raw file](https://github.com/starkware-libs/cairo/blob/a08c28f787b2e1441389530ab8c0307f6108c28b/corelib/src/iter.cairo#L179)):*

I did a quick translation of the rust documentation and have no idea how it's done in Cairo. We can remove this part maybe.
</blockquote></details>

It is correct though.
while, while let, and for are the same here as in rust.

julio4:
*[`corelib/src/iter/adapters/map.cairo` line 25 at r1](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFbq0nJ9xqajlHXNBGI:-OFl8Zp11QVgihqrXkQz:bh7egoj) ([raw file](https://github.com/starkware-libs/cairo/blob/f64ccf6bbf3e643c06f55ef392632d771ec975ea/corelib/src/iter/adapters/map.cairo#L25)):*

oh - you should definitely use `Fn` and not `FnOnce` in the iterator map case.
</blockquote></details>

Yes figured that out and updated the implementation. 

One thing I wasn't able to figure out is how to use an `Fn` as an `FnOnce`.
`Option::map` takes a `FnOnce`:
```cairo
    fn map<U, F, +Destruct<F>, +core::ops::FnOnce<F, (T,)>[Output: U]>(
        self: Option<T>, f: F,
    ) -> Option<U>;
```

And there's an implementation of `FnOnce` for types that implement `Fn`:
```cairo
/// An implementation of `FnOnce` when `Fn` is implemented.
/// Makes sure we can always pass an `Fn` to a function that expects an `FnOnce`.
impl FnOnceImpl<T, Args, +Destruct<T>, +Fn<T, Args>> of FnOnce<T, Args> {
    type Output = Fn::<T, Args>::Output;
    fn call(self: T, args: Args) -> Self::Output {
        Fn::call(@self, args)
    }
}
```

But this don't works:
```cairo
    fn next(ref self: Map<I, F>) -> Option<func::Output> {
        self.iter.next().map(@self.f)
    }
```

So I propose adding a new `FnOnceSnapImpl`:
```cairo
/// An implementation of `FnOnce` when `Fn` is implemented and the receiver is by-reference.
/// Makes sure we can always pass an `Fn` to a function that expects an `FnOnce`.
impl FnOnceSnapImpl<T, Args, +Destruct<T>, +Fn<T, Args>> of FnOnce<@T, Args> {
    type Output = Fn::<T, Args>::Output;
    fn call(self: @T, args: Args) -> Self::Output {
        Fn::call(self, args)
    }
}
```

If there's a better way to do it let me know.

julio4:
And here's some benchmark:
A:
```cairo
let a = array![1, 2, 3];
let mut iter = a.into_iter().map(|x| 2 * x);
let _ = iter.next();
let _ = iter.next();
let _ = iter.next();
```
B:
```cairo
let a = array![1, 2, 3];
let mut b = array![];
for x in a {
    b.append(2 * x);
};
```

A: gas usage est.: 3000
B: gas usage est.: 9680

So this could be used for optimizing some logic

julio4:
This makes sense, the optimization was mostly not copying each function with `Fn`. Thank you for the clear explanation!

julio4:
*[`corelib/src/iter.cairo` line 179 at r2](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFhTzxF5nWTABFVfpCR:-OFlCS7c-hQZBbB43ZSd:b-o0bu2) ([raw file](https://github.com/starkware-libs/cairo/blob/a08c28f787b2e1441389530ab8c0307f6108c28b/corelib/src/iter.cairo#L179)):*

It is correct though.
while, while let, and for are the same here as in rust.
</blockquote></details>

It is slightly different than the rust one though, where I tried to adapt it to Cairo syntax as closely as possible

orizi (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/6949)* status: 1 of 8 files reviewed, 12 unresolved discussions (waiting on @enitrat)

___
*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFlGGlXDwB_H5a7qkW5:-OFlGGlYD4q-drNnDUeY:b4ymqcq) (no related file):*
Adding @TomerStarkware for 2nd eye, and more insights for the FnOnce vs Fn stuff.

orizi (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/6949)* status: 1 of 8 files reviewed, 12 unresolved discussions (waiting on @enitrat)

___
*[`corelib/src/iter/adapters/map.cairo` line 25 at r1](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFbq0nJ9xqajlHXNBGI:-OFlGwLD851CeGNFdBP1:bi9ttzi) ([raw file](https://github.com/starkware-libs/cairo/blob/f64ccf6bbf3e643c06f55ef392632d771ec975ea/corelib/src/iter/adapters/map.cairo#L25)):*

Yes figured that out and updated the implementation. 

One thing I wasn't able to figure out is how to use an `Fn` as an `FnOnce`.
`Option::map` takes a `FnOnce`:
```cairo
    fn map<U, F, +Destruct<F>, +core::ops::FnOnce<F, (T,)>[Output: U]>(
        self: Option<T>, f: F,
    ) -> Option<U>;
```

And there's an implementation of `FnOnce` for types that implement `Fn`:
```cairo
/// An implementation of `FnOnce` when `Fn` is implemented.
/// Makes sure we can always pass an `Fn` to a function that expects an `FnOnce`.
impl FnOnceImpl<T, Args, +Destruct<T>, +Fn<T, Args>> of FnOnce<T, Args> {
    type Output = Fn::<T, Args>::Output;
    fn call(self: T, args: Args) -> Self::Output {
        Fn::call(@self, args)
    }
}
```

But this don't works:
```cairo
    fn next(ref self: Map<I, F>) -> Option<func::Output> {
        self.iter.next().map(@self.f)
    }
```

So I propose adding a new `FnOnceSnapImpl`:
```cairo
/// An implementation of `FnOnce` when `Fn` is implemented and the receiver is by-reference.
/// Makes sure we can always pass an `Fn` to a function that expects an `FnOnce`.
impl FnOnceSnapImpl<T, Args, +Destruct<T>, +Fn<T, Args>> of FnOnce<@T, Args> {
    type Output = Fn::<T, Args>::Output;
    fn call(self: @T, args: Args) -> Self::Output {
        Fn::call(self, args)
    }
}
```

If there's a better way to do it let me know.
</blockquote></details>

@TomerStarkware

TomerStarkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/6949)* status: 1 of 8 files reviewed, 12 unresolved discussions (waiting on @enitrat and @orizi)

___
*[`corelib/src/iter/adapters/map.cairo` line 25 at r1](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFbq0nJ9xqajlHXNBGI:-OFmpxGU3yUeXxJyylJg:bxmlx2h) ([raw file](https://github.com/starkware-libs/cairo/blob/f64ccf6bbf3e643c06f55ef392632d771ec975ea/corelib/src/iter/adapters/map.cairo#L25)):*

@TomerStarkware
</blockquote></details>

I agree with adding FnOnceSnapImpl, and it looks like Rust is doing that as well, 
it might be better to have it implement Fn<@T,Args> and then the regular FnOnceImpl will work

enitrat (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/6949)* status: 1 of 8 files reviewed, 3 unresolved discussions (waiting on @julio4, @orizi, and @TomerStarkware)

___
*[`corelib/src/iter.cairo` line 25 at r2](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFhRtPw3J-gp5lcYkgi:-OFnGAws9irmdnR4fJB3:b-g7cmva) ([raw file](https://github.com/starkware-libs/cairo/blob/a08c28f787b2e1441389530ab8c0307f6108c28b/corelib/src/iter.cairo#L25)):*

What format should I use for now?
</blockquote></details>

let's keep this for now
___
*[`corelib/src/option.cairo` line 513 at r2](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFhV1vNDPzVQeuxaP_J:-OFnGuIY1Q6-7j4TrQdr:b-1e5zb4) ([raw file](https://github.com/starkware-libs/cairo/blob/825b42d05399ba8d38d27cf3f30496c4acafa1fe/corelib/src/option.cairo#L513)):*

Done.
I don't like the way it's show up in the documentation hover by the language server so I removed it and forgot to revert.
It's more a lsp issue

</blockquote></details>

can you open a bug report?
___
*[`corelib/src/iter/traits/iterator.cairo` line 79 at r2](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFhZ5K0CPmZOrpkfA0k:-OFhZ5K0CPmZOrpkfA0l:b-fcc1yq) ([raw file](https://github.com/starkware-libs/cairo/blob/a08c28f787b2e1441389530ab8c0307f6108c28b/corelib/src/iter/traits/iterator.cairo#L79)):*
> ```Cairo
>     ///
>     /// ```
>     /// # #![allow(unused_must_use)]
> ```

does this tag exist?

julio4 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/6949)* status: 1 of 8 files reviewed, 3 unresolved discussions (waiting on @enitrat, @orizi, and @TomerStarkware)

___
*[`corelib/src/option.cairo` line 513 at r2](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFhV1vNDPzVQeuxaP_J:-OFnkmQt5yhAGUGn_5c4:b-pckfov) ([raw file](https://github.com/starkware-libs/cairo/blob/825b42d05399ba8d38d27cf3f30496c4acafa1fe/corelib/src/option.cairo#L513)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

can you open a bug report?
</blockquote></details>

[Yes I already opened it](https://github.com/software-mansion/cairols/issues/134)
___
*[`corelib/src/iter/traits/iterator.cairo` line 79 at r2](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFhZ5K0CPmZOrpkfA0k:-OFnkckl648T17FiKOdc:b-5k2wj8) ([raw file](https://github.com/starkware-libs/cairo/blob/a08c28f787b2e1441389530ab8c0307f6108c28b/corelib/src/iter/traits/iterator.cairo#L79)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

does this tag exist?
</blockquote></details>

No it don't exist

julio4:
*[`corelib/src/iter/adapters/map.cairo` line 25 at r1](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFbq0nJ9xqajlHXNBGI:-OFnkw8i-o5hX0J4ButQ:by2s74f) ([raw file](https://github.com/starkware-libs/cairo/blob/f64ccf6bbf3e643c06f55ef392632d771ec975ea/corelib/src/iter/adapters/map.cairo#L25)):*
<details><summary><i>Previously, TomerStarkware wrote…</i></summary><blockquote>

I agree with adding FnOnceSnapImpl, and it looks like Rust is doing that as well, 
it might be better to have it implement Fn<@T,Args> and then the regular FnOnceImpl will work
</blockquote></details>

@TomerStarkware I added it in another PR: https://github.com/starkware-libs/cairo/pull/6992

orizi (changes requested):
> Reviewed 3 of 3 files at r3, 1 of 2 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/6949)* status: 6 of 8 files reviewed, 4 unresolved discussions (waiting on @enitrat, @julio4, and @TomerStarkware)

___
*[`corelib/src/iter/traits/iterator.cairo` line 79 at r2](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFhZ5K0CPmZOrpkfA0k:-OFp9p0g81GXqPOF5rUN:b-jh7tj1) ([raw file](https://github.com/starkware-libs/cairo/blob/a08c28f787b2e1441389530ab8c0307f6108c28b/corelib/src/iter/traits/iterator.cairo#L79)):*

No it don't exist
</blockquote></details>

do a `let _ = ...` instead - valid code, and explains the same effect.
___
*[`corelib/src/iter/adapters/map.cairo` line 23 at r5](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFpAUjC1TUouoW1gZF_:-OFpAUjC1TUouoW1gZFa:b-dyqrqm) ([raw file](https://github.com/starkware-libs/cairo/blob/70b04ac1d54b2e693b3ca688eef90d6f044ab1fe/corelib/src/iter/adapters/map.cairo#L23)):*
> ```Cairo
>         Map { iter, f }
>     }
> }
> ```

consider - just to avoid the confusion with the starknet storage's MapTrait,
or just having a non-member function - as you don't even use `self` in it.



_Suggestion:_
```Cairo
#[must_use]
#[derive(Drop, Clone)]
pub struct Map<I, F> {
    pub(crate) iter: I,
    pub(crate) f: F,
}
//// or:
pub(crate) fn mapped_iterator<I, F>(iter: I, f: F) -> Map<I, F> {
    Map { iter, f }
}
```

julio4 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/6949)* status: 6 of 8 files reviewed, 4 unresolved discussions (waiting on @enitrat, @orizi, and @TomerStarkware)

___
*[`corelib/src/iter/adapters/map.cairo` line 23 at r5](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFpAUjC1TUouoW1gZF_:-OFqvnj83tW0ApSsgMjV:b-896fix) ([raw file](https://github.com/starkware-libs/cairo/blob/70b04ac1d54b2e693b3ca688eef90d6f044ab1fe/corelib/src/iter/adapters/map.cairo#L23)):*

consider - just to avoid the confusion with the starknet storage's MapTrait,
or just having a non-member function - as you don't even use `self` in it.
</blockquote></details>

Done.
___
*[`corelib/src/iter/traits/iterator.cairo` line 79 at r2](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFhZ5K0CPmZOrpkfA0k:-OFqud_5DNGixCz5IMJN:b-896fix) ([raw file](https://github.com/starkware-libs/cairo/blob/a08c28f787b2e1441389530ab8c0307f6108c28b/corelib/src/iter/traits/iterator.cairo#L79)):*

do a `let _ = ...` instead - valid code, and explains the same effect.
</blockquote></details>

Done.

enitrat (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/6949#-:-OFrC55dCt-SQ0-b23vW:bnfp4nl)* status: 3 of 8 files reviewed, 2 unresolved discussions (waiting on @orizi and @TomerStarkware)

orizi (changes requested):
> Reviewed 1 of 4 files at r6.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/6949)* status: 4 of 8 files reviewed, 5 unresolved discussions (waiting on @enitrat, @julio4, and @TomerStarkware)

___
*[`corelib/src/iter/adapters/map.cairo` line 20 at r6](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFrBwbX8Ln_S0mH8_G-:-OFrBwbX8Ln_S0mH8_G0:b-gpf8ps) ([raw file](https://github.com/starkware-libs/cairo/blob/46f1928a45fb6f7d582c905ce11e0b49a98e36f5/corelib/src/iter/adapters/map.cairo#L20)):*
> ```Cairo
> pub(crate) fn mapped_iterator<I, F>(iter: I, f: F) -> Map<I, F> {
>     Map { iter, f }
> }
> ```

only one `pub(crate)` in the path is relevant.

this is contained in a private module.



_Suggestion:_
```Cairo
pub fn mapped_iterator<I, F>(iter: I, f: F) -> Map<I, F> {
    Map { iter, f }
}
```
___
*[`corelib/src/test/iter_test.cairo` line 1 at r6](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFrB4Vq1Npj2Li61o_C:-OFrB4Vq1Npj2Li61o_D:b-t37hfo) ([raw file](https://github.com/starkware-libs/cairo/blob/46f1928a45fb6f7d582c905ce11e0b49a98e36f5/corelib/src/test/iter_test.cairo#L1)):*
> ```Cairo
> use crate::iter::{IntoIterator, Iterator};
> ```

rebase - in prelude now.
___
*[`corelib/src/iter/adapters.cairo` line 2 at r6](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFrBgkjA836107pMzh6:-OFrBgkjA836107pMzh7:b-u33mpt) ([raw file](https://github.com/starkware-libs/cairo/blob/46f1928a45fb6f7d582c905ce11e0b49a98e36f5/corelib/src/iter/adapters.cairo#L2)):*
> ```Cairo
> mod map;
> pub use map::{Map, mapped_iterator};
> ```

this just made `mapped_iterator` pub.



_Suggestion:_
```Cairo
pub use map::Map;
pub(crate) use map::mapped_iterator;
```
___
*[`corelib/src/iter/traits/iterator.cairo` line 82 at r6](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFrBNE8CPAOZVld3RGf:-OFrBNE8CPAOZVld3RGg:b-479f1i) ([raw file](https://github.com/starkware-libs/cairo/blob/46f1928a45fb6f7d582c905ce11e0b49a98e36f5/corelib/src/iter/traits/iterator.cairo#L82)):*
> ```Cairo
>     /// let _ = (0..5_usize).into_iter().map(|x| println!("{x}"));
>     ///
>     /// // it won't even execute, as it is lazy. Cairo will warn you about this.
> ```

or something similar (as it doesn't in this case, due to the `let _ = `



_Suggestion:_
```Cairo
    /// // it won't even execute, as it is lazy. Cairo will warn you about this if not specifically ignored, as is done here..
```

enitrat (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/6949)* status: 4 of 8 files reviewed, 5 unresolved discussions (waiting on @julio4, @orizi, and @TomerStarkware)

___
*[`corelib/src/iter/adapters/map.cairo` line 20 at r6](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFrBwbX8Ln_S0mH8_G-:-OFrGZwl2RN56GRr4VmM:bxc63um) ([raw file](https://github.com/starkware-libs/cairo/blob/46f1928a45fb6f7d582c905ce11e0b49a98e36f5/corelib/src/iter/adapters/map.cairo#L20)):*

only one `pub(crate)` in the path is relevant.

this is contained in a private module.
</blockquote></details>

why are we marking this as public when we don't have the intent of making this accessible (as the module is private)?

julio4 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/6949)* status: 3 of 8 files reviewed, 5 unresolved discussions (waiting on @enitrat, @orizi, and @TomerStarkware)

___
*[`corelib/src/iter/adapters.cairo` line 2 at r6](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFrBgkjA836107pMzh6:-OFrNlpbB8gn9Zd7n3G2:blawlho) ([raw file](https://github.com/starkware-libs/cairo/blob/57d331902f5b8958562af0d7b9f22304ef0d98bd/corelib/src/iter/adapters.cairo#L2)):*

this just made `mapped_iterator` pub.
</blockquote></details>

Done.
I just don't know why I had warning unused import: `core::iter::adapters::mapped_iterator` but it was used it Iterator
___
*[`corelib/src/iter/adapters/map.cairo` line 20 at r6](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFrBwbX8Ln_S0mH8_G-:-OFrNOmK0q-2w3PLPCPH:bulh30y) ([raw file](https://github.com/starkware-libs/cairo/blob/57d331902f5b8958562af0d7b9f22304ef0d98bd/corelib/src/iter/adapters/map.cairo#L20)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

why are we marking this as public when we don't have the intent of making this accessible (as the module is private)?
</blockquote></details>

We make it public in the scope of the map module, so that in adapters mod we can use it and make it public for the crate?
___
*[`corelib/src/iter/traits/iterator.cairo` line 82 at r6](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFrBNE8CPAOZVld3RGf:-OFrM8Ck2eipeWtboFhd:b-896fix) ([raw file](https://github.com/starkware-libs/cairo/blob/57d331902f5b8958562af0d7b9f22304ef0d98bd/corelib/src/iter/traits/iterator.cairo#L82)):*

or something similar (as it doesn't in this case, due to the `let _ = `
</blockquote></details>

Done.
___
*[`corelib/src/test/iter_test.cairo` line 1 at r6](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFrB4Vq1Npj2Li61o_C:-OFrNNqe2bUJ6cjc8oe_:b-896fix) ([raw file](https://github.com/starkware-libs/cairo/blob/57d331902f5b8958562af0d7b9f22304ef0d98bd/corelib/src/test/iter_test.cairo#L1)):*

rebase - in prelude now.
</blockquote></details>

Done.

orizi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/6949)* status: 3 of 8 files reviewed, 4 unresolved discussions (waiting on @enitrat, @julio4, and @TomerStarkware)

___
*[`corelib/src/iter/adapters/map.cairo` line 20 at r6](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFrBwbX8Ln_S0mH8_G-:-OFrxGCi5HKsMNqeXTHS:b-2y03dl) ([raw file](https://github.com/starkware-libs/cairo/blob/57d331902f5b8958562af0d7b9f22304ef0d98bd/corelib/src/iter/adapters/map.cairo#L20)):*

We make it public in the scope of the map module, so that in adapters mod we can use it and make it public for the crate?
</blockquote></details>

exactly.

julio4 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/6949)* status: 3 of 8 files reviewed, 2 unresolved discussions (waiting on @enitrat, @orizi, and @TomerStarkware)

___
*[`corelib/src/iter/adapters/map.cairo` line 20 at r6](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFrBwbX8Ln_S0mH8_G-:-OFs9TIEDRx_wHI6ww60:b-896fix) ([raw file](https://github.com/starkware-libs/cairo/blob/57d331902f5b8958562af0d7b9f22304ef0d98bd/corelib/src/iter/adapters/map.cairo#L20)):*

exactly.
</blockquote></details>

Done.

orizi (changes requested):
> Reviewed 1 of 2 files at r4, 4 of 4 files at r8.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/6949)* status: all files reviewed, 2 unresolved discussions (waiting on @julio4)

___
*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFlBa1h0lVl-vBzzJl4:-OFuhysrD_G2w3CyGggk:bxsvv40) (no related file):*

And here's some benchmark:
A:
```cairo
let a = array![1, 2, 3];
let mut iter = a.into_iter().map(|x| 2 * x);
let _ = iter.next();
let _ = iter.next();
let _ = iter.next();
```
B:
```cairo
let a = array![1, 2, 3];
let mut b = array![];
for x in a {
    b.append(2 * x);
};
```

A: gas usage est.: 3000
B: gas usage est.: 9680

So this could be used for optimizing some logic
</blockquote></details>

Can you please rebase and check again?
Also, note that loops has their own overhead in general.
___
*[`corelib/src/ops/function.cairo` line 27 at r8](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFuhcrk4YK432bA9UwJ:-OFuhcrk4YK432bA9UwK:b-6xjmqb) ([raw file](https://github.com/starkware-libs/cairo/blob/fbe32680caedf95762c1857a810c31730081f5b5/corelib/src/ops/function.cairo#L27)):*
> ```Cairo
> 
> /// An implementation of `FnOnce` when `Fn` is implemented and the receiver is by-reference.
> /// Makes sure we can always pass an `Fn` to a function that expects an `FnOnce`.
> ```

rebase please.

julio4 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/6949)* status: 6 of 8 files reviewed, 2 unresolved discussions (waiting on @orizi)

___
*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFlBa1h0lVl-vBzzJl4:-OFv12ad3gPzz-H14vLN:bs0nbkx) (no related file):*

Can you please rebase and check again?
Also, note that loops has their own overhead in general.
</blockquote></details>

Yes, I considered loops because in most case I think users will use for loops.
Here's maybe a more accurate benchmark:

A
```cairo
let a: Array<u8> = array![1, 2, 3];
let mut iter = a.into_iter().map(|x| 2 * x);
loop {
    if let Option::Some(x) = iter.next() {
        x;
    } else {
        break;
    }
};
```

B
```cairo
let a: Array<u8> = array![1, 2, 3];
let mut iter = a.into_iter();
loop {
    if let Option::Some(x) = iter.next() {
        x * 2;
    } else {
        break;
    }
};
```

C
```cairo
let a: Array<u8> = array![1, 2, 3];
for x in a {
    2 * x;
};
```

A: gas usage est.: 12090  
B: gas usage est.: 10290
C: gas usage est.: 9990

There's a overhead with map, and it seems to be linear with the size of the iterator.
Using a RangeIterator of size 30 gives:

A: gas usage est.: 153620  
B: gas usage est.: 80870
C: gas usage est.: 83870

However the strengths of iterators is the ability to define mutation in a declarative way, and using composition/chaining with multiple adapters (filter, take, ...), so it's still worthwhile to try to optimize in my opinion.

Let's take a last example where we need to perform two distinct mutation over an array.
With imperative approach, we have to use an intermediate array (supposing that we're not minimizing the two mutations in only one): 

A
```cairo
let mut iter = (0..100_u8)
    .into_iter()
    .map(|x| {
        if (x % 2) == 0 {
            x * 2
        } else {
            x + 1
        }
    })
    .map(|x| x + 1);

let mut collect = array![];
for x in iter {
    collect.append(x);
}
```

B
```cairo
let mut iter = (0..100_u8).into_iter();

let mut intermediate_arr = array![];
for x in iter {
    if (x % 2) == 0 {
        intermediate_arr.append(x * 2);
    } else {
        intermediate_arr.append(x + 1);
    }
};

let mut collect = array![];
for x in intermediate_arr {
    collect.append(x + 1);
}
```

A: gas usage est.: 748030
B: gas usage est.: 747140

Given the map iterator overhead, it is same performance as creating a new intermediate array.
Do you think there could be way to optimize iterator performances?
___
*[`corelib/src/ops/function.cairo` line 27 at r8](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFuhcrk4YK432bA9UwJ:-OFuyDG70bh9_vZuPmdh:b-896fix) ([raw file](https://github.com/starkware-libs/cairo/blob/fbe32680caedf95762c1857a810c31730081f5b5/corelib/src/ops/function.cairo#L27)):*

rebase please.
</blockquote></details>

Done.

orizi (changes requested):
> Reviewed 2 of 2 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/6949)* status: all files reviewed, 2 unresolved discussions (waiting on @julio4)

___
*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFlBa1h0lVl-vBzzJl4:-OFvKzGsBi3OX4XnOxF2:btdzqr7) (no related file):*

Yes, I considered loops because in most case I think users will use for loops.
Here's maybe a more accurate benchmark:

A
```cairo
let a: Array<u8> = array![1, 2, 3];
let mut iter = a.into_iter().map(|x| 2 * x);
loop {
    if let Option::Some(x) = iter.next() {
        x;
    } else {
        break;
    }
};
```

B
```cairo
let a: Array<u8> = array![1, 2, 3];
let mut iter = a.into_iter();
loop {
    if let Option::Some(x) = iter.next() {
        x * 2;
    } else {
        break;
    }
};
```

C
```cairo
let a: Array<u8> = array![1, 2, 3];
for x in a {
    2 * x;
};
```

A: gas usage est.: 12090  
B: gas usage est.: 10290
C: gas usage est.: 9990

There's a overhead with map, and it seems to be linear with the size of the iterator.
Using a RangeIterator of size 30 gives:

A: gas usage est.: 153620  
B: gas usage est.: 80870
C: gas usage est.: 83870

However the strengths of iterators is the ability to define mutation in a declarative way, and using composition/chaining with multiple adapters (filter, take, ...), so it's still worthwhile to try to optimize in my opinion.

Let's take a last example where we need to perform two distinct mutation over an array.
With imperative approach, we have to use an intermediate array (supposing that we're not minimizing the two mutations in only one): 

A
```cairo
let mut iter = (0..100_u8)
    .into_iter()
    .map(|x| {
        if (x % 2) == 0 {
            x * 2
        } else {
            x + 1
        }
    })
    .map(|x| x + 1);

let mut collect = array![];
for x in iter {
    collect.append(x);
}
```

B
```cairo
let mut iter = (0..100_u8).into_iter();

let mut intermediate_arr = array![];
for x in iter {
    if (x % 2) == 0 {
        intermediate_arr.append(x * 2);
    } else {
        intermediate_arr.append(x + 1);
    }
};

let mut collect = array![];
for x in intermediate_arr {
    collect.append(x + 1);
}
```

A: gas usage est.: 748030
B: gas usage est.: 747140

Given the map iterator overhead, it is same performance as creating a new intermediate array.
Do you think there could be way to optimize iterator performances?
</blockquote></details>

can you compare instead of `B` with:
```
let mut iter = (0..100_u8).into_iter();

let mut intermediate_arr = array![];
for x in iter {
    intermediate_arr.append(if (x % 2) == 0 {
        x * 2
    } else {
        x + 1
    });
};

let mut collect = array![];
for x in intermediate_arr {
    collect.append(x + 1);
}
```
(a bit more equivalent)
___
*[`corelib/src/ops/function.cairo` line 26 at r9](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFvJvg1-GIOJd4NR5yz:-OFvJvg292Ej7HS7tLeK:bmrxaui) ([raw file](https://github.com/starkware-libs/cairo/blob/0363ecd533c8dee82265a7a4eb1adcdb74305d31/corelib/src/ops/function.cairo#L26)):*
> ```Cairo
> }
> 
> /// An implementation of `FnOnce` when `Fn` is implemented and the receiver is by-reference.
> ```

i don't know what do you mean "by-reference" in cairo - but in any case - why is this impl required now?

julio4 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/6949)* status: 7 of 8 files reviewed, 2 unresolved discussions (waiting on @orizi)

___
*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFlBa1h0lVl-vBzzJl4:-OFvykroERVGtM8Kxrrd:blnq82f) (no related file):*

can you compare instead of `B` with:
```
let mut iter = (0..100_u8).into_iter();

let mut intermediate_arr = array![];
for x in iter {
    intermediate_arr.append(if (x % 2) == 0 {
        x * 2
    } else {
        x + 1
    });
};

let mut collect = array![];
for x in intermediate_arr {
    collect.append(x + 1);
}
```
(a bit more equivalent)
</blockquote></details>

True, this gives:
B: gas usage est.: 737140
___
*[`corelib/src/ops/function.cairo` line 26 at r9](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFvJvg1-GIOJd4NR5yz:-OFvyEBS61qLg1OWAATF:bkhks4s) ([raw file](https://github.com/starkware-libs/cairo/blob/0363ecd533c8dee82265a7a4eb1adcdb74305d31/corelib/src/ops/function.cairo#L26)):*

i don't know what do you mean "by-reference" in cairo - but in any case - why is this impl required now?
</blockquote></details>

It was actually by-snapshot and forgot to revert after rebase with #6992
Removed

orizi (approved):
:lgtm:

> Reviewed 1 of 1 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/6949#-:-OFw3SDs4qdPYySTAnPM:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @julio4)

___
*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFlBa1h0lVl-vBzzJl4:-OFw2PE9CRbT3d774bfg:b-ld6gc5) (no related file):*

True, this gives:
B: gas usage est.: 737140
</blockquote></details>

interesting.
in any case - i don't see any reason not to approve the PR - we should probably find the cause for the degradation, and improve it, (in cairo code, or by an additional new optimization stage) but it is not that extreme just yet.
___
*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFlGGlXDwB_H5a7qkW5:-OFw3WPVEr_CHREHUt3G:bb7mm0n) (no related file):*

Adding @TomerStarkware for 2nd eye, and more insights for the FnOnce vs Fn stuff.
</blockquote></details>

ping

julio4 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/6949)* status: all files reviewed, 1 unresolved discussion (waiting on @orizi)

___
*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFlBa1h0lVl-vBzzJl4:-OFwBBE752NFWKiNyQW_:b-lxk58o) (no related file):*

interesting.
in any case - i don't see any reason not to approve the PR - we should probably find the cause for the degradation, and improve it, (in cairo code, or by an additional new optimization stage) but it is not that extreme just yet.
</blockquote></details>

Any recommendations to correctly profile cairo code and try to find optimizations? Or it would be more at the compiler level?
It's not extreme, but worst case seems to be almost 100% overhead

orizi (approved):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/6949)* status: all files reviewed, 1 unresolved discussion (waiting on @julio4)

___
*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFlBa1h0lVl-vBzzJl4:-OFwCf6LDxAFX7cauAJH:b-e2poi6) (no related file):*

Any recommendations to correctly profile cairo code and try to find optimizations? Or it would be more at the compiler level?
It's not extreme, but worst case seems to be almost 100% overhead
</blockquote></details>

in the particular case of the 100% overhead i also believe that the `Map` implementation isn't being dropped - the issue there was mostly that in the good case there was no need to actually pass a value around (as it basically just checked if x*2 had an overflow).

so i'll drop the performance issue for now - probably heavier inlining of closures would do a lot of the work.

orizi (changes requested):
> Reviewed 3 of 3 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/6949)* status: all files reviewed, 2 unresolved discussions (waiting on @julio4)

___
*[`corelib/src/iter.cairo` line 234 at r11](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFz73Ag19V5GNyiro2L:-OFz73Ah3T5FfnnCPYdx:b-ay9n31) ([raw file](https://github.com/starkware-libs/cairo/blob/9b15a82b26cb35d71b0a767bf0a4daf5536d842e/corelib/src/iter.cairo#L234)):*
> ```Cairo
> mod adapters;
> mod traits;
> pub use traits::{IntoIterator, Iterator};
> ```

add endline.



_Code quote:_
```Cairo
pub use traits::{IntoIterator, Iterator};
```

julio4 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/6949)* status: 7 of 8 files reviewed, 2 unresolved discussions (waiting on @orizi)

___
*[`corelib/src/iter.cairo` line 234 at r11](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OFz73Ag19V5GNyiro2L:-OG-CERCDqFoa1pwZTWR:b-896fix) ([raw file](https://github.com/starkware-libs/cairo/blob/9b15a82b26cb35d71b0a767bf0a4daf5536d842e/corelib/src/iter.cairo#L234)):*

add endline.
</blockquote></details>

Done.

orizi (approved):
> Reviewed 1 of 1 files at r12, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/6949)* status: all files reviewed, 1 unresolved discussion (waiting on @julio4)

julio4 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/6949)* status: 7 of 8 files reviewed, 1 unresolved discussion (waiting on @orizi and @TomerStarkware)

___
*[`corelib/src/iter/traits/iterator.cairo` line 98 at r12](https://reviewable.io/reviews/starkware-libs/cairo/6949#-OGA7tfL04vZe_7OVRAx:-OGANzkq1xpVmVFEIjkV:b-896fix) ([raw file](https://github.com/starkware-libs/cairo/blob/2fd601f0381201cb27ee0b198a000f10d29e205a/corelib/src/iter/traits/iterator.cairo#L98)):*
<details><summary><i>Previously, TomerStarkware wrote…</i></summary><blockquote>

can we use Fn instead of  Fnonce + Copy?
</blockquote></details>

Done.

orizi (approved):
:lgtm:

> Reviewed 1 of 1 files at r13, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/6949#-:-OGB2n2q9xVmbodgz8Jk:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @TomerStarkware)

TomerStarkware (approved):
> Reviewed 1 of 1 files at r14, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/6949)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @julio4)

PR #6972, opened 2025-01-04
feat: add support for the `..=` RangeInclusive syntax

cairoIover:
Adds support for the `DotDotEq` token in the parser, link `BinaryOperator::DotDotEq(_)` to `RangeInclusiveOp` and implement it in the core library.

PR #7006, opened 2025-01-05
feat(corelib): add RangeBounds and Bound

cairoIover:
With the addition of new range operators and range types (WIP in https://github.com/starkware-libs/cairo/pull/6972) , we are going to have different kinds of bounds for ranges.

Implements the `Bound` enum (representing all 3 kinds of possible bounds) and the `RangeBounds` associated trait to get the bounds of a range. 

Later this will be useful to abstract some inner impls (like whether a range `contains` a number or not) an use a single common implementation.

I left out some comments regarding internal bugs to discuss.

### DIFF
--- a/corelib/src/ops.cairo
+++ b/corelib/src/ops.cairo
@@ -11,7 +11,7 @@ mod range;
 // `RangeOp` is used internally by the compiler.
 #[allow(unused_imports)]
 use range::RangeOp;
-pub use range::{Range, RangeIterator, RangeTrait};
+pub use range::{Bound, Range, RangeBounds, RangeIterator, RangeTrait};
 
 mod function;
 pub use function::{Fn, FnOnce};
--- a/corelib/src/ops/range.cairo
+++ b/corelib/src/ops/range.cairo
@@ -167,3 +167,99 @@ impl SierraRangeIntoIterator<
         }
     }
 }
+
+/// An endpoint of a range of keys.
+///
+/// # Examples
+///
+/// `Bound`s are range endpoints:
+///
+/// ```
+/// use core::ops::{Bounds, RangeBounds};
+///
+/// assert_eq!((1..12).start_bound(), Bound::Included(@1));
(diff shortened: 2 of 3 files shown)


orizi (changes requested):
if it is useful only for `contains` i rather just have an implementation for `contains`.

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/7006#-:-OFtpXCSDwyh_WVaBXvb:b-9vp29v)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @cairoIover)

___
*[`corelib/src/ops/range.cairo` line 263 at r1](https://reviewable.io/reviews/starkware-libs/cairo/7006#-OFtpKa92syyaRO-kdxI:-OFtpKa92syyaRO-kdxJ:b6zpr8z) ([raw file](https://github.com/starkware-libs/cairo/blob/bea1a2aa438f439e24fa8f55b023e24e1b765628/corelib/src/ops/range.cairo#L263)):*
> ```Cairo
>                 Bound::Excluded(end) => item < end,
>                 Bound::Unbounded => true,
>             })
> ```

implementing the basic implementation is more efficient.



_Suggestion:_
```Cairo
        start <= item && item < end
```

cairoIover (commented):
understandable. I was thinking it would be interesting to make reasoning for `contains` based on bounds types, but we can also go for simplicity and avoid abstractions. I can close 

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/7006#-:-OFuntMrDlwsjFoxTmXk:btpjxbx)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @orizi)

orizi (commented):
until we find further use cases - i think this is preferable.

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/7006#-:-OFupLQm3zOkCiSCtsvP:b-da0mpi)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @cairoIover)

PR #6843, opened 2024-12-06
Migrated the compiler to per-crate plugin queries

integraledelebesgue:

PR #7016, opened 2025-01-07
fix: remove 'assert_eq!' test_mode checks

enitrat:
Following our discussion

### DIFF
--- a/crates/cairo-lang-test-plugin/src/inline_macros/assert.rs
+++ b/crates/cairo-lang-test-plugin/src/inline_macros/assert.rs
@@ -165,15 +165,6 @@ trait CompareAssertionPlugin: NamedPlugin {
             ",
         });
         let (content, code_mappings) = builder.build();
-        let mut diagnostics = vec![];
-        if !metadata.cfg_set.contains(&Cfg::kv("target", "test"))
-            && !metadata.cfg_set.contains(&Cfg::name("test"))
-        {
-            diagnostics.push(PluginDiagnostic::error(
-                syntax,
-                format!("`{}` macro is only available in test mode.", Self::NAME),
-            ));
-        }
         InlinePluginResult {
             code: Some(PluginGeneratedFile {
                 name: format!("{}_macro", Self::NAME).into(),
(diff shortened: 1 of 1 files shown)


orizi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/7016)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @enitrat)

___
*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/7016#-OG-CSGhCE4NKlhjGjMK:-OG-CSGhCE4NKlhjGjML:b-6mnjjr) (no related file):*
tagging @mkaput  and @maciektr  to validate that we are indeed only adding this plugin in tests - unless very specifically added.

mkaput (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/7016)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @maciektr, @mkaput, and @orizi)

___
*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/7016#-OG-CSGhCE4NKlhjGjMK:-OG-HT9s-DKhy2CRwppL:b-shzna9) (no related file):*

tagging @mkaput  and @maciektr  to validate that we are indeed only adding this plugin in tests - unless very specifically added.
</blockquote></details>

Until #6843 is merged and correctly integrated into the ecosystem, we still suffer from the transient nature of compiler plugins, i.e. snforge_std pulls assert macros, so all Foundry project have these macros available everywhere.

maciektr:

Until #6843 is merged and correctly integrated into the ecosystem, we still suffer from the transient nature of compiler plugins, i.e. snforge_std pulls assert macros, so all Foundry project have these macros available everywhere.
</blockquote></details>

If guarding on test target being present is too limiting, we should guard on `target: 'starknet-contract'` instead, and disallow compilation if this is enabled. This way we can use `assert_eq` outside of tests and still disallow using it in contracts.

maciektr:
> we are indeed only adding this plugin in tests

We are not adding it, users add it themselves explicitly, like this: https://github.com/OpenZeppelin/cairo-contracts/blob/39950b704feb356389c979c11d987d957409d0dc/packages/token/Scarb.toml#L32-L33

mkaput (changes requested):
> Reviewed 1 of 1 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/7016)* status: all files reviewed, 1 unresolved discussion (waiting on @orizi)

PR #7042, opened 2025-01-09
feat(corelib): ByteArray impl Hash trait

### LINKED ISSUE #7015: feat: ByteArray should impl hash
In #6817 I said it should impl IntoIterator, and in a following comment https://github.com/starkware-libs/cairo/issues/6817#issuecomment-2517753410 I suggested I could impl the hash too

IntoIterator was taken care of in #6938, but hash is still missing

hudem1:
Close #7015

Implement `Hash` trait for ByteArray

**Example**

```cairo
let byte_array: ByteArray = "A simple ByteArray";
PoseidonTrait::new().update_with(byte_array).finalize()
```

### DIFF
--- a/corelib/src/hash.cairo
+++ b/corelib/src/hash.cairo
@@ -229,3 +229,16 @@ impl TupleNextHash<
         state.update_with(head).update_with(rest)
     }
 }
+
+impl ByteArrayHash<S, +HashStateTrait<S>, +Drop<S>> of Hash<ByteArray, S> {
+    fn update_state(mut state: S, value: ByteArray) -> S {
+        state = state.update(value.data.len().into());
+
+        for word_index in 0..value.data.len() {
+            let word_in_data = (*value.data.at(word_index)).into();
+            state = state.update(word_in_data);
+        };
+
+        state.update(value.pending_word_len.into()).update(value.pending_word)
+    }
+}
--- a/corelib/src/test/hash_test.cairo
+++ b/corelib/src/test/hash_test.cairo
@@ -85,3 +85,24 @@ fn test_user_defined_hash() {
         'Bad hash of StructForHash',
     );
 }
+
+#[test]
+fn test_byte_array_hash() {
+    let byte_array: ByteArray = "This is a sentence that is longer than 31 characters.";
+
+    let word_1 = 'This is a sentence that is long';
+    let data_length = 1;
+    let pending_word = 'er than 31 characters.';
+    let pending_word_len = 22;
+
+    assert_eq(
+        @PoseidonTrait::new().update_with(byte_array).finalize(),
+        @PoseidonTrait::new()
+            .update(data_length)
+            .update(word_1)
+            .update(pending_word_len)
+            .update(pending_word)
+            .finalize(),
+        'Bad hash for ByteArray',
+    );
+}


hudem1:
> > _[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/7042)_ status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @hudem1)
> 
> _[`corelib/src/hash.cairo` line 244 at r1](https://reviewable.io/reviews/starkware-libs/cairo/7042#-OGEnRmM-YnBUjQfPhS1:-OGEnRmM-YnBUjQfPhS2:b-qr329t) ([raw file](https://github.com/starkware-libs/cairo/blob/1380127b329ece4e5e5b31d77b2f1e5a0384d3e3/corelib/src/hash.cairo#L244)):_
> 
> > ```cairo
> >         state.update(value.pending_word_len.into()).update(value.pending_word)
> >     }
> > }
> > ```
> 
> should probably consider this internally. this looks like the correct implementation for `pedersen` but no necessarily for `poseidon`.
> 
> _Code quote:_
> 
> ```cairo
> impl ByteArrayHash<S, +HashStateTrait<S>, +Drop<S>> of Hash<ByteArray, S> {
>     fn update_state(mut state: S, value: ByteArray) -> S {
>         state = state.update(value.data.len().into());
> 
>         for word_index in 0..value.data.len() {
>             let word_in_data = (*value.data.at(word_index)).into();
>             state = state.update(word_in_data);
>         };
> 
>         state.update(value.pending_word_len.into()).update(value.pending_word)
>     }
> }
> ```

Hi @orizi,

Thank you for reviewing my PR! :)

I understand the concern to generalize the implementation for the 2 different hashing algorithms. Could you clarify a bit what you are suggesting to do ? Do you mean you are going to ask internally at Starkware ?

Also, what do you refer to to make the statement that it looks like the correct implementation for `pedersen` but not necessarily for `poseidon`? Understanding this should help me better grasp how both hashing algorithms handle `ByteArray` and therefore improve my PR :)

Thank you for your feedback!

orizi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/7042)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @hudem1)

___
*[`corelib/src/hash.cairo` line 244 at r1](https://reviewable.io/reviews/starkware-libs/cairo/7042#-OGEnRmM-YnBUjQfPhS1:-OGF2PA50cxURFzZIcJ2:bgl3mr2) ([raw file](https://github.com/starkware-libs/cairo/blob/1380127b329ece4e5e5b31d77b2f1e5a0384d3e3/corelib/src/hash.cairo#L244)):*
<details><summary><i>Previously, hudem1 wrote…</i></summary><blockquote>

Hi @orizi,

Thank you for reviewing my PR! :)

I understand the concern to generalize the implementation for the 2 different hashing algorithms. Could you clarify a bit what you are suggesting to do ? Do you mean you are going to ask internally at Starkware ?

Also, what do you refer to to make the statement that it looks like the correct implementation for `pedersen` but not necessarily for `poseidon`? Understanding this should help me better grasp how both hashing algorithms handle `ByteArray` and therefore improve my PR :)

Thank you for your feedback!

</blockquote></details>

No problem.

I meant discuss internally at starkware indeed.

specifically - poseidon has a capacity part of its hash state, unlike pedersen - this enables us to know what is a "valid" start state, and what isn't.

For example:
assuming the hash of (a, b, c) is hash(hash(a, b), c):
in pedersen - it would look exactly this way.
in poseidon - it would look more like hash(hash((a, 0), b), c)

so in pedersen i would easily claim this is the hash of the pair (not the triplet!) (hash(a, b), c) - and you'd have no choice but to believe me.
in posiedon you won't believe me - as the state would be not the hash of hash((hash(a, b), 0), c).

this doesn't matter if we already only hash triples - but in hashing arrays it might matter!

solution in pedersen - we add the array length to the hash, so you won't be able to start from a middle state (as you gotta hash the length first)

in poseidon - you don't need it, as the capacity takes care of this issue.

tdelabro:
@orizi has some discussion/progress been made internally on this question?

PR #7082, opened 2025-01-14
feat(corelib): OptionTrait::get_or_insert_with

hudem1:
Inserts a value computed from a provided closure into the option if it is [`None`], then returns the contained value.

### Example

```cairo
let mut x = Option::None;
let y = x.get_or_insert_with(|| 5);
assert_eq!(y, 5);

let mut x = Option::Some(2);
let y = x.get_or_insert_with(|| 5);
assert_eq!(y, 2);
```

### DIFF
--- a/corelib/src/option.cairo
+++ b/corelib/src/option.cairo
@@ -595,6 +595,24 @@ pub trait OptionTrait<T> {
     /// assert_eq!(y, Option::None);
     /// ```
     fn take(ref self: Option<T>) -> Option<T>;
+
+    /// Inserts a value computed from `f` into the option if it is [`None`],
+    /// then returns the contained value.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// let mut x = Option::None;
+    /// let y = x.get_or_insert_with(|| 5);
+    /// assert_eq!(y, 5);
+    ///
+    /// let mut x = Option::Some(2);
+    /// let y = x.get_or_insert_with(|| 5);
+    /// assert_eq!(y, 2);
+    /// ```
+    fn get_or_insert_with<F, +core::ops::FnOnce<F, ()>[Output: T], +Drop<F>, +Copy<T>, +Drop<T>>(
+        ref self: Option<T>, f: F,
+    ) -> T;
 }
 
 pub impl OptionTraitImpl<T> of OptionTrait<T> {
@@ -781,6 +799,17 @@ pub impl OptionTraitImpl<T> of OptionTrait<T> {
         let value = self;
         self = Option::None;
         value
+    }
+
+    #[inline]
+    fn get_or_insert_with<F, +core::ops::FnOnce<F, ()>[Output: T], +Drop<F>, +Copy<T>, +Drop<T>>(
+        ref self: Option<T>, f: F,
+    ) -> T {
+        if self.is_none() {
+            self = Option::Some(f());
+        };
+
+        self.unwrap()
     }
 }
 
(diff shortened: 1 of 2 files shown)


orizi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/7082)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @hudem1)

___
*[`corelib/src/option.cairo` line 812 at r1](https://reviewable.io/reviews/starkware-libs/cairo/7082#-OGavhdL8VbjkZ0TGnSK:-OGavhdL8VbjkZ0TGnSL:undefined) ([raw file](https://github.com/starkware-libs/cairo/blob/5218066f60ba23e5cc3587ba99d175dde45ef10e/corelib/src/option.cairo#L812)):*
> ```Cairo
>         };
> 
>         self.unwrap()
> ```





_Suggestion:_
```Cairo
        match self {
            Option::Some(value) => value,
            Option::None => {
                let value = f();
                self = Option::Some(value);
                value    
            }
        }
```

orizi (commented):
i'm not sure we should have this, as it makes much more sense if there's mutability in the language, which there isn't.

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/7082#-:-OGawF4B9aQaE8yYp7FK:b-vvhb3k)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @hudem1)

hudem1 (commented):
Yes, I see what you mean! What should we do then ? The function `take` is in the same case too then.

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/7082#-:-OGfKyVj5klB2FRBtW7b:bqalwpw)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @orizi)

___
*[`corelib/src/option.cairo` line 812 at r1](https://reviewable.io/reviews/starkware-libs/cairo/7082#-OGavhdL8VbjkZ0TGnSK:-OGfM-Z17DVt6dPyBZ7Z:b-896fix) ([raw file](https://github.com/starkware-libs/cairo/blob/5218066f60ba23e5cc3587ba99d175dde45ef10e/corelib/src/option.cairo#L812)):*
> ```Cairo
>         };
> 
>         self.unwrap()
> ```

Done.

orizi (commented):
no - the take is actually ok - as its return value isn't a reference in rust as well.

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/7082#-:-OGfrZjuAHsSWIG6_nUf:b-eeiqxy)* status: 0 of 2 files reviewed, 1 unresolved discussion

hudem1 (commented):
Ok yes okay! Makes sense!

Let me know once you've decided whether we want to include this method. In my opinion, the method still fulfills its core purpose - retrieving a value if it exists, or inserting one if it doesn't - so, it can still be useful. But I understand it won't be as versatile as in Rust where we can get a mutable reference to the value inside.

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/7082#-:-OGl2c181t48aWr6zjZF:b-5m50gv)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @orizi)

orizi (commented):
i think it shouldn't - because:
1. the current version is too easily replaced with:
```
let x = opt.unwrap_or_else(f);
opt = Some(x);
x
```
2. If we do have some sort of mutability added - this would block the function from providing it.

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/7082#-:-OGoiAwU6j_JlEhctgGL:b-1gkd3y)* status: 0 of 2 files reviewed, 1 unresolved discussion

hudem1 (commented):
yeah ok I understand! I will close the PR then!

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/7082#-:-OGpbNpv2lp34Q86_WcJ:b92ypxg)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @orizi)

PR #7096, opened 2025-01-15
fix: Remove unnecessary semicolon after while let block

famouswizard:
I noticed there was an extra semicolon after the closing curly brace of a `while let` block.
In Cairo (similar to Rust), this isn't required and can cause a syntax error. I've removed it to fix the issue.

### DIFF
--- a/examples/complex_input.cairo
+++ b/examples/complex_input.cairo
@@ -17,7 +17,7 @@ fn complex_input(
     r += a_input.val;
     while let Option::Some(x) = a_input.arr.pop_front() {
         r += x;
-    };
+    }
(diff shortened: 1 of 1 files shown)


orizi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/7096)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @famouswizard)

___
*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/7096#-OGftfxa7z1Cy_xbxEA4:-OGftfxa7z1Cy_xbxEA5:b7opwo8) (no related file):*
in general you are correct - in practice you are wrong - this doesn't compile.

please close PR.

PR #7453, opened 2025-03-11
Allow adding `end_ptr` and additional `notes` to `PluginDiagnostic`s

DelevoXDG:


orizi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/7453)* status: 0 of 5 files reviewed, 2 unresolved discussions

___
*[`crates/cairo-lang-defs/src/plugin.rs` line 73 at r1](https://reviewable.io/reviews/starkware-libs/cairo/7453#-OL8Kb8uDHBPOJ4ooxt9:-OL8Kb8uDHBPOJ4ooxtA:b-45b5sa) ([raw file](https://github.com/starkware-libs/cairo/blob/179b9bc62b3d06ea345b676ab8ed56924f8dbcc9/crates/cairo-lang-defs/src/plugin.rs#L73)):*
> ```XML
> pub struct PluginDiagnostic {
>     pub stable_ptr: SyntaxStablePtrId,
>     pub end_ptr: Option<SyntaxStablePtrId>,
> ```

you example didn't actually require the end ptr - you just pointed to another stable syntax?
do you have a particular example that actually requires this?

in any case - separate this PR for the ability to have a note and this in any case.
___
*[`crates/cairo-lang-defs/src/plugin.rs` line 76 at r1](https://reviewable.io/reviews/starkware-libs/cairo/7453#-OL8LAR275K_FogoCAAP:-OL8LAR275K_FogoCAAQ:bvmdm1x) ([raw file](https://github.com/starkware-libs/cairo/blob/179b9bc62b3d06ea345b676ab8ed56924f8dbcc9/crates/cairo-lang-defs/src/plugin.rs#L76)):*
> ```XML
>     pub message: String,
>     pub severity: Severity,
>     pub note: Option<String>,
> ```

can't your example of "originating attribute" be just handled from the outside by the macro generation itself?

just add the info about the causing attr when remapping the code?

orizi (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/7453)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @DelevoXDG)

___
*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/7453#-OL9hscy5COH5bwm2q7S:-OL9hscy5COH5bwm2q7T:b-szgcn9) (no related file):*
@gilbens-starkware as he worked on something similar as part of his work of user defined inline macros.

DelevoXDG:
*[`crates/cairo-lang-defs/src/plugin.rs` line 76 at r1](https://reviewable.io/reviews/starkware-libs/cairo/7453#-OL8LAR275K_FogoCAAP:-OL9gzTb30tSizDKuyKk:b-gxmpkd) ([raw file](https://github.com/starkware-libs/cairo/blob/179b9bc62b3d06ea345b676ab8ed56924f8dbcc9/crates/cairo-lang-defs/src/plugin.rs#L76)):*

can't your example of "originating attribute" be just handled from the outside by the macro generation itself?

just add the info about the causing attr when remapping the code?
</blockquote></details>

I'm not sure I understand what you mean by "outside". 

Diagnostics, including notes, are formatted by the compiler, not during macro generation itself:
```rs
pub fn format_with_severity(
    &self,
    db: &TEntry::DbType,
    file_notes: &OrderedHashMap<FileId, DiagnosticNote>,
) -> Vec<FormattedDiagnosticEntry> {
    // ...
    for entry in &self.get_diagnostics_without_duplicates(db) {
        // ...
        for note in entry.notes(db) {
            msg += &format!("note: {:?}\n", note.debug(files_db))
        }
        for note in parent_file_notes {
            msg += &format!("note: {:?}\n", note.debug(files_db))
        }
        // ...
    }
    // ...
}
```

On Scarb level during macro expansion, we cannot display, but only can generate such notes, which we already do in other circumstances - for compiler-emitted diagnostics:
```rs
PluginResult {
    code: Some(PluginGeneratedFile {
        // ...
        diagnostics_note: Some(format!(
            "this error originates in the attribute macro: `{}`",
            input.expansion.name
        )),
        // ...
    }),
    diagnostics: into_cairo_diagnostics(
        db, 
        result.diagnostics, 
        call_site.stable_ptr,
    ),
    remove_original_item: true,
}
```

We don't always generate a plugin file that would include the diagnostics note. This is skipped if the macro did not produce any changes:
```rs
if last && result.aux_data.is_none() && original == result.token_stream.to_string() {
    return PluginResult {
        code: None,
        remove_original_item: false,
        diagnostics: into_cairo_diagnostics(
            db,
            result.diagnostics, 
            call_site.stable_ptr,
            // Without these parameters, we can't add the note
            Some(&input.expansion.name),
            Some(input.expansion.kind.clone()),
        ),
    };
}
```

Since we cannot use the samis concept for all cases, I added something similar to what we already have for compiler diagnostics, except for plugin diagnostics.

Can you specify why do you think suggested handling of notes is problematic, and how do you envision they should be stored/handled?

DelevoXDG:
*[`crates/cairo-lang-defs/src/plugin.rs` line 76 at r1](https://reviewable.io/reviews/starkware-libs/cairo/7453#-OL8LAR275K_FogoCAAP:-OLhaZuC-izDE8vVRPfR:b-ykoal7) ([raw file](https://github.com/starkware-libs/cairo/blob/179b9bc62b3d06ea345b676ab8ed56924f8dbcc9/crates/cairo-lang-defs/src/plugin.rs#L76)):*
> what i'm saying is that you can use it when mapping diagnostics through the code-mappings, and add the specific notes there.

Could you clarify what do you mean by "there"? Are you suggesting somehow adding diagnostics to the specific mappings? If so, I’m not sure that would work, as we should avoid conflicts between notes used by regular diagnostics and those from macro-emitted diagnostics.

Additionally, I don’t believe we have direct access to the **name** and **type** of the expansion at the level where mappings are used, which we need for generating a proper note. 

Is there a specific issue with the current approach that you’re concerned about? 

I've considered adding `note` to `PluginResult` instead so we have less duplicate note strings saved (if that is problematic 🤔), but access to it doesn’t seem to persist at the `DiagnosticEntry` level where notes are created.

In any case, current approach seems to be in line in with the prior diagnostics handling, as similar notes were generated at Scarb level, not during mapping or elsewhere. I'd like to better understand the specific issue you’re trying to solve - then it'd be easier to discuss alternative approaches.

orizi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/7453)* status: 0 of 6 files reviewed, 2 unresolved discussions (waiting on @DelevoXDG)

___
*[`crates/cairo-lang-defs/src/plugin.rs` line 76 at r1](https://reviewable.io/reviews/starkware-libs/cairo/7453#-OL8LAR275K_FogoCAAP:-OLiO7uvDToHQB0E4bFQ:b-n99ibv) ([raw file](https://github.com/starkware-libs/cairo/blob/179b9bc62b3d06ea345b676ab8ed56924f8dbcc9/crates/cairo-lang-defs/src/plugin.rs#L76)):*
<details><summary><i>Previously, DelevoXDG (Maksim Zdobnikau) wrote…</i></summary><blockquote>

> what i'm saying is that you can use it when mapping diagnostics through the code-mappings, and add the specific notes there.

Could you clarify what do you mean by "there"? Are you suggesting somehow adding diagnostics to the specific mappings? If so, I’m not sure that would work, as we should avoid conflicts between notes used by regular diagnostics and those from macro-emitted diagnostics.

Additionally, I don’t believe we have direct access to the **name** and **type** of the expansion at the level where mappings are used, which we need for generating a proper note. 

Is there a specific issue with the current approach that you’re concerned about? 

I've considered adding `note` to `PluginResult` instead so we have less duplicate note strings saved (if that is problematic 🤔), but access to it doesn’t seem to persist at the `DiagnosticEntry` level where notes are created.

In any case, current approach seems to be in line in with the prior diagnostics handling, as similar notes were generated at Scarb level, not during mapping or elsewhere. I'd like to better understand the specific issue you’re trying to solve - then it'd be easier to discuss alternative approaches.
</blockquote></details>

the main thing is - you are adding a method to do something that is going to be EXACTLY the same for every plugin diagnostic - adding the causing location as well.

it makes no sense to me for the plugin developers to do so - therefore - they shouldn't do it.

do note that enriching the amount of information appearing on `VirtualFile`s and using these while mapping a diagnostic location to add note - would work.

currently, as you know, we map every diagnostics location through all the parents files location mappings, we can just add additional steps to add extra information. but it would be workable for all diagnostics, not just a few by hand.

DelevoXDG:
*[`crates/cairo-lang-defs/src/plugin.rs` line 76 at r1](https://reviewable.io/reviews/starkware-libs/cairo/7453#-OL8LAR275K_FogoCAAP:-OLo5Jc6AS4OJe7Ug8uE:b-o61ctm) ([raw file](https://github.com/starkware-libs/cairo/blob/179b9bc62b3d06ea345b676ab8ed56924f8dbcc9/crates/cairo-lang-defs/src/plugin.rs#L76)):*
> the main thing is - you are adding a method to do something that is going to be EXACTLY the same for every plugin diagnostic - adding the causing location as well.

I don't think it's entirely accurate to say it will be exactly the same, because:
1. The note will vary depending on the macro.
2. Even for the **same macro**, the notes will differ based on context:
  - We only want to add such note for plugin diagnostics **with span** for use case described in the description:
     ```
    error: Plugin diagnostic: Tuple type is not supported for
    --> [..]lib.cairo:4:8
    y: (u32, u64),
    ^^^^^^^^^^^
    note: this diagnostic originates in the attribute macro: `some`
    ```
  - Otherwise, for plugin diagnostics with no location, we just point to the expansion call site, and in that case, additional note would be just redundant:
    ```
    error: Plugin diagnostic: Some error from macro.
     --> [..]lib.cairo:1:1
    #[some]
    ^^^^^^^
    ```
 
> it makes no sense to me for the plugin developers to do so - therefore - they shouldn't do it.

The method is entirely **optional**: plugin developers don't have to use it. Diagnostics can still be emitted as before, and for cases where diagnostic **with location** is emmited, Scarb will handle adding the note automatically.

> do note that enriching the amount of information appearing on `VirtualFile`s and using these while mapping a diagnostic location to add note - would work.

As I previously mentioned, we do not generate `PluginGeneratedFile` for every expansion, as an optimization. This means not every expansion has a corresponding `VirtualFile`:
```
fn priv_module_sub_files(
    db: &dyn DefsGroup,
    module_id: ModuleId,
    file_id: FileId,
) -> Maybe<Arc<PrivModuleSubFiles>> {
    ...
    let mut diagnostics_notes = OrderedHashMap::default();
    for item_ast in item_asts.elements(syntax_db) {
        ...
        for plugin_id in db.crate_macro_plugins(crate_id).iter() {
            ...
            let result = plugin.generate_code(db.upcast(), item_ast.clone(), &metadata);
            ...
            if let Some(generated) = result.code {
                let generated_file_id = FileLongId::External(
                    PluginGeneratedFileLongId {
                        module_id,
                        stable_ptr: item_ast.stable_ptr().untyped(),
                        name: generated.name.clone(),
                    }
                    .intern(db)
                    .as_intern_id(),
                )
                .intern(db);
                if let Some(text) = generated.diagnostics_note {
                    diagnostics_notes.insert(
                        generated_file_id,
                        DiagnosticNote { text, location: None }
                    );
                }
                files.insert(
                    generated_file_id,
                    VirtualFile {
                        parent: Some(file_id),
                        name: generated.name,
                        content: generated.content.into(),
                        code_mappings: generated.code_mappings.into(),
                        kind: FileKind::Module,
                    },
                );
                ...
            }
            ...
        }
        ...
    }
    ...
    Ok(res.into())
}
```
Should we be creating `VirtualFile` for every expansion, at the cost of performance and adding **unnecessary** notes?

orizi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/7453)* status: 0 of 6 files reviewed, 2 unresolved discussions (waiting on @DelevoXDG)

___
*[`crates/cairo-lang-defs/src/plugin.rs` line 76 at r1](https://reviewable.io/reviews/starkware-libs/cairo/7453#-OL8LAR275K_FogoCAAP:-OM7PeMoExWQEurOIal2:bdwsynf) ([raw file](https://github.com/starkware-libs/cairo/blob/179b9bc62b3d06ea345b676ab8ed56924f8dbcc9/crates/cairo-lang-defs/src/plugin.rs#L76)):*
<details><summary><i>Previously, DelevoXDG (Maksim Zdobnikau) wrote…</i></summary><blockquote>

> the main thing is - you are adding a method to do something that is going to be EXACTLY the same for every plugin diagnostic - adding the causing location as well.

I don't think it's entirely accurate to say it will be exactly the same, because:
1. The note will vary depending on the macro.
2. Even for the **same macro**, the notes will differ based on context:
  - We only want to add such note for plugin diagnostics **with span** for use case described in the description:
     ```
    error: Plugin diagnostic: Tuple type is not supported for
    --> [..]lib.cairo:4:8
    y: (u32, u64),
    ^^^^^^^^^^^
    note: this diagnostic originates in the attribute macro: `some`
    ```
  - Otherwise, for plugin diagnostics with no location, we just point to the expansion call site, and in that case, additional note would be just redundant:
    ```
    error: Plugin diagnostic: Some error from macro.
     --> [..]lib.cairo:1:1
    #[some]
    ^^^^^^^
    ```
 
> it makes no sense to me for the plugin developers to do so - therefore - they shouldn't do it.

The method is entirely **optional**: plugin developers don't have to use it. Diagnostics can still be emitted as before, and for cases where diagnostic **with location** is emmited, Scarb will handle adding the note automatically.

> do note that enriching the amount of information appearing on `VirtualFile`s and using these while mapping a diagnostic location to add note - would work.

As I previously mentioned, we do not generate `PluginGeneratedFile` for every expansion, as an optimization. This means not every expansion has a corresponding `VirtualFile`:
```
fn priv_module_sub_files(
    db: &dyn DefsGroup,
    module_id: ModuleId,
    file_id: FileId,
) -> Maybe<Arc<PrivModuleSubFiles>> {
    ...
    let mut diagnostics_notes = OrderedHashMap::default();
    for item_ast in item_asts.elements(syntax_db) {
        ...
        for plugin_id in db.crate_macro_plugins(crate_id).iter() {
            ...
            let result = plugin.generate_code(db.upcast(), item_ast.clone(), &metadata);
            ...
            if let Some(generated) = result.code {
                let generated_file_id = FileLongId::External(
                    PluginGeneratedFileLongId {
                        module_id,
                        stable_ptr: item_ast.stable_ptr().untyped(),
                        name: generated.name.clone(),
                    }
                    .intern(db)
                    .as_intern_id(),
                )
                .intern(db);
                if let Some(text) = generated.diagnostics_note {
                    diagnostics_notes.insert(
                        generated_file_id,
                        DiagnosticNote { text, location: None }
                    );
                }
                files.insert(
                    generated_file_id,
                    VirtualFile {
                        parent: Some(file_id),
                        name: generated.name,
                        content: generated.content.into(),
                        code_mappings: generated.code_mappings.into(),
                        kind: FileKind::Module,
                    },
                );
                ...
            }
            ...
        }
        ...
    }
    ...
    Ok(res.into())
}
```
Should we be creating `VirtualFile` for every expansion, at the cost of performance and adding **unnecessary** notes?
</blockquote></details>

you definitely already ARE creating a VirtualFile for every expansion - which is exposed through an ExternalFileId.
it is converted into a VirtualFile for the actual usage. (it used to be a VirtualFile, and i converted it to this method, in order to have more stable ptrs).

My issue is that you keep giving the exact same example of:
having an error pointing at a peace of code, and than mapping it to another piece of code with additional info - which was ALWAYS just the originating stable ptr of the code causing the expansion.

it sounds here as if - everytime you get an example that is mapped to the original code, you just want to add a note for what is the code that caused the expansion.

this seems like a possible complication of implementation of plugins - where all you actually wanted was this additional info.

no - there won't be any additional performance issues because of this.

orizi (commented):
> Reviewed 2 of 5 files at r1, 3 of 3 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/7453)* status: all files reviewed, 2 unresolved discussions (waiting on @DelevoXDG)

maciektr:
Closing for now, we will reopen if re-attempted

PR #7576, opened 2025-04-03
Made bounded-int pub, and blocked by unstable feature.

### LINKED ISSUE #7543: feat: Make bounded_int, upcast, downcast public in corelib
Those are very useful and I don't understand why they aren't public. 

Say for example i want to split a u128 into two u64. 

I could use div_rem(u128, 2^64_u128) to get two u128, then i need to call .try_into().unwrap() on them although they are in range by construction (that's the nice part about bounded ints that we can't use)

With bounded int i can use div_rem then upcast() which is clearly cheaper

I use those extensively and this prevents me to use the latest Cairo 2024 edition.

orizi:
Fixes #7543.


gilbens-starkware (approved):
:lgtm:

> Reviewed 9 of 9 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/7576#-:-ON9FbEc0Si6FJAWWj9T:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @orizi)

PR #7593, opened 2025-04-06
fix: argument order in cargo clippy command

0xwitty:
Hey, I noticed that user-provided arguments (`$@`) were placed before the standard `cargo clippy` flags. This could cause conflicts or unexpected behavior if user args overlap with cargo's own flags.  

I moved `$@` to the correct spot—after `cargo clippy` but before `--all-targets`, etc.—so everything gets passed as intended.

### DIFF
--- a/scripts/clippy.sh
+++ b/scripts/clippy.sh
@@ -2,5 +2,5 @@
 
 export RUSTUP_TOOLCHAIN="${RUSTUP_TOOLCHAIN:-nightly-2025-04-04}"
 
-cargo clippy "$@" --all-targets --all-features -- -D warnings -D future-incompatible \
+cargo clippy --all-targets --all-features "$@" -- -D warnings -D future-incompatible \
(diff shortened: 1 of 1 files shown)


orizi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/7593)* status: 0 of 1 files reviewed, 2 unresolved discussions

___
*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/7593#-ON9jgkQ3ND3wDDqKSgo:-ON9jgkR3pmMierY3rH1:b-704x0k) (no related file):*
this is for our internal usage - so not really relevant.
___
*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/7593#-ON9kjIl67siEpUzdUcr:-ON9kjIl67siEpUzdUcs:b-yfsm08) (no related file):*
fixing of internal scripts for ease of use is done when running into issues.

we no longer accept PRs for such improvements.

PR #7637, opened 2025-04-10
Add matches macro

wawel37:
## Rationale

1. It will be helpful to have such a macro as it really makes it easier to write specific code and makes the code less bloated. It's also implemented for the Rust language, so people should already be familiar with it.
2. It will be really helpful for the fixer inside the `cairo-lint`, as we want to fix such code:
```cairo
enum Abc {
    A: u32,
    B: u32,
    C: u32
}

fn main() {
    let a = Abc::A(1);
    if let Abc::A(x) = a {
        panic!("a shouldn't be type of Abc::A");
    }
}
```

with just 

```cairo
fn main() {
    let a = Abc::A(1);
    assert!(!matches!(a, Abc::A(_)), "a shouldn't be type of Abc::A");
}
```

rather than

```cairo
fn main() {
    let a = Abc::A(1);
    assert!(!{match a {
        Abc::A(_) => true,
        _ => false,
    }}, "a shouldn't be type of Abc::A");
}
```

### DIFF
--- a/corelib/src/test/language_features/matches_test.cairo
+++ b/corelib/src/test/language_features/matches_test.cairo
@@ -1,0 +1,33 @@
+#[test]
+fn test_basic_match_case() {
+    let a = Some(1);
+    if !matches!(a, Some(_)) {
+        panic!("Expected `a` to be Option::Some, not Option::None");
+    }
+}
+
+enum TestEnum {
+    A,
+    B,
+}
+
+enum TestEnumWithData {
+    A: u32,
+    B: u32,
+}
+
+#[test]
+fn test_enum_match_case() {
+    let a = TestEnum::A;
+    if !matches!(a, TestEnum::A) {
+        panic!("Expected `a` to be TestEnum::A");
+    }
+}
+
+#[test]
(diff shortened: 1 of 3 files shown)


integraledelebesgue (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/7637#-:-ONUaFTa9KF2u7VuPsPe:bnfp4nl)* status: 0 of 3 files reviewed, all discussions resolved (waiting on @orizi)

orizi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/7637)* status: 0 of 3 files reviewed, 1 unresolved discussion

___
*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/7637#-ONUfWTM8P51Mnf8eI4l:-ONUfWTM8P51Mnf8eI4m:bqtfn58) (no related file):*
user defined inline macro support is on the way (and very close).
no allowing further core plguins for the time being without an extremely good reason for the time being.

PR #7827, opened 2025-06-10
Fix lint diagnostics for `write!` inline macro

wawel37:
The `(())` is not necessary here as the type is inferred. This change is applied to the cairo-lint standards of the code. For more context see https://github.com/software-mansion/cairo-lint/blob/84709c709d339740c6781b74a09e88ddd10ded38/src/lints/redundant_brackets_in_enum_call.rs#L47

### DIFF
--- a/crates/cairo-lang-semantic/src/inline_macros/write.rs
+++ b/crates/cairo-lang-semantic/src/inline_macros/write.rs
@@ -389,7 +389,7 @@ impl FormattingInfo {
 
         self.flush_pending_chars(builder, &mut pending_chars, ident_count);
         self.add_indentation(builder, ident_count);
-        builder.add_str("core::result::Result::<(), core::fmt::Error>::Ok(())\n");
+        builder.add_str("core::result::Result::<(), core::fmt::Error>::Ok\n");
         while ident_count > 1 {
             ident_count -= 1;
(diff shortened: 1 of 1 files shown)


orizi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/7827)* status: 0 of 1 files reviewed, 1 unresolved discussion

___
*[`crates/cairo-lang-semantic/src/inline_macros/write.rs` line 392 at r1](https://reviewable.io/reviews/starkware-libs/cairo/7827#-OSP3Vp57b8NYJDK24ay:-OSP3Vp57b8NYJDK24az:bil7ijn) ([raw file](https://github.com/starkware-libs/cairo/blob/e57a949d190e790e2c44f8ea8d98083fd4b7fa09/crates/cairo-lang-semantic/src/inline_macros/write.rs#L392)):*
> ```XML
>         self.flush_pending_chars(builder, &mut pending_chars, ident_count);
>         self.add_indentation(builder, ident_count);
>         builder.add_str("core::result::Result::<(), core::fmt::Error>::Ok\n");
> ```

only explicit unit types should be done that way.
so i don't think this should be prefered (Result::Ok is incidentally a unit type, but not in the enum itself)

PR #7851, opened 2025-06-18
Expose core gas utilities for external consumption

ksew1:
Will be used in `cairo-profiler`

### DIFF
--- a/crates/cairo-lang-sierra-gas/src/lib.rs
+++ b/crates/cairo-lang-sierra-gas/src/lib.rs
@@ -28,7 +28,7 @@ use thiserror::Error;
 
 pub mod compute_costs;
 pub mod core_libfunc_cost;
-mod core_libfunc_cost_base;
+pub mod core_libfunc_cost_base;
 mod core_libfunc_cost_expr;
 mod cost_expr;
 pub mod gas_info;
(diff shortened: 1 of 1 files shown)


orizi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/7851)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @ksew1 and @THenry14)

___
*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/7851#-OT28x70CUAC4wpsAlN3:-OU0AnJEDWbB0rn_BeZ7:bk9hx8m) (no related file):*
<details><summary><i>Previously, THenry14 (Wojciech Szymczyk) wrote…</i></summary><blockquote>

I see your point, but I don't think it is that simple - `core\_libfunc\_cost` is quite large, and from what I can see it still is getting some modifications (I can see some commits from last month) - not only it would mean we basically need to copy over bigger part of cairo-lang-sierra-gas crate (at least), that would also mean we need to constantly track upstream changes to make sure we do not omit something new accidentally. Given that there is no eta on the refactors (I don't think it will happen any time soon for various reasons), and plans to put a bit less effort into profiler in the near future, I'm sure you see how this is somewhat suboptimal from our PoV.

Btw do you mind sharing what are the reasons why these should not be public? that would perhaps help me better understand the exact issue/pain point here
</blockquote></details>

you don't need to duplicate the large logic at all - you just need to implement simple implementations for:
`InvocationApChangeInfoProvider`  and `InvocationCostInfoProvider`.

this would allow you to get the basic prices - no need to duplicate the full logic.

PR #7876, opened 2025-06-23
Skip `generate_code` calls if possible

Draggu:
This allow LS to reduce `generate_code` part from 18.5 to 7.8 second while testing on https://github.com/OpenZeppelin/cairo-contracts.

Used Mac m4 Pro 24gb.

### DIFF
--- a/crates/cairo-lang-defs/src/db.rs
+++ b/crates/cairo-lang-defs/src/db.rs
@@ -1,4 +1,4 @@
-use std::collections::VecDeque;
+use std::collections::{HashSet, VecDeque};
 use std::sync::Arc;
 
 use cairo_lang_diagnostics::{DiagnosticNote, Maybe, PluginFileDiagnosticNotes, ToMaybe};
@@ -10,7 +10,7 @@ use cairo_lang_syntax::attribute::consts::{
     STARKNET_INTERFACE_ATTR, UNSTABLE_ATTR,
 };
 use cairo_lang_syntax::attribute::structured::AttributeStructurize;
-use cairo_lang_syntax::node::ast::MaybeModuleBody;
+use cairo_lang_syntax::node::ast::{MaybeImplBody, MaybeModuleBody, MaybeTraitBody};
 use cairo_lang_syntax::node::db::SyntaxGroup;
 use cairo_lang_syntax::node::element_list::ElementList;
 use cairo_lang_syntax::node::helpers::QueryAttrs;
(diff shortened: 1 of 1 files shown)


orizi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/7876)* status: 0 of 1 files reviewed, 1 unresolved discussion

___
*[`crates/cairo-lang-defs/src/db.rs` line 801 at r1](https://reviewable.io/reviews/starkware-libs/cairo/7876#-OTRcWyl8BvSIuZuW2wv:-OTRcWyl8BvSIuZuW2ww:b-udszo) ([raw file](https://github.com/starkware-libs/cairo/blob/f1090f350b6d5a00c923731edbe24a0610eb1634/crates/cairo-lang-defs/src/db.rs#L801)):*
> ```XML
>             {
>                 continue;
>             };
> ```

couldn't your plugin just check it?
why is this here?

this changes the ACTUAL definition of plugin macros - which can do anything in general.



_Code quote:_
```XML
            if !plugin.declared_attributes().into_iter().any(|attr| item_ast.has_attr(db, &attr))
                // Plugins can implement own derives.
                && !item_ast.has_attr(db, "derive")
                // Plugins does not declare module inline macros they support.
                && !matches!(item_ast, ast::ModuleItem::InlineMacro(_))
            {
                continue;
            };
```

orizi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/7876)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @Draggu)

___
*[`crates/cairo-lang-defs/src/db.rs` line 801 at r1](https://reviewable.io/reviews/starkware-libs/cairo/7876#-OTRcWyl8BvSIuZuW2wv:-OTSyGb87p-YTsLnOm5E:b-tnfshz) ([raw file](https://github.com/starkware-libs/cairo/blob/f1090f350b6d5a00c923731edbe24a0610eb1634/crates/cairo-lang-defs/src/db.rs#L801)):*
<details><summary><i>Previously, Draggu (Piotr Figiela) wrote…</i></summary><blockquote>

>couldn't your plugin just check it?

It does, we skip mostly built-in plugins with this.

>why is this here?

If u believe there is a better place for this let me know but I think it belongs here.

>this changes the ACTUAL definition of plugin macros - which can do anything in general.

And the question here is if they should? Like it's name states `MacroPlugin` should deal with macros. This change does not touch `AnalyzerPlugin`s which I believe should run on everything.  We can use the fact that these plugins declare what they are interested in and optimize expansion process. Execution time gain is significant here.
</blockquote></details>

this seems like a thing that a plugin macro can do by itself - this literaly has nothing that a plugin macro cannot already do.

which plugin does not already filter out irrelevant parts according to attribute? where is the actual gain?

at the most extreme case - adding a "preliminary_elimination" function for plugin macro and having this as a specific default implementation would make no sense.

i still don't see which built-in macro actually costs more due to this specific lack of filtering.

PR #7894, opened 2025-06-29
Correctly handle inline macro in tail position

dean-starkware:


dean-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/7894)* status: 0 of 2 files reviewed, 4 unresolved discussions (waiting on @orizi)

___
*[`crates/cairo-lang-semantic/src/expr/compute.rs` line 670 at r1](https://reviewable.io/reviews/starkware-libs/cairo/7894#-OTvkQKD9Ljd2FWumI83:-OU-CLdJ5nO5dkDkREX4:b-896fix) ([raw file](https://github.com/starkware-libs/cairo/blob/af9d3ba81c44410cce05091c076ef100e6f5c24d/crates/cairo-lang-semantic/src/expr/compute.rs#L670)):*

this is an implementation detail - the main aim of this function is not the recursion - it is just to lower the tail of a block.
</blockquote></details>

Done.
___
*[`crates/cairo-lang-semantic/src/expr/compute.rs` line 677 at r1](https://reviewable.io/reviews/starkware-libs/cairo/7894#-OTvksDTBnTx7iqdDqDB:-OTzmIiaCp7nCjyadiem:b-896fix) ([raw file](https://github.com/starkware-libs/cairo/blob/af9d3ba81c44410cce05091c076ef100e6f5c24d/crates/cairo-lang-semantic/src/expr/compute.rs#L677)):*

it seems you decided to add a new block when expanding a tail.

it indeed should always work - but you have do heavily doc in the implementation here why this works.
</blockquote></details>

Done.
___
*[`crates/cairo-lang-semantic/src/expr/compute.rs` line 751 at r1](https://reviewable.io/reviews/starkware-libs/cairo/7894#-OTvlOkz7zdrcPgEfsWx:-OTztKl0-wbFtCa5hxpT:b-hy2e1y) ([raw file](https://github.com/starkware-libs/cairo/blob/af9d3ba81c44410cce05091c076ef100e6f5c24d/crates/cairo-lang-semantic/src/expr/compute.rs#L751)):*

won't this just work already?
</blockquote></details>

You're right, since `compute_tail_semantic` already handles inline macros recursively, we don't need to special-case them here.
___
*[`crates/cairo-lang-semantic/src/expr/compute.rs` line 776 at r1](https://reviewable.io/reviews/starkware-libs/cairo/7894#-OTvl_1Q5lxRunH6JOXo:-OU-BL3y-WxZR6nw4vTA:b-xjxstk) ([raw file](https://github.com/starkware-libs/cairo/blob/af9d3ba81c44410cce05091c076ef100e6f5c24d/crates/cairo-lang-semantic/src/expr/compute.rs#L776)):*

this should work due to your block tail hack.
</blockquote></details>

I though so at first too.
But macros in statement position must be expanded and processed as statements, not just as expressions.
-   Many diagnostics (like "Identifier not found" in nested macro expansions) are only produced if you recursively process all statements produced by macros in statement position.
-   Without this, errors in nested macros can be silently missed.

dean-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/7894)* status: 0 of 2 files reviewed, 5 unresolved discussions (waiting on @orizi)

___
*[`crates/cairo-lang-semantic/src/expr/compute.rs` line 493 at r2](https://reviewable.io/reviews/starkware-libs/cairo/7894#-OU-I9N03KU-I1VWWNBL:-OU-_Qxi3htAeXbHwnla:b-omi72u) ([raw file](https://github.com/starkware-libs/cairo/blob/d4c093c48ef2617491f8a749e1a9c6a41936698f/crates/cairo-lang-semantic/src/expr/compute.rs#L493)):*

this almost certainly should not be here.

at the very least - this has NOTHING to do with the FieldInitShorthand case.
</blockquote></details>

I agree that this logic shouldn’t be tied to `FieldInitShorthand`, and I’ll fix that so the logic related to `Missing` is separated properly.
However, I’m struggling to find an elegant way to move this code outside of the `Missing` arm and out of `maybe_compute_expr_semantic`.
Do you have any idea or suggestion for where this logic could be moved? Or should we just keep it in this function (not under `FieldInitShorthand` though)?

dean-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/7894)* status: 0 of 2 files reviewed, 5 unresolved discussions (waiting on @orizi)

___
*[`crates/cairo-lang-semantic/src/expr/compute.rs` line 776 at r1](https://reviewable.io/reviews/starkware-libs/cairo/7894#-OTvl_1Q5lxRunH6JOXo:-OU4Ag0W1-8NyuyM02cX:b-896fix) ([raw file](https://github.com/starkware-libs/cairo/blob/af9d3ba81c44410cce05091c076ef100e6f5c24d/crates/cairo-lang-semantic/src/expr/compute.rs#L776)):*

this should all be processed while expanding the macro.
if it isn't - the bug is somewhere else.
</blockquote></details>

Done.
___
*[`crates/cairo-lang-semantic/src/expr/compute.rs` line 493 at r2](https://reviewable.io/reviews/starkware-libs/cairo/7894#-OU-I9N03KU-I1VWWNBL:-OU4AhrZ-r081YAurmn3:b-896fix) ([raw file](https://github.com/starkware-libs/cairo/blob/d4c093c48ef2617491f8a749e1a9c6a41936698f/crates/cairo-lang-semantic/src/expr/compute.rs#L493)):*
<details><summary><i>Previously, dean-starkware wrote…</i></summary><blockquote>

I agree that this logic shouldn’t be tied to `FieldInitShorthand`, and I’ll fix that so the logic related to `Missing` is separated properly.
However, I’m struggling to find an elegant way to move this code outside of the `Missing` arm and out of `maybe_compute_expr_semantic`.
Do you have any idea or suggestion for where this logic could be moved? Or should we just keep it in this function (not under `FieldInitShorthand` though)?
</blockquote></details>

Done.

dean-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/7894)* status: 0 of 2 files reviewed, 3 unresolved discussions (waiting on @orizi)

___
*[`crates/cairo-lang-semantic/src/expr/compute.rs` line 649 at r3](https://reviewable.io/reviews/starkware-libs/cairo/7894#-OU4Mz3n84_GWRgOXyjh:-OU4ciSc1OVKTLtHa14i:b-qyxbyf) ([raw file](https://github.com/starkware-libs/cairo/blob/dd2ee3f221750a032a05e014f0edd3a372ce5fa9/crates/cairo-lang-semantic/src/expr/compute.rs#L649)):*

doesn't this have to be called from `compute_expr_block_semantic` as well?
add a failing test before fixing.
</blockquote></details>

You're right. Before the refactor, this wasn’t possible because the implementation wasn’t usable in `compte_expr_block_semantic`.  
Now that we've simplified everything, it’s possible, and been fixed.  
I verified it with a suitable test before making the fix.
___
*[`crates/cairo-lang-semantic/src/expr/compute.rs` line 665 at r3](https://reviewable.io/reviews/starkware-libs/cairo/7894#-OU4NhXHERC76wOn7bfH:-OU4i3252gxTW8gLdCeZ:bn4ooht) ([raw file](https://github.com/starkware-libs/cairo/blob/dd2ee3f221750a032a05e014f0edd3a372ce5fa9/crates/cairo-lang-semantic/src/expr/compute.rs#L665)):*

why not just call the `expand_macro_for_statement`?
</blockquote></details>

We need to use a dummy pointer here because `expand_macro_for_statement` requires a `statement_stable_ptr`, but when expanding an inline macro in a tail expression position, we don’t have a real statement node. The best we can do is use the stable pointer of the macro expression itself. However, I would prefer not to use a dummy pointer at all, which is why I chose not to use `expand_macro_for_statement` in this context.
___
*[`crates/cairo-lang-semantic/src/expr/compute.rs` line 2182 at r3](https://reviewable.io/reviews/starkware-libs/cairo/7894#-OU4Mp434tAF09YWrhc3:-OU4gaTR0U0B4DvxhGuW:b-896fix) ([raw file](https://github.com/starkware-libs/cairo/blob/dd2ee3f221750a032a05e014f0edd3a372ce5fa9/crates/cairo-lang-semantic/src/expr/compute.rs#L2182)):*

isn't this the exact same logic as in `compute_loop_body_semantic`?
</blockquote></details>

Done.

orizi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/7894)* status: 0 of 2 files reviewed, 4 unresolved discussions (waiting on @dean-starkware)

___
*[`crates/cairo-lang-semantic/src/expr/compute.rs` line 665 at r3](https://reviewable.io/reviews/starkware-libs/cairo/7894#-OU4NhXHERC76wOn7bfH:-OU4yY9MBIfKV6r15yxb:b-zksuq) ([raw file](https://github.com/starkware-libs/cairo/blob/dd2ee3f221750a032a05e014f0edd3a372ce5fa9/crates/cairo-lang-semantic/src/expr/compute.rs#L665)):*
<details><summary><i>Previously, dean-starkware wrote…</i></summary><blockquote>

We need to use a dummy pointer here because `expand_macro_for_statement` requires a `statement_stable_ptr`, but when expanding an inline macro in a tail expression position, we don’t have a real statement node. The best we can do is use the stable pointer of the macro expression itself. However, I would prefer not to use a dummy pointer at all, which is why I chose not to use `expand_macro_for_statement` in this context.
</blockquote></details>

`expand_macro_for_statement` actually doesn't need that param - you can directly use the `st::StatementExpr` you took the tail from there.

just change `statements_and_tail`s return type to be `ast::StatementExpr` instead of the inner `ast::Expr` type.
___
*[`crates/cairo-lang-semantic/src/expr/compute.rs` line 713 at r4](https://reviewable.io/reviews/starkware-libs/cairo/7894#-OU4zUdC7dMnHvpUMOo3:-OU4zUdDCAAiTutxvT2a:bgx7c31) ([raw file](https://github.com/starkware-libs/cairo/blob/3a17a14b9fcdf2b9257f1838da2e516dd2398fc9/crates/cairo-lang-semantic/src/expr/compute.rs#L713)):*
> ```XML
>             }
>         },
>         None => None,
> ```

this function should probably not handle everything with options.



_Suggestion:_
```XML
        None => None,
        Some(tail_expr) => match compute_tail_semantic(ctx, Some(tail_expr)) {
            Ok(None) => None,
            Ok(Some(expr_id)) => {
                let expr = ctx.arenas.exprs[expr_id].clone();
                Some(ExprAndId { id: expr_id, expr })
            }
            Err(diag) => {
                let expr = wrap_maybe_with_missing(ctx, Err(diag), tail_expr.stable_ptr(ctx.db));
                Some(ExprAndId { id: ctx.arenas.exprs.alloc(expr.clone()), expr })
            }
        },
```
___
*[`crates/cairo-lang-semantic/src/expr/compute.rs` line 2178 at r4](https://reviewable.io/reviews/starkware-libs/cairo/7894#-OU4z4eq-L5b8u6C4jHG:-OU4z4eq-L5b8u6C4jHH:b-kgo0l7) ([raw file](https://github.com/starkware-libs/cairo/blob/3a17a14b9fcdf2b9257f1838da2e516dd2398fc9/crates/cairo-lang-semantic/src/expr/compute.rs#L2178)):*
> ```XML
>     ctx.arenas.exprs.alloc(Expr::Block(ExprBlock {
>         statements: statements_semantic,
>         tail: tail_semantic_expr.as_ref().map(|expr| expr.id),
> ```

why all the `as_ref` calls?



_Suggestion:_
```XML
        tail: tail_semantic_expr.map(|expr| expr.id),
```

dean-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/7894)* status: 0 of 2 files reviewed, 5 unresolved discussions (waiting on @orizi)

___
*[`crates/cairo-lang-semantic/src/expr/compute.rs` line 667 at r5](https://reviewable.io/reviews/starkware-libs/cairo/7894#-OU5Fhzs81HKn3pMqQmQ:-OU9zA8A2wJCH8gM3xqF:b-896fix) ([raw file](https://github.com/starkware-libs/cairo/blob/bfdc19afb8a319c673a70455081f0be80dceee65/crates/cairo-lang-semantic/src/expr/compute.rs#L667)):*

this just drops all the statements that are not last.

the function MUST get the `statement_ids` as an input from the outer scope.

have a test failing this.
</blockquote></details>

Done.
___
*[`crates/cairo-lang-semantic/src/expr/compute.rs` line 2004 at r5](https://reviewable.io/reviews/starkware-libs/cairo/7894#-OU5FH_71ZXgAi-xuoID:-OU9Gqwq53fXe6WBytay:bss4a2b) ([raw file](https://github.com/starkware-libs/cairo/blob/bfdc19afb8a319c673a70455081f0be80dceee65/crates/cairo-lang-semantic/src/expr/compute.rs#L2004)):*

useless doc.
make the function name state that instead.
</blockquote></details>

Removed.

orizi (approved):
:lgtm:

> Reviewed 2 of 2 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/7894#-:-OUAcuRDBNAeWr8Zs6vM:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @dean-starkware)

PR #7993, opened 2025-07-18
Return `Option` from `TextSpan::take`

integraledelebesgue:
## Rationale
Current implementation panics if the given span does not fit in the provided text.
Returning `Option` makes the method more explicit and safe and simplifies a lot of logic in CairoLS (text edits, macro expansions, etc.).

### DIFF
--- a/crates/cairo-lang-diagnostics/src/location_marks.rs
+++ b/crates/cairo-lang-diagnostics/src/location_marks.rs
@@ -53,7 +53,7 @@ fn get_single_line_location_marks(
     };
 
     let first_line_span = TextSpan { start: first_line_start, end: first_line_end };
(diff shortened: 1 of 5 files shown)


orizi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/7993)* status: 0 of 5 files reviewed, 1 unresolved discussion

___
*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/7993#-OVlyGMl9knnAmMBLLUd:-OVlyGMl9knnAmMBLLUe:b-jzqrlj) (no related file):*
as discussed - we should probably understand how these are actually produced before adding special handling for things that should never happen.

integraledelebesgue:
Fixed in LS, closing

PR #8127, opened 2025-08-08
fix(scripts): improve path matching and quoting in version bump and Sierra check scripts

fifalodm:
scripts/bump_version.sh:
- Corrected `find` command grouping by adding parentheses around `-o` expressions to ensure both file patterns are matched under `-type f`.
- This change prevents unintended matches and improves cross-platform compatibility.

scripts/sierra_update_check.sh:
- Fixed crate path patterns from `crate/...` to `crates/...` to align with the actual repository structure.
- Added missing backslashes for proper multi-line `grep -q -E` syntax.
- Quoted all variable references (`"$BASE_BRANCH"`, `"$HEAD_BRANCH"`, `"$MERGE_BASE"`) to avoid issues with spaces or special characters.

### DIFF
--- a/scripts/bump_version.sh
+++ b/scripts/bump_version.sh
@@ -49,5 +49,5 @@ check_scarb_version_sync() {
 check_scarb_version_sync
 
 sed -i "s/$CURRENT_VERSION/$NEW_VERSION/g" \
-    $(find . -type f -iname "*.toml"  -o -iname "Scarb.lock") \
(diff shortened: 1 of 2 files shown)


fifalodm:
@orizi 
I added parentheses so `-type f` checks apply to both patterns.  
Without them, it could match `Scarb.lock` items that aren't files.  
Now it only finds:
- files ending with `.toml`
- the file named `Scarb.lock`

PR #8128, opened 2025-08-08
Add APIs to pass external hint processor when running sierra casm in starknet context

mkaput:
#### Changes

This PR extends two crates:
1. `cairo-lang-runner`:
    - Expose a public entry point to run a Cairo program with a caller‑provided `HintProcessor`.
    - Allow injecting a custom `CairoRunner` instance and an initialization closure for VM setup.
    - Thread the external `HintProcessor` through the Starknet execution path for Sierra/CASM runs.
    - Minor refactors to make the execution pipeline accept externally created hint processors without changing existing defaults.

2. `cairo-lang-test-runner`:
    - Add support for a custom hint processor via an optional factory stored in the test runner.
    - Provide a builder-style API to register a custom hint processor for test execution.
    - Integrate the custom hint processor into the Starknet test execution flow (Sierra/CASM), while preserving existing behaviour when none is provided.

#### Motivation

This change allows Scarb to integrate oracles into `cairo-test`. Accompanying PR: https://github.com/software-mansion/scarb/pull/2513.

#### Alternatives

1. Scarb implements oracles in the `scarb-oracle-hint-service` crate. It is architected in such a way that it deliberately has no dependencies on any `cairo-lang-*` crates, nor `cairo-vm`. This design will make long-term maintenance much easier because upgrading Cairo in e.g. Starknet Foundry will not require waiting for `scarb-oracle-hint-service` upgrade. Technically, I could integrate `scarb-oracle-hint-service` directly into `cairo-lang-test-runner`, but this makes a dependency cycle between Cairo and Scarb repositories in case we need to upgrade oracles. With this approach, we avoid this problem. On the other hand, we would get oracles support in bare `cairo-test` binary -- but I doubt anyone needs this.
2. I could also do very heavy refactoring of `CairoHintProcessor` to allow providing extra custom cheatcode providers. Tried to do it, it is a very heavy refactoring, requiring complex type system shenanigans, that will break everyone using `cairo-lang-runner`. And the effort is dubious because Starknet Foundry would not be able to use this mechanism as they have their own complex system of building extensible hint processors.

### DIFF
--- a/crates/cairo-lang-test-runner/src/lib.rs
+++ b/crates/cairo-lang-test-runner/src/lib.rs
@@ -36,11 +36,15 @@ use rayon::prelude::{IntoParallelIterator, ParallelIterator};
 #[cfg(test)]
 mod test;
 
+type ArcCustomHintProcessorFactory = Arc<
+    dyn (for<'a> Fn(CairoHintProcessor<'a>) -> Box<dyn StarknetHintProcessor + 'a>) + Send + Sync,
+>;
+
 /// Compile and run tests.
 pub struct TestRunner<'db> {
     compiler: TestCompiler<'db>,
     config: TestRunConfig,
-    custom_hint_processor_factory: Option<Arc<dyn StarknetHintProcessorFactory>>,
+    custom_hint_processor_factory: Option<ArcCustomHintProcessorFactory>,
 }
 
 impl<'db> TestRunner<'db> {
@@ -81,7 +85,10 @@ impl<'db> TestRunner<'db> {
     /// Make this runner run tests using a custom hint processor.
     pub fn with_custom_hint_processor(
         &mut self,
-        factory: impl StarknetHintProcessorFactory + 'static,
+        factory: impl (for<'a> Fn(CairoHintProcessor<'a>) -> Box<dyn StarknetHintProcessor + 'a>)
+        + Send
+        + Sync
+        + 'static,
     ) -> &mut Self {
         self.custom_hint_processor_factory = Some(Arc::new(factory));
         self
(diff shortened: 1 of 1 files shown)


mkaput (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/8128)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @orizi)

___
*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/8128#-OXSOQZp8UKDwfQECdR8:-OXT8QcT4l6iF6xgrzKb:braobbu) (no related file):*

I think adding 2 traits make some more sense:
1. Hint-Processor-Factory.
2. Hint-Processor that additionally adds the features for counting resources to access through it.

and just use these 2 traits all around.
</blockquote></details>

Sorry, I cannot understand your idea.  There is a `HintProcessor` trait in `cairo-vm` so changing it seems like an undoable thing. Can you elaborate a bit more on what would you like to accomplish?

orizi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/8128)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @mkaput)

___
*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/8128#-OXSOQZp8UKDwfQECdR8:-OXTByjS6WgPtp6BPIqP:bhsitgs) (no related file):*

Sorry, I cannot understand your idea.  There is a `HintProcessor` trait in `cairo-vm` so changing it seems like an undoable thing. Can you elaborate a bit more on what would you like to accomplish?
</blockquote></details>

you did something that keeps requesting an extended interface from the hint-processor - `fn compose(` related.
you can have an extended trait defined here (not changing the original to use it) - and make our HintProcessor implement it.
Additioanlly - we can have a factory here returning it (and always use a factory Fn, not the option concept you added.)

mkaput:
This change is part of the following stack:

- #8128 ◀

<sub>Change managed by [git-spice](https://abhinav.github.io/git-spice/).</sub>

PR #8494, opened 2025-10-01
Make file_syntax_data function public

wawel37:
1. We make it public as we need it in the cairo-lint and LS.
2. We change the type to `dyn Database`, as we want to use this struct in LS, where we use the `Database` type from salsa already

### DIFF
--- a/crates/cairo-lang-parser/src/db.rs
+++ b/crates/cairo-lang-parser/src/db.rs
@@ -50,7 +50,7 @@ struct SyntaxData<'db> {
 
 /// Parses a file and returns the result and the generated [ParserDiagnostic].
 #[salsa::tracked(returns(ref))]
-fn file_syntax_data<'db>(db: &'db dyn Database, file_id: FileId<'db>) -> SyntaxData<'db> {
+pub fn file_syntax_data<'db>(db: &'db dyn Database, file_id: FileId<'db>) -> SyntaxData<'db> {
     let mut diagnostics = DiagnosticsBuilder::default();
     let syntax = db.file_content(file_id).to_maybe().map(|s| match file_id.kind(db) {
         FileKind::Module => Parser::parse_file(db, &mut diagnostics, file_id, s).as_syntax_node(),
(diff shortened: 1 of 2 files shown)


orizi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/8494)* status: 0 of 2 files reviewed, 1 unresolved discussion

___
*[`crates/cairo-lang-parser/src/db.rs` line 53 at r1](https://reviewable.io/reviews/starkware-libs/cairo/8494#-OaVsAa859e0AcbSomym:-OaVsAa859e0AcbSomyn:b1li8ra) ([raw file](https://github.com/starkware-libs/cairo/blob/f19356f20201c27a287eba4d52c2d0c0e5ce306f/crates/cairo-lang-parser/src/db.rs#L53)):*
> ```XML
> /// Parses a file and returns the result and the generated [ParserDiagnostic].
> #[salsa::tracked(returns(ref))]
> pub fn file_syntax_data<'db>(db: &'db dyn Database, file_id: FileId<'db>) -> SyntaxData<'db> {
> ```

you don't need it - you have all the members of `ParserGroup` as the selectors for this data struct.

orizi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/8494)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @wawel37)

___
*[`crates/cairo-lang-parser/src/db.rs` line 53 at r1](https://reviewable.io/reviews/starkware-libs/cairo/8494#-OaVsAa859e0AcbSomym:-Oa_KU2Y9e0DMjhJ11yW:b8ha263) ([raw file](https://github.com/starkware-libs/cairo/blob/f19356f20201c27a287eba4d52c2d0c0e5ce306f/crates/cairo-lang-parser/src/db.rs#L53)):*
<details><summary><i>Previously, wawel37 (Mateusz Kowalski) wrote…</i></summary><blockquote>

Ok, so cant we just make the `syntax_file` function part of the query group? As it's used in existing selectors anyway. And we don't want to duplicate logic of choosing right selector based on the `FileId` kind
</blockquote></details>

i still don't see WHY you need it  - what is the actual context?

PR #8522, opened 2025-10-11
fix: correct doc comment for function_inline_config

MozirDmitriy:
correct doc comment reference for function_inline_config


orizi:
outdated now - closing.

PR #8605, opened 2025-11-04
Chore: remove unused TopLevelLanguageElementId import

Doryu001:
Remove the unused TopLevelLanguageElementId import from cairo-size-profiler’s main entry point keep the binary’s dependency list lean and free of dead-import warnings


orizi:
No response. closing.

PR #8614, opened 2025-11-05
Remove unused TypedSyntaxNode import from cairo-size-profiler

Doryu001:
Drop the unused TypedSyntaxNode import from crates/bin/cairo-size-profiler/src/main.rs keep only the Terminal and ast imports that the module actually relies on


orizi:
No response. closing.

PR #8624, opened 2025-11-06
fix: align MAX_STACK_TRACE_DEPTH panic message

dmytroheknt:
Tweaked the panic text around MAX_STACK_TRACE_DEPTH so it names the env var we actually read. Keeps the guidance consistent when someone mistypes the value.


dmytroheknt:
@orizi fixed the rustfmt CI error.

PR #8627, opened 2025-11-07
fix: correct fixed array CLI help text

ruslan0012:
The help message for `--fixed-array-line-breaking` said the default kept each array element on its own line. In reality, `SingleBreakPoint` is the default and it tries to keep items on one line. I updated the comment to match the actual behavior so users aren’t misled when reading the CLI help.

### DIFF
--- a/crates/bin/cairo-format/src/main.rs
+++ b/crates/bin/cairo-format/src/main.rs
@@ -47,9 +47,9 @@ struct FormatterArgs {
     /// same line (as space permits). Defaults to line-by-line.
     #[arg(long)]
     tuple_line_breaking: Option<bool>,
-    /// Controls fixed array breaking behavior. Set to 'single-break-point' (default) to format
-    /// each array item on a new line, or 'line-by-line' to keep as many items as possible on the
-    /// same line (as space permits). Defaults to single line.
+    /// Controls fixed array breaking behavior. Set to 'single-break-point' (default) to keep as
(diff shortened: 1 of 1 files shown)


orizi (changes requested):
> @orizi reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/8627)* status: 0 of 1 files reviewed, 1 unresolved discussion

___
*[`crates/bin/cairo-format/src/main.rs` line 51 at r1](https://reviewable.io/reviews/starkware-libs/cairo/8627#-OdSgCEIBZJzB_dtVkpy:-OdSgCEIBZJzB_dtVkpz:b-lwgen0) ([raw file](https://github.com/starkware-libs/cairo/blob/1925936cfbe39194d1a76ce7f206c2abf139f42f/crates/bin/cairo-format/src/main.rs#L51)):*
> ```XML
>     tuple_line_breaking: Option<bool>,
>     /// Controls fixed array breaking behavior. Set to 'single-break-point' (default) to keep as
>     /// many items as possible on the same line, or 'line-by-line' to format each array item on a
> ```

you flipped the meaning.

ruslan0012:
@orizi thanks! I double-checked the formatter code: the default `single-break-point` setting tries to keep items on one line, while `line-by-line` puts each item on its own line. The help text had those reversed, so I swapped the descriptions to match the actual behavior.

PR #8630, opened 2025-11-07
fix: rename test parameter from `modules` to `modulus` in circuit_test

deus-labs:
Fixed a parameter naming inconsistency in the test_invert_or_nullify function. The parameter was incorrectly named `modules` (plural, suggesting program modules) but should be `modulus` (singular, the mathematical term used in modular arithmetic).

This matches the original invert_or_nullify function signature which uses `modulus` as the parameter name, and is semantically correct since we're doing modular inversion where the second argument is the modulus for the mod operation.

### DIFF
--- a/crates/cairo-lang-runner/src/casm_run/circuit_test.rs
+++ b/crates/cairo-lang-runner/src/casm_run/circuit_test.rs
@@ -53,8 +53,8 @@ use crate::casm_run::circuit::invert_or_nullify;
     "invert_or_nullify(7, 8)"
 )]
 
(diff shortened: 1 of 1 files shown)


orizi (approved):
:lgtm:

> @orizi reviewed 1 of 1 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/8630#-:-OdYxbQ-E0sS1GnMRtCv:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @deus-labs)

PR #8647, opened 2025-11-11
Chore: remove unused rand::Rng import from CASM runner

Doryu001:
remove the unused rand::Rng import from crates/cairo-lang-runner/src/casm_run/mod.rs silence the unused import warning and keep the runner dependency surface minimal


orizi:
No response. closing.

PR #8650, opened 2025-11-12
fix: node ID mismatch in test_root_points_to_cycle documentation

sewqasreedas:
Fixed copy-paste error in test_root_points_to_cycle comment. 
Comment said `Note 10 is used as a root` but code actually uses node 0.

Originally tried fixing the code to match the comment (changing id from 0 to 10),but tests failed. Turns out the comment was wrong, not the code.The comment was copied from strongly_connected_components_test.rs but wasn't updated when the test logic was adapted for feedback_set.

### DIFF
--- a/crates/cairo-lang-utils/src/graph_algos/feedback_set_test.rs
+++ b/crates/cairo-lang-utils/src/graph_algos/feedback_set_test.rs
@@ -61,8 +61,8 @@ fn test_root_points_to_cycle() {
     // And 10 (the root) has and edge to 0.
     graph.push(/* 10: */ vec![0]);
 
-    // Note 10 is used as a root (which is part of the cycle).
-    let fset = HashSet::<usize>::from_iter(calc_feedback_set(IntegerNode { id: 10, graph }.into()));
+    // Note 0 is used as a root (which is part of the cycle).
+    let fset = HashSet::<usize>::from_iter(calc_feedback_set(IntegerNode { id: 0, graph }.into()));
     assert_eq!(fset, HashSet::from([0]));
 }
 


orizi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/8650)* status: 0 of 1 files reviewed, 1 unresolved discussion

___
*[`crates/cairo-lang-utils/src/graph_algos/feedback_set_test.rs` line 65 at r1](https://reviewable.io/reviews/starkware-libs/cairo/8650#-OdseCNyAzEQ2bff85jp:-OdseCNzCz4cW7-J2AdE:undefined) ([raw file](https://github.com/starkware-libs/cairo/blob/fdbfd24ea29bcd1d176d034417587b3b27095ca5/crates/cairo-lang-utils/src/graph_algos/feedback_set_test.rs#L65)):*
> ```XML
> 
>     // Note 0 is used as a root (which is part of the cycle).
>     let fset = HashSet::<usize>::from_iter(calc_feedback_set(IntegerNode { id: 0, graph }.into()));
> ```





_Suggestion:_
```XML
    // Note 10 is used as a root (which is part of the cycle).
    let fset = HashSet::<usize>::from_iter(calc_feedback_set(IntegerNode { id: 10, graph }.into()));
```

orizi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/8650)* status: 0 of 1 files reviewed, 1 unresolved discussion

___
*[`crates/cairo-lang-utils/src/graph_algos/feedback_set_test.rs` line 64 at r2](https://reviewable.io/reviews/starkware-libs/cairo/8650#-Odw0jB16VplDBQAHAXY:-Odw0jB16VplDBQAHAXZ:binfkcy) ([raw file](https://github.com/starkware-libs/cairo/blob/08ce0213e26d3683403bd6037fd50c03ed16ef41/crates/cairo-lang-utils/src/graph_algos/feedback_set_test.rs#L64)):*
> ```XML
>     graph.push(/* 10: */ vec![0]);
> 
>     // Note 10 is used as a root (which is part of the cycle).
> ```

it isn't part of the cycle - it should only *point* to the cycle.


the graph is: 
0 -> 1 -> 2 -> 3 -> 4 -> 5 -> 6 -> 7 -> 8 -> 9 -> 0
with also 10 -> 0.

the idea is that we should find `0` as a feedback set, if the root we use leads to the cycle.

in any case - it does not seem to me you understand what actually happens here.

sewqasreedas:
@orizi as mentioned in the PR description, changing to id: 10 breaks the test - it returns {} instead of {0}. Node 10 isn't in the cycle's SCC, so calc_feedback_set doesn't find anything. Kept the original fix (just the comment change).

sewqasreedas:
got it, will close

PR #8665, opened 2025-11-13
 chore: remove unused cairo-lang imports from ensure-no_std harness

conomist:
drop the unused cairo_lang_casm and cairo_lang_utils imports from the no-std entry point
keep the minimal no-std bootstrap free of unnecessary dependencies


orizi:
No response. closing.

PR #8669, opened 2025-11-14
Remove redundant clone in GenStatement::map

soniseth0:
eliminate an unnecessary Vec<VarId> clone when mapping GenStatement::Return, keeping the original allocation intact retain existing behavior because self is consumed and we still return the owned results vector reduces avoidable allocation/copy in the hot path


orizi:
No response. closing.

PR #8670, opened 2025-11-14
Remove unused BigInteger import from casm_run::mod

Doryu001:
drop the unused BigInteger import from cairo-lang-runner’s casm_run::mod keep the remaining PrimeField import intact to satisfy the elliptic-curve modulus checks


orizi:
No response. closing.

PR #8693, opened 2025-11-18
fix: clarify profiler option doc comment

dmytroheknt:
Aligned the clap doc comment with the actual `Option<ProfilerConfig>` type so the CLI help no longer implies a slice is returned.


dmytroheknt:
It’s wired into the `--run-profiler` flag in both `cairo-test` and `cairo-execute` (see `crates/bin/cairo-test/src/main.rs` and `crates/bin/cairo-execute/src/main.rs`).

PR #8699, opened 2025-11-18
refactor(runner): remove incorrect dead_code suppression from DictTrackerExecScope

conomist:
Removes an incorrect #[allow(dead_code)] attribute from the idx field in DictTrackerExecScope. This field is actively used and the suppression attribute was misleading.While reviewing code quality attributes, identified that DictTrackerExecScope::idx is incorrectly marked with #[allow(dead_code)]. The field is:Initialized in the constructor (new(), L25)Publicly accessed via get_dict_infos_index() (L72).This false suppression masks legitimate compiler warnings and misleads contributors about the field's usage.


orizi:
bad PR - no response. closing.

PR #8700, opened 2025-11-19
Remove redundant clone in GasReserveCreateLibfunc

deus-labs:
Removed unnecessary `.clone()` call on `gas_reserve_type` in the success branchof `GasReserveCreateLibfunc::specialize_signature()`. The variable is moved and not used afterwards, making the clone operation redundant.


deus-labs:
I tested locally — compiles fine and all tests pass. `gas_reserve_type` is used only once (line 68), so the clone is redundant.
could you double-check?

PR #8765, opened 2025-11-26
fix(corelib): correct append_byte documentation example

sewqasreedas:
The doc example for `append_byte` was using `append_byte(0)` and asserting equality with `"0"`, which would never be true.

`append_byte(0)` adds a NULL byte (0x00), not the character '0' (0x30). The assert would always fail since the byte array would contain a null character, not the string "0".

Fixed by using the char literal `'A'` which is more readable and idiomatic.

### DIFF
--- a/corelib/src/byte_array.cairo
+++ b/corelib/src/byte_array.cairo
@@ -178,8 +178,8 @@ pub impl ByteArrayImpl of ByteArrayTrait {
     ///
     /// ```
     /// let mut ba = "";
-    /// ba.append_byte(0);
-    /// assert!(ba == "0");
+    /// ba.append_byte('A');
+    /// assert!(ba == "A");
     /// ```
     fn append_byte(ref self: ByteArray, byte: u8) {
         if self.pending_word_len == 0 {


orizi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/8765)* status: 0 of 1 files reviewed, 1 unresolved discussion

___
*[`corelib/src/byte_array.cairo` line 182 at r1](https://reviewable.io/reviews/starkware-libs/cairo/8765#-Of0I6FJ6Vw9MxThzpUk:-Of0I6FJ6Vw9MxThzpUl:undefined) ([raw file](https://github.com/starkware-libs/cairo/blob/b2fbbb6eb940957ef1ec20ff7cd31bd587266ba2/corelib/src/byte_array.cairo#L182)):*
> ```Cairo
>     /// let mut ba = "";
>     /// ba.append_byte(0x41); // Appending a single byte 'A'
>     /// assert!(ba == "A");
> ```





_Suggestion:_
```Cairo
    /// let mut ba = "";
    /// ba.append_byte('A');
    /// assert!(ba == "A");
```

PR #8774, opened 2025-11-26
 avoid redundant clone when recording function costs

conomist:
drop the unnecessary to_string() allocation in run_e2e_test
reuse the existing String when inserting "function_costs" into the result map

### DIFF
--- a/tests/e2e_test.rs
+++ b/tests/e2e_test.rs
@@ -296,7 +296,7 @@ fn run_e2e_test(
             .iter()
             .map(|(func_id, cost)| format!("{func_id}: {cost:?}"))
             .join("\n");
-        res.insert("function_costs".into(), function_costs_str.to_string());
+        res.insert("function_costs".into(), function_costs_str);
     }
 
     TestRunnerResult::success(res)


orizi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/8774)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @conomist)

___
*[`tests/e2e_test.rs` line 299 at r2](https://reviewable.io/reviews/starkware-libs/cairo/8774#-OfDfCcj2cBDYU0mYOtd:-OfDfCcj2cBDYU0mYOte:bfybr7f) ([raw file](https://github.com/starkware-libs/cairo/blob/05f1b2b41152c4cb07bf1b5ad590aeb32830f49e/tests/e2e_test.rs#L299)):*
> ```XML
>             .map(|(func_id, cost)| format!("{func_id}: {cost:?}"))
>             .join("\n");
>         res.insert("function_costs".into(), function_costs_str.to_string());
> ```

this seems to be adding a to_string.

PR #8775, opened 2025-11-26
Avoid redundant cloning in SCC context updates

justinbay:
update compute_scc_recursive to refresh lowlink in-place via get_mut, remove the per-iteration clone of the wrapper node and node id


orizi:
No response. Closing.

PR #8779, opened 2025-11-27
 Remove unnecessary clone operations

ruslan0012:
Eliminates redundant clone operations across the codebase. The removed clones were not contributing any value and were creating unnecessary allocations. This improves performance slightly and makes the code cleaner.


ruslan0012:
tests passed successfully

PR #8788, opened 2025-11-28
Avoid redundant Sierra program clone in cairo-lang-test-plugin

conomist:
take sierra_program as mut directly from SierraProgramWithDebug
drop the extra clone() and reuse the original buffer when enriching debug info


orizi:
Closing PR - as not relevant and doesn't compile.

PR #8793, opened 2025-11-30
Enforce only type or const generic parameters for negative impls.

ilyalesokhin-starkware:


orizi (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/8793)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @ilyalesokhin-starkware and @TomerStarkware)

___
*[`crates/cairo-lang-semantic/src/expr/test_data/neg_impl` line 305 at r1](https://reviewable.io/reviews/starkware-libs/cairo/8793#-OfKAfaA6J4hUBkK3zfw:-OfKD3in750HsHyKQn9N:bd9uhhb) ([raw file](https://github.com/starkware-libs/cairo/blob/e9c077e6986847e52a900dc384446c8da3ef6334/crates/cairo-lang-semantic/src/expr/test_data/neg_impl#L305)):*
<details><summary><i>Previously, ilyalesokhin-starkware wrote…</i></summary><blockquote>

does this make sense?

Do we want to an error for when an impl is used as a paramater of a negative impl?
</blockquote></details>

do you mean - when it is unused? it seems strange to me.

unused happens with `+` as well?

PR #8805, opened 2025-12-01
perf(semantic): remove unnecessary clones in mapper rewriters

SashaMalysehko:
Replaced struct update clones with moves in Mapper::internal_rewrite() and Mapper::internal_rewrite() in canonic.rs. long() returns owned ImplVar/NegativeImplVar, so using ..var is safe and avoids redundant allocations. No behavior change; reduces copies and slightly improves performance.


orizi:
Code does not compile.
Closing PR.

PR #8811, opened 2025-12-02
fix variable name conflict in cairo

brawncode:
ran into an issue where a variable was declared twice within the same scope.
unlike some languages, Cairo (similar to Rust) doesn’t allow redeclaring a name - the old one isn’t shadowed.
to resolve this, I renamed the second variable so the code compiles cleanly and behaves as intended.


orizi:
This is completely untrue. (not even the part about rust).
Closing.

PR #8838, opened 2025-12-02
docs: fix invalid Rust nightly date

defitricks:
updated the command to use the correct form `rustup install nightly` for actual Rust release behavior.


orizi:
that is the version of nightly we are using.

closing PR.

PR #8841, opened 2025-12-02
fix: invalid `if let` syntax in `Goto` handling

zkpepe:
noticed that the existing conditional combined `if let` with a logical `&&`, which Rust doesn’t allow on the right-hand side of an `if let`. this caused a syntax error during compilation. i split the logic into two sequential `if` statements to preserve the intended behavior while keeping the code valid.

updated version now properly checks the `Introduction::Remapping` case and returns the appropriate `ValidationError` when needed.


orizi:
PR submitter obviously didn't try to compile the code.

the repository compile successfully.

Closing PR.

PR #8884, opened 2025-12-04
fix(starknet-plugin): remove unused declaration rewrite in entry_point handle

SashaMalysehko:
Removed the unused declaration_node rewrite in handle_entry_point() that attempted to strip mut from parameters. The mutated node was never used in any generated output and only added extra allocations and complexity. The entry point wrapper is generated independently from the AST, and the exposed symbol is a pub use alias, so this rewrite had no effect


orizi (approved):
> @orizi reviewed 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/8884)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @SashaMalysehko)

PR #8953, opened 2025-12-08
refactor: drop unnecessary clone in lifetime analysis

strmfos:
`demand` is already owned from the iterator, no need to clone before mutating


strmfos:
thanks

PR #8966, opened 2025-12-08
fix: refine wording in never-type documentation

prestoalvarez:
Clarify phrasing about evaluation context for the `never` type.

### DIFF
--- a/docs/reference/src/components/cairo/modules/language_constructs/pages/never-type.adoc
+++ b/docs/reference/src/components/cairo/modules/language_constructs/pages/never-type.adoc
@@ -1,4 +1,4 @@
 = Never type
 
-A _never type_ is a type of an expression that should not be evaluated in the encapsulating block,
+A _never type_ is a type of an expression that should not be evaluated within the enclosing block,
 but should cause a flow control change - this may be caused by `return` or `break` statements.


prestoalvarez:
> > _[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/8966)_ status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @prestoalvarez)
> 
> _[`docs/reference/src/components/cairo/modules/language_constructs/pages/never-type.adoc` line 3 at r1](https://reviewable.io/reviews/starkware-libs/cairo/8966#-OfxxKdv8WFujmwHqz8O:-OfxxKdv8WFujmwHqz8P:bq50p73) ([raw file](https://github.com/starkware-libs/cairo/blob/b6101536346304ff81630da3dd79235784c7c825/docs/reference/src/components/cairo/modules/language_constructs/pages/never-type.adoc#L3)):_
> 
> > ```
> > = Never type
> > 
> > A _never type_ is a type of an expression that should not be evaluated within the enclosing block,
> > ```
> 
> why is that better?

@orizi I rephrased to clarify the semantics: never marks expressions that change control flow (return/break/panic) so their value is never produced, rather than suggesting the expression itself is forbidden.

PR #9004, opened 2025-12-09
chore: remove unused _name param from  create_impl_generics

Galoretka:
This function didn’t use the name parameter and it was explicitly prefixed with _, indicating it was dead weight. The function is private and only used in this module, so changing the signature is safe. Updated the two call sites to pass just generics


orizi:
No response. Closing.

PR #9029, opened 2025-12-10
chore: minor improvements

PivasDesant:


PivasDesant:
@orizi sorry boss, there was a problem with fmt, now should be fine

PR #9030, opened 2025-12-10
Eliminate redundant clones in MergeCollection implementations

0xxFloki:
drop the extra Clone constraint from both OrderedHashMap and SmallOrderedMap MergeCollection impls, reuse occupied entry values via mem::replace so merge logic no longer allocates/clones unnecessarily, keep zero-handling behavior unchanged while reducing per-merge overhead


0xxFloki:
done

orizi:
No response. Closing.

PR #9031, opened 2025-12-10
chore: Align executable compile docs with current API

sashass1315:
Update the documentation for the executable compilation helpers to match their current signatures and behavior. The previous comments were leftovers from an older API: prepare_db was documented as if it compiled a function by path, while it actually only constructs a RootDatabase configured for #[executable] functions, and compile_executable_function_in_prepared_db still mentioned a diagnostics_reporter argument, described its config parameter as a boolean, and claimed to return Ok(Vec<String>). These inaccuracies made the docs misleading for users of the public API, so the comments were rewritten to describe the real responsibilities, arguments, and return types, and to clarify that diagnostics must be ensured by the caller before invoking the low-level helper.

### DIFF
--- a/crates/cairo-lang-executable/src/compile.rs
+++ b/crates/cairo-lang-executable/src/compile.rs
@@ -77,8 +77,10 @@ pub struct CompileExecutableResult<'db> {
     pub debug_info: SierraProgramDebugInfo<'db>,
 }
 
-/// Compile the function given by path.
-/// Errors if there is ambiguity.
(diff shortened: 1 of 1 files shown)


orizi (approved):
:lgtm:

> @orizi reviewed 1 of 1 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/9031#-:-Og9IUAzBpm7D_docfyO:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @sashass1315)

PR #9039, opened 2025-12-11
fix: avoid repeated allocation of unknown cairo function name

kurahin:
Use a single preallocated "unknown" String as the fallback function identifier in process_cairo_function_weights instead of calling unwrap_or(&"unknown".to_string()). unwrap_or evaluates its argument eagerly, which previously caused an extra heap allocation on every iteration even when a real function name was present. Reusing one String preserves the existing behavior while removing unnecessary allocations in the profiling hot path.

### DIFF
--- a/crates/cairo-lang-runner/src/profiling.rs
+++ b/crates/cairo-lang-runner/src/profiling.rs
@@ -655,14 +655,14 @@ impl<'a> ProfilingInfoProcessor<'a> {
         require(params.process_by_cairo_function)?;
 
         let mut cairo_functions = UnorderedHashMap::<_, _>::default();
-        let unknown_function_identifier = "unknown".to_string();
         for (statement_idx, weight) in sierra_statement_weights {
             // TODO(Gil): Fill all the `Unknown functions` in the Cairo functions profiling.
             let function_identifier = self
                 .statements_functions
                 .get(statement_idx)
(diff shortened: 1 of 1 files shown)


orizi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/9039)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @kurahin)

___
*[`crates/cairo-lang-runner/src/profiling.rs` line 665 at r1](https://reviewable.io/reviews/starkware-libs/cairo/9039#-OgChw-a9Selkk-C5I_T:-OgChw-b8MFrpn5d5aV6:undefined) ([raw file](https://github.com/starkware-libs/cairo/blob/37b659a3fb3e9ac09b4a0c804f609655d368e049/crates/cairo-lang-runner/src/profiling.rs#L665)):*
> ```XML
>                 .get(statement_idx)
>                 .unwrap_or(&unknown_function_identifier)
>                 .clone();
> ```





_Suggestion:_
```XML
        for (statement_idx, weight) in sierra_statement_weights {
            // TODO(Gil): Fill all the `Unknown functions` in the Cairo functions profiling.
            let function_identifier = self
                .statements_functions
                .get(statement_idx)
                .cloned()
                .unwrap_or_else(||"unknown".to_string());
```

PR #9041, opened 2025-12-11
docs: fix expect_diagnostics_input comment

MozirDmitriy:
Update the documentation for expect_diagnostics_input to reflect that it parses the expect_diagnostics argument into ExpectDiagnostics rather than a boolean. The new comment documents the accepted values (true, false, warnings_only) and clarifies the behavior, avoiding confusion for readers and test authors.


orizi:
No response. Closing.

PR #9054, opened 2025-12-12
docs: fix ref argument shorthand documentation

JoshMa-mq:
Clarify ref argument shorthand syntax example, make the explanation consistent with the pattern shown for regular arguments.


JoshMa-mq:
@orizi ,did it, sir.

PR #9067, opened 2025-12-12
docs(eq-solver): fix const expression docs in Expr

radik878:
The constructor doc comments used the term “cost expression”, which is misleading in this context and can be confused with cost modeling. Updated the wording to “const expression” to accurately describe what these helpers construct.

### DIFF
--- a/crates/cairo-lang-eq-solver/src/expr.rs
+++ b/crates/cairo-lang-eq-solver/src/expr.rs
@@ -17,12 +17,12 @@ pub struct Expr<Var: Clone + Debug + Eq + Hash> {
     pub var_to_coef: OrderedHashMap<Var, i64>,
 }
 impl<Var: Clone + Debug + Eq + Hash> Expr<Var> {
-    /// Creates a cost expression based on const value only.
+    /// Creates a const expression based on const value only.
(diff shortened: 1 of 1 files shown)


orizi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/9067)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @radik878)

___
*[`crates/cairo-lang-eq-solver/src/expr.rs` line 21 at r1](https://reviewable.io/reviews/starkware-libs/cairo/9067#-OgISqBFDkMFJQdSANIQ:-OgISqBFDkMFJQdSANIR:bo7zqfz) ([raw file](https://github.com/starkware-libs/cairo/blob/f59d66cab218a49d30d1ee4e1b2ecd9895e4bafd/crates/cairo-lang-eq-solver/src/expr.rs#L21)):*
> ```XML
> impl<Var: Clone + Debug + Eq + Hash> Expr<Var> {
>     /// Creates a const expression based on const value only.
>     pub fn from_const(const_term: i32) -> Self {
> ```

original was correct.
close PR.

PR #9069, opened 2025-12-12
Cache path string once in setup_single_file_project

0xlupin:
store path.to_string_lossy().into_owned() once and reuse it for all error cases, avoid repeated allocations while keeping error messages unchanged


orizi:
No response. Closing.

PR #9106, opened 2025-12-15
Fix (docs) AST specification wording

atakoraka:
this refers to one spec as a concept, so it’s now *AST specification* instead of *AST specifications*.


atakoraka:
[orizi](https://github.com/orizi) pls check

orizi:
No response. Closing.

PR #9116, opened 2025-12-15
docs: add `snapshot-type.adoc`

bobtajson:
Added `snapshot-type.adoc` file


bobtajson:
@orizi changed sir, thank you

PR #9184, opened 2025-12-18
fix: correct value range in sqrt comment

erkidonik:
Fix comment in `build_sqrt` function. The comment incorrectly stated `value is in [0, 2**250)`, but this function handles u8/u16/u32/u64/u128 types where max value is 2^128. Changed to `[0, 2**128)`.


orizi:
No response.

Closing.

PR #9198, opened 2025-12-18
refactor: remove redundant #[allow(dead_code)] in bigint schema helpers

marukai67:
Removes unnecessary `#[allow(dead_code)]` attributes from helper types in the `big_int_schema` function.


orizi:
PR was obviously wrong as you can see from CI.
Closing.

PR #9231, opened 2025-12-19
Fix (docs): grammar issue in comments

indmind:


indmind:
[orizi](https://github.com/orizi) corrected

PR #9250, opened 2025-12-21
Complete impl parameter usage examples

jschnelder:
Adds a missing example showing how to use TDisplay trait directly.

The documentation already shows calling `.display()` as a method, but doesn't demonstrate the trait usage pattern `TDisplay::display(value)`.


orizi:
No response.

Closing.

PR #9269, opened 2025-12-21
docs: correct numeric match threshold flag comment

radik878:
## Summary

Fixed reference in comments

---

## Type of change

Please check **one**:

---

## Why is this change needed?

- A behavior was undocumented or unclear in a way that caused real confusion

---

Docs incorrectly referenced `unsafe_panic` instead of  `numeric_match_optimization_min_arms_threshold`

Docs now reference `numeric_match_optimization_min_arms_threshold` as it should be

---


radik878:
But I still didn't see template when opened PR, so i copied it from .github/PULL_REQUEST_TEMPLATE/pull_request_template.md

PR #9273, opened 2025-12-21
perf: remove redundant clone in test compilation pipeline

0xlupin:
Removed unnecessary .clone() call on sierra_program in compile_test_prepared_db()


orizi:
This does not compile.

Closing PR.

PR #9274, opened 2025-12-21
perf(bigint): eliminate redundant clones in serialize_big_ints

0xxFloki:
The `serialize_big_ints` function was unnecessarily cloning each `BigInt` when serializing, wrapping each element in `BigIntAsHex { value: num.clone() }`.This introduced O(n) redundant allocations proportional to the vector size and BigInt magnitude.


orizi:
`./scripts/clippy.sh` does not pass successfully.

and a duplicate of another PR.

Closing.

PR #9277, opened 2025-12-22
fix(defs): remove stale todo

Forostovec:
## Summary

Previously interpolate_patched would panic if a placeholder name was not found in the patches map, as noted by the existing TODO. This meant that any mismatch between the template string and the patch keys could crash the compiler instead of producing a recoverable error in generated code.  

---

## Type of change

---

## Why is this change needed?

To treat unknown placeholders $name$ as a literal substring, emitting "$name$" into the output rather than panicking.

---

### DIFF
--- a/crates/cairo-lang-defs/src/patcher.rs
+++ b/crates/cairo-lang-defs/src/patcher.rs
@@ -188,11 +188,9 @@ impl<'db> RewriteNode<'db> {
                 pending_text.clear();
             }
             // Replace the substring with the relevant rewrite node.
-            if let Some(node) = patches.get(&name).cloned() {
-                children.push(node);
-            } else {
-                children.push(RewriteNode::text(&format!("${}$", name)));
-            }
+            children.push(
+                patches.get(&name).cloned().unwrap_or_else(|| panic!("No patch named {name}.")),
+            );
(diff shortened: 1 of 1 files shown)


Forostovec:
> > @orizi made 1 comment.
> > _[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/9277)_ status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @Forostovec).
> 
> _[`crates/cairo-lang-defs/src/patcher.rs` line 191 at r1](https://reviewable.io/reviews/starkware-libs/cairo/9277#-Oh4Hby1CPZfDLpyAAkt:-Oh4Hby1CPZfDLpyAAku:bo7i408) ([raw file](https://github.com/starkware-libs/cairo/blob/646106c442fc50a43e8ba5ae0c7a22566c8904cf/crates/cairo-lang-defs/src/patcher.rs#L191)):_
> 
> > ```
> >             }
> >             // Replace the substring with the relevant rewrite node.
> >             if let Some(node) = patches.get(&name).cloned() {
> > ```
> 
> panic is completely preferable.
> 
> if you wish - you can remove the TODO.

Did it

PR #9305, opened 2025-12-23
perf: avoid heap allocation for syntax node key fields

yyhrnk:
## Summary

SyntaxNodeId::Child no longer stores key_fields as a boxed slice. Instead, key fields are derived on demand from the underlying GreenNode children using key_fields_range.

---

## Type of change

Please check **one**:

- [x] Performance improvement

---

## Why is this change needed?

This removes a per-node heap allocation and copy while keeping the stable pointer semantics and cache format intact by computing key fields lazily in both the syntax tree and the defs cache.

### DIFF
--- a/crates/cairo-lang-defs/src/cache/mod.rs
+++ b/crates/cairo-lang-defs/src/cache/mod.rs
@@ -1642,12 +1642,15 @@ impl SyntaxNodeIdCached {
                 let green = syntax_node.green_node(ctx.db).clone().intern(ctx.db);
                 Self::Root(FileIdCached::new(*file_id, ctx), GreenIdCached::new(green, ctx))
             }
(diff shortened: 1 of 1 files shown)


eytan-starkware (changes requested):
The SyntaxNodeId is the unique identifier for the SyntaxNode tracked struct. Removing the key_fields will prevent salsa from tracking the changes correctly in the LS leading to unneeded computations.

> @eytan-starkware made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/9305#-:-OhJgnXO231ttyBPlXcd:b-174toc)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @orizi and @yyhrnk).

PR #9307, opened 2025-12-23
perf: avoid redundant String clones in doc parser

Bashmunta:
## Summary

This change removes unnecessary String allocations when parsing documentation links. The link resolver now accepts &str instead of owning a String, and the CommentLinkToken Display implementation no longer clones its path field.

---

## Type of change

Please check **one**:

- [x] Performance improvement

---

## Why is this change needed?

As a result, we reuse a single String instance per link and pass references into the resolver, keeping the behavior unchanged while reducing allocation overhead in the documentation parser.

---

### DIFF
--- a/crates/cairo-lang-doc/src/parser.rs
+++ b/crates/cairo-lang-doc/src/parser.rs
@@ -211,17 +211,16 @@ impl<'db> DocumentationCommentParser<'db> {
                                     current_link = Some(CommentLinkToken {
                                         label: "".to_string(),
                                         path: None,
-                                        resolved_item: self.resolve_linked_item(item_id, path), /* Or resolve item here */
+                                        resolved_item: self.resolve_linked_item(item_id, &path), /* Or resolve item here */
                                     });
                                 }
                                 _ => {
+                                    let path = dest_url.clone().into_string();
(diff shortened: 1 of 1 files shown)


orizi (changes requested):
> @orizi made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/9307)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @Bashmunta).

___
*[`crates/cairo-lang-doc/src/parser.rs` line 219 at r1](https://reviewable.io/reviews/starkware-libs/cairo/9307#-OhCTuhk2ZFXqWK0v7RM:-OhCTSCn9KiyF4QZI1PT:bv4wd25) ([raw file](https://github.com/starkware-libs/cairo/blob/9a9e4558b6ee3bd7dbe41a4c9c3a1b1e4e10b48e/crates/cairo-lang-doc/src/parser.rs#L219)):*
> ```XML
>                                 _ => {
>                                     let path = dest_url.clone().into_string();
>                                     let resolved_item = self.resolve_linked_item(item_id, &path);
> ```

this still alocates the string in the SmolStrId::from - so not really any better.

PR #9311, opened 2025-12-24
docs: clarify prelude auto-import behavior

Princetimix69:
## Summary

Clarifies that prelude cannot be disabled means items are always in scope, explaining why types like `Option` and `Result` don't require explicit imports.

---

## Type of change

Please check **one**:

- [ ] New feature
- [ ] Performance improvement

---

## Why is this change needed?

The original documentation states "It is not possible to disable standard library prelude importing" but doesn't explain what this means in practice. New Cairo developers frequently wonder why certain types like `Option`, `Result`, and common traits are available without explicit imports, leading to confusion about the language's import system.

---

The prelude documentation stated the fact but didn't connect it to the practical consequence:
- Said prelude importing cannot be disabled
- No explanation of what "always imported" means for everyday code
- No examples of which types benefit from this behavior

---

The documentation now explains:
- Prelude items are always in scope without explicit imports
- Provides concrete examples: `Option`, `Result`, and common traits
- Directly connects the "cannot disable" restriction to the "always available" benefit

---

N/A

---

## Additional context

N/A

### DIFF
--- a/docs/reference/src/components/cairo/modules/language_constructs/pages/prelude.adoc
+++ b/docs/reference/src/components/cairo/modules/language_constructs/pages/prelude.adoc
@@ -2,8 +2,9 @@
 
 A _prelude_ is a collection of items that are always available in every module in a program.
 
-The _standard library prelude_ is a module defined in the standard library: `std::prelude::v1`.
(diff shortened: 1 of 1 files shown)


Princetimix69:
@orizi , Fixed. Pls check, thanks!

PR #9340, opened 2025-12-25
docs: document panic_with_felt252 in const evaluation

radik878:
## Summary

The const evaluation docs did not mention core::panic_with_felt252 as an allowed function call, even though the implementation treats it as const-eligible via is_function_const and always produces FailedConstantCalculation at runtime.

---

## Type of change

Please check **one**:

---

## Why is this change needed?

This change documents that special case so that the language semantics reference matches the current compiler behavior and makes the error behavior of const panics explicit.

---


orizi (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/9340#-:-OhKx36T6Pguw0ilSlBT:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @radik878).

PR #9348, opened 2025-12-26
refactor: convert secp256 curve size to u256 constants

kseniaeremekno:
## Summary

This PR resolves TODOs in corelib/src/starknet/secp256k1.cairo and secp256r1.cairo by converting the get_curve_size() functions to return u256 constants instead of inline literals.

---

## Type of change

Please check **one**:

- [ ] New feature
- [ ] Performance improvement


orizi:
That was not the meaning of that TODO - if done, it would be internal.

close PR.

PR #9350, opened 2025-12-26
docs: complete generics PartialEq example in equality operators

avorylli:
## Summary

Adds a usage example to the "Generics requiring PartialEq" section. The example now includes a `main` function demonstrating how to call the generic `contains` function, making the code example complete and compilable.

---

## Type of change

Please check **one**:

- [ ] New feature
- [ ] Performance improvement

> ⚠️ Note:

---

## Why is this change needed?

The example showed only a function definition without demonstrating how to use it. This made the example incomplete, non-executable, and unclear for users trying to understand how generic functions with trait bounds work in practice.

---

The section contained only:

fn contains<T, +PartialEq<T>>(a: @T, b: @T) -> bool {
    PartialEq::<T>::eq(a, b)
}
No usage example was provided, making it unclear how to call the function.

---

The section now includes a complete, executable example:

fn contains<T, +PartialEq<T>>(a: @T, b: @T) -> bool {
    PartialEq::<T>::eq(a, b)
}

fn main() {
    let p1 = Point { x: 1, y: 2 };
    let p2 = Point { x: 1, y: 2 };
    assert!(contains(@p1, @p2));
}This makes the example compilable and demonstrates practical usage of `PartialEq` trait bounds in generic functions.

---

---

## Additional context

This change improves consistency with `comparison-operators.adoc`, which includes complete examples with usage. The example uses the `Point` struct already defined earlier in the same document.

### DIFF
--- a/docs/reference/src/components/cairo/modules/language_constructs/pages/equality-operators.adoc
+++ b/docs/reference/src/components/cairo/modules/language_constructs/pages/equality-operators.adoc
@@ -1,42 +1,42 @@
 :cairo-repo: https://github.com/starkware-libs/cairo/blob/main
 
 = Equality operators
- 
+
  Equality in Cairo is provided by the `==` (equality) and `!=` (inequality) operators.
  Both operators evaluate to `bool`.
- 
+
  == Syntax
- 
+
  - `lhs == rhs` → `bool`
  - `lhs != rhs` → `bool`
- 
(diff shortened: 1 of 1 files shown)


orizi:
the generic example already showed the usage of the operator - using the function itself is useless for the example.

Closing.

PR #9354, opened 2025-12-26
test: expand invert_or_nullify test coverage with multiple moduli

mattvaux:
## Summary

Expands test coverage for `invert_or_nullify` by adding test cases with prime moduli (7, 11) and composite moduli (9, 12, 15). Previously, all tests used modulus 8.

---

## Type of change

Please check **one**:

- [ ] New feature
- [ ] Performance improvement

> ⚠️ Note:

---

## Why is this change needed?

The function works with `BigUint` and should handle any modulus, but tests only used modulus 8. This could miss issues with prime moduli (common in cryptography) and different composite moduli. Other tests in the codebase use varied moduli, and the current coverage doesn't reflect this.

---

All 8 test cases used `modulus = 8` only.

---

Added 16 test cases covering prime moduli (7, 11) and composite moduli (9, 12, 15). Original 8 tests preserved. Total: 24 test cases covering both invertible and non-invertible cases across different modulus types.

---

None.

---

## Additional context

The function uses the extended Euclidean algorithm. Added tests verify behavior with small primes, powers of primes, and products of distinct primes, aligning with patterns in other test files (`circuit_test.cairo`, `math_test.cairo`).

### DIFF
--- a/crates/cairo-lang-runner/src/casm_run/circuit_test.rs
+++ b/crates/cairo-lang-runner/src/casm_run/circuit_test.rs
@@ -52,6 +52,107 @@ use crate::casm_run::circuit::invert_or_nullify;
     7_u32;
     "invert_or_nullify(7, 8)"
 )]
+// Tests with prime modulus 7
+#[test_case(
+    1_u32,
+    7_u32,
+    1_u32;
+    "invert_or_nullify(1, 7) - prime modulus"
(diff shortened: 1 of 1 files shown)


orizi:
This is not the only test for circuits - it is actually not even the main test for it.

Close PR.

PR #9355, opened 2025-12-26
bugfix(semantic): avoid panic when resolving use path parent

Olexandr88:
Avoid a potential panic when resolving use self paths.


orizi:

Additionally - this panic cannot actually happen - so this is no improvement.

Closing.

PR #9356, opened 2025-12-26
perf: Avoid double token stream reconstruction

Snezhkko:
## Summary

SimpleParserDatabase::parse_token_stream used Parser::parse_token_stream, which internally calls primitive_token_stream_content_and_offset again on the same token stream. In this context we already created a virtual file from the reconstructed content, so the second reconstruction and consistency check were redundant.

---

## Type of change

- [x] Performance improvement

## Why is this change needed?

Route the call through get_syntax_root_and_diagnostics instead, keeping the same public behaviour while avoiding an extra token stream traversal and string allocation.

### DIFF
--- a/crates/cairo-lang-parser/src/utils.rs
+++ b/crates/cairo-lang-parser/src/utils.rs
@@ -86,9 +86,7 @@ impl SimpleParserDatabase {
             original_item_removed: false,
         })
         .intern(self);
-        let mut diagnostics = DiagnosticsBuilder::default();
-        let syntax = Parser::parse_token_stream(self, &mut diagnostics, file_id, token_stream);
-        (syntax.as_syntax_node(), diagnostics.build())
+        get_syntax_root_and_diagnostics(self, file_id)
     }
 
     /// Parses a token stream (based on a single expression).


orizi (changes requested):
> @orizi made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/9356)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @Snezhkko).

___
*[`crates/cairo-lang-parser/src/utils.rs` line 89 at r1](https://reviewable.io/reviews/starkware-libs/cairo/9356#-OhRixwMBbd1P10gxbp9:-OhRixwMBbd1P10gxbpA:b4gmcd6) ([raw file](https://github.com/starkware-libs/cairo/blob/ff6e65243fee82d445b53bfa97123ceb69c537a8/crates/cairo-lang-parser/src/utils.rs#L89)):*
> ```XML
>         })
>         .intern(self);
>         get_syntax_root_and_diagnostics(self, file_id)
> ```

this does not seem the same at all - looking at the implementation of the removed function.

in any case - it seems this function is called from testing code only - so i'm even less sure this is in any way helpful.

PR #9363, opened 2025-12-28
fix(proc-macros): fix inaccurate comment about enum pattern bindings

hawkadrian:
## Summary

Fixes inaccurate comment in `HeapSize` derive macro that incorrectly stated enum pattern bindings are always references. Updated to accurately describe Rust's automatic borrowing behavior.

---

## Type of change

Please check **one**:

- [ ] New feature
- [ ] Performance improvement

> ⚠️ Note:

---

## Why is this change needed?

The comment incorrectly stated bindings are always references. In reality, non-Copy types are auto-referenced, while Copy types are auto-borrowed when passed to `&T` methods. The inaccurate comment could mislead developers maintaining this macro.

---

// In enum patterns, the bindings are already references, so we don't need &---

t
// In enum patterns when matching on &self, bindings for non-Copy types
// are automatically references. For Copy types, Rust automatically borrows
// them when passed to methods expecting &T, so we don't need & here.---

---

## Additional context

The code was already correct. This only fixes the documentation to accurately reflect Rust's automatic borrowing semantics.

### DIFF
--- a/crates/cairo-lang-proc-macros/src/heap_size.rs
+++ b/crates/cairo-lang-proc-macros/src/heap_size.rs
@@ -96,7 +96,9 @@ fn emit_variant_heap_size(fields: &syn::Fields) -> (TokenStream2, TokenStream2) 
         pattern = quote! { #pattern #field_ident, };
 
         if !should_skip {
-            // In enum patterns, the bindings are already references, so we don't need &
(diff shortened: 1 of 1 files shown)


orizi:
This doesn't actually fix the documentation, and does not pass the bar to be a "oncrete technical impact"

PR #9394, opened 2026-01-04
fix(sierra-gen): remove unnecessary mut from KnownStack::get

strmfos:
## Summary

Remove unnecessary `&mut self` from `KnownStack::get` method - it only reads data, never mutates.

## Type of change

## Why is this change needed?

`KnownStack::get` takes `&mut self` but doesn't mutate anything internally. This is misleading and unnecessarily restricts callers.


orizi:
This is definitely only internal styling issue.

Closing PR.

PR #9414, opened 2026-01-05
perf: cache SyntaxKind name in parser printer

Galoretka:
## Summary

Cache the formatted SyntaxKind name and the ignored-kind flag in print_internal_node instead of calling format!("{kind:?}") and scanning ignored_kinds multiple times per node

---

## Type of change

- [x] Performance improvement

---

## Why is this change needed?

This keeps the public API and printed output exactly the same, but reduces temporary string allocations and duplicate linear searches over the ignored kinds list for each internal node in the printed syntax tree.

### DIFF
--- a/crates/cairo-lang-parser/src/printer.rs
+++ b/crates/cairo-lang-parser/src/printer.rs
@@ -150,8 +150,9 @@ impl<'a> Printer<'a> {
         kind: SyntaxKind,
         under_top_level: bool,
     ) {
+        let kind_name = format!("{kind:?}");
         let current_is_top_level =
-            !under_top_level && self.top_level_kind == Some(format!("{kind:?}"));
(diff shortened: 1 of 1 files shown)


orizi (changes requested):
> @orizi made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/9414)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @Galoretka).

___
*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/9414#-OiE0A61DuC6pt-NIicP:-OiE0A61DuC6pt-NIicQ:b4nyxrt) (no related file):*
this is only relevant for testing code of the parser - and not ALWAYS more efficient - i don't think it is worth it.

PR #9440, opened 2026-01-10
refactor(lowering): eliminate redundant Vec clone in optimization strategy iteration

0xlupin:
Removes an unnecessary `Vec::clone()` allocation when iterating over optimization phases in `OptimizationStrategyId::apply_strategy`.


orizi:
made obsolete by #9443

PR #9449, opened 2026-01-11
fix: replace panic with fallback in colored printer token handling

hawkadrian:
## Summary

Replaces `panic!` in `set_color` function with a graceful fallback that returns uncolored text for unhandled token kinds. This prevents crashes when encountering tokens like `TokenAt`, `TokenDollar`, `TokenPub`, or newly added tokens that aren't explicitly handled in the match expression.

---

## Type of change

Please check **one**:

- [ ] New feature
- [ ] Performance improvement

> ⚠️ Note:

---

## Why is this change needed?

The `set_color` function in `colored_printer.rs` uses `panic!` for unhandled `SyntaxKind` variants. Several existing tokens (e.g., `TokenAt`, `TokenDollar`, `TokenPub`, `TokenMacro`, `TokenContinue`, `TokenBreak`, assignment operators like `TokenDivEq`, `TokenMinusEq`, etc.) are not explicitly handled, which causes the program to crash when these tokens are encountered during colored printing. Additionally, any future token additions to the language would trigger the same panic.

This is particularly problematic because `print_colored` is used in test utilities, and crashes prevent proper debugging output.

---

When `set_color` encountered an unhandled token kind (either existing unhandled tokens or newly added ones), it would call `panic!("Unexpected syntax kind: {kind:?}")`, causing the entire program to crash. The TODO comment on line 126 acknowledged this limitation but didn't provide a solution.

---

Unhandled token kinds now return `text.clear()` (uncolored text) instead of panicking. This allows the colored printer to continue functioning even when encountering tokens that aren't explicitly colorized, making the code resilient to new token additions without requiring immediate updates to the color mapping.

---

Addresses the TODO comment on line 126: `// TODO(yuval): Can this be made exhaustive?`

---

## Additional context

The change is minimal and maintains backward compatibility. All existing handled tokens continue to work as before. The fallback behavior (uncolored text) is consistent with how other "neutral" tokens like `TokenEndOfFile`, `TokenMissing`, and `TokenEmpty` are handled in the same function.

All tests pass (69 tests), and the code compiles without warnings.

### DIFF
--- a/crates/cairo-lang-parser/src/colored_printer.rs
+++ b/crates/cairo-lang-parser/src/colored_printer.rs
@@ -123,9 +123,13 @@ fn set_color(text: &str, kind: SyntaxKind) -> ColoredString {
         | SyntaxKind::TokenWhitespace
         | SyntaxKind::TokenNewline
         | SyntaxKind::TokenEmpty => text.clear(),
-        // Fallback for unhandled token kinds (e.g., TokenAt, TokenDollar, TokenPub, etc.)
-        // Returns uncolored text to avoid panicking when new tokens are added.
-        _ => text.clear(),
+        _ => {
+            if kind.is_token() {
+                panic!("Unhandled token kind in colored printer: {:?}", kind);
+            }
+            // For non-token kinds, return uncolored text
+            text.clear()
+        }
     }
 }
 


orizi (changes requested):
> @orizi made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/9449)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @hawkadrian).

___
*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/9449#-OiiMPQ841kROXUAbqBL:-OiiMPQ841kROXUAbqBM:bthjygg) (no related file):*
since the printer isn't used in any of the production tools - rather panic and find the missing kinds not handled when actually used.

PR #9455, opened 2026-01-12
docs: align constant evaluation with const calc implementation

yyhrnk:
## Summary

Update the constant evaluation reference to match the current const calculation implementation.

---

## Type of change

Please check **one**:

---

## Why is this change needed?

The downcast externs list was corrected to drop u128_try_from_felt252, which is not actually allowed in const evaluation, and the diagnostics section was expanded to include FailedConstantCalculation, ConstCycle, ConstTypeNotVarFree, and LiteralError so that the documented behavior reflects the compiler’s real error reporting in const contexts.

---

### DIFF
--- a/docs/reference/src/components/cairo/modules/language_semantics/pages/constant-evaluation.adoc
+++ b/docs/reference/src/components/cairo/modules/language_semantics/pages/constant-evaluation.adoc
@@ -70,7 +70,7 @@ Allowed externs (from `ConstCalcInfo::new`):
     enabling const upcast to `u256` and downcast to `u128`.
 - Downcast / trimming (returning an `Option`-like enum; some are reversed per configuration):
   * `core::internal::bounded_int::{downcast,bounded_int_trim_min,bounded_int_trim_max}`.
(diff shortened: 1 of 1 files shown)


orizi (changes requested):
> @orizi made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/9455)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @yyhrnk).

___
*[`docs/reference/src/components/cairo/modules/language_semantics/pages/constant-evaluation.adoc` line 106 at r1](https://reviewable.io/reviews/starkware-libs/cairo/9455#-OimYvLdDl6g4x2D2AlY:-OimZAcmBQvaNpIa7rUT:b-66isej) ([raw file](https://github.com/starkware-libs/cairo/blob/1db863aa036d639987d3fa1db389b694f0d118b4/docs/reference/src/components/cairo/modules/language_semantics/pages/constant-evaluation.adoc#L106)):*

These failures have nothing to do with const evaluation - only with actual const items.
</blockquote></details>

LiteralError - is does happen - but just by basic known values not fitting their type - not malformed.

PR #9472, opened 2026-01-13
docs: add function example with return value to ABI documentation

forkfury:
## Summary

Adds a function example with return value to the ABI documentation. The existing example only showed a function with empty `outputs`, which doesn't demonstrate how to specify return values in ABI JSON.

---

## Type of change

Please check **one**:

- [ ] New feature
- [ ] Performance improvement

> ⚠️ Note:

---

## Why is this change needed?

The ABI documentation showed only one function example with empty `outputs: []`. While there is an example with non-empty `outputs` in the interface section, it's not immediately obvious. Many real-world functions return values, and developers need a clear example of how to specify return types in the `outputs` array.

---

The documentation had a function example with empty `outputs: []` (the `transfer` function). An example with non-empty `outputs` existed only nested inside an interface example, which is less discoverable.

---

The documentation now includes a dedicated function example (`balance_of`) that demonstrates how to specify return values in the `outputs` array, using a view function returning `u256`. This example is placed right after the basic function example.

---

---

## Additional context

The example uses `balance_of` returning `u256` as it's a familiar pattern from ERC20-like contracts and matches the existing interface example in the same file.

### DIFF
--- a/docs/reference/src/components/cairo/modules/language_semantics/pages/application-binary-interface.adoc
+++ b/docs/reference/src/components/cairo/modules/language_semantics/pages/application-binary-interface.adoc
@@ -43,6 +43,20 @@ Function item:
 }
 ----
 
+Function item (with return value):
+[source,json]
+----
+{
+  "type": "function",
+  "name": "balance_of",
+  "inputs": [
+    { "name": "account", "type": "ContractAddress" }
+  ],
+  "outputs": [ { "type": "u256" } ],
+  "state_mutability": "view"
+}
+----
+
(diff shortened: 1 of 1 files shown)


orizi (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/9472#-:-OitWakI462pH8Zeng6H:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @forkfury).

PR #9475, opened 2026-01-14
docs: add array and fixed size array examples to ABI

JayeTurn:
Adds missing ABI JSON examples for `Array` and `FixedSizeArray` types.

While tuples and basic types are documented, the specific serialization formats for arrays (`core::array::Array::<T>`) and fixed-size arrays (`[T; N]`) were missing. These examples clarify the exact string representation expected in the generated ABI, preventing ambiguity for tooling developers.


JayeTurn:
> > @orizi made 1 comment.
> > _[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/9475)_ status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @JayeTurn).
> 
> _[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/9475#-Oiwg_V1CUGL5pRx5zFq:-Oiwg_V1CUGL5pRx5zFr:bc4dezp) (no related file):_ this is not a spec file - no reason to have an example for each and every option.

Hi @orizi 
I understand this isn't meant to be exhaustive. However, arrays are a fundamental type that many contracts use, and the string representation (`core::array::Array::<T>` vs just `Array<T>`) isn't obvious from the existing examples.
Would it be acceptable to add just one array example (perhaps combining both Array and FixedSizeArray into a single function example) to clarify the namespace format?

JayeTurn:
> > @orizi made 1 comment.
> > _[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/9475)_ status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @JayeTurn).
> 
> _[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/9475#-Oiwg_V1CUGL5pRx5zFq:-Oja5pFU0Uj-2rhhgRKo:b-c5an7b) (no related file):_
> 
> Previously, JayeTurn (Jaye Turner) wrote…
> why should it matter to anyone reading this doc specifically? this still seems useless.

The main utility of this example is for developers building tooling (like SDKs or indexers) or interacting directly with the ABI who need to verify the precise string representation of types.

Since generic Arrays are such a fundamental type, explicitly documenting that they appear as `core::array::Array<T>` (rather than just `Array`) removes ambiguity. This visual reference saves developers from having to compile a dummy contract just to reverse-engineer the expected type string when writing integration tests or off-chain scripts.

I can combine the Array and FixedSizeArray into a single example as suggested to keep the entry concise, if that sounds good?

PR #9476, opened 2026-01-14
performance(sierra-gas): Made vec be allocated with capacity.

orizi:
## Summary

Optimize memory allocation in the `compute_costs` function by pre-allocating capacity for the `branch_costs` vector based on the program's statement count, rather than using the default capacity.

---

## Type of change

Please check **one**:

- [ ] New feature
- [x] Performance improvement

---

## Why is this change needed?

This change improves performance by reducing memory reallocations during the cost computation process. By pre-allocating the `branch_costs` vector with the exact capacity needed (based on the program's statement count), we avoid multiple costly reallocations that would occur as the vector grows.

---

Previously, `branch_costs` was initialized with `Default::default()`, which creates a vector with zero initial capacity. This meant the vector would need to reallocate memory multiple times as elements were added during processing.

---

Now `branch_costs` is initialized with a capacity equal to the number of statements in the program, ensuring that all elements can be added without triggering any reallocations.

---

## Additional context

This is a small but meaningful optimization for the gas cost computation process, which can be performance-sensitive when dealing with large programs.

### DIFF
--- a/crates/cairo-lang-sierra-gas/src/compute_costs.rs
+++ b/crates/cairo-lang-sierra-gas/src/compute_costs.rs
@@ -124,7 +124,7 @@ pub fn compute_costs<
 ) -> Result<GasInfo, CostError> {
     let mut context = CostContext {
         program,
-        branch_costs: Default::default(),
+        branch_costs: Vec::with_capacity(program.statements.len()),
         enforced_wallet_values,
         costs: Default::default(),
         target_values: Default::default(),


orizi:
* **#9476** <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/9476?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/9476?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #9501, opened 2026-01-20
Refactor compiler db string creation

MrEeeeet111:
## Summary

Replaced instances of `"".to_string()` and `"".into()` with `String::new()` across the codebase.

---

## Type of change

Please check **one**:

- [ ] New feature
- [ ] Performance improvement
- [x] Style, wording, formatting, or typo-only change

> ⚠️ Note:

---

## Why is this change needed?

`String::new()` is the idiomatic way to create an empty string in Rust. It is clearer and more explicit than converting an empty string literal. While `"".to_string()` is often optimized, `String::new()` guarantees no allocation check and is the standard practice for initializing empty strings.

---

The code used `"".to_string()` or `"".into()` to generate empty `String` instances.

---

The code now consistently uses `String::new()` for empty string generation.

---

---


orizi:
> To keep maintainer workload sustainable, we generally do not accept PRs that

PR #9516, opened 2026-01-22
Refine felt252 const op reference info

AaronAjose:
Updates `felt252_add_const` and `felt252_sub_const` to use `Deferred(AddConst)` instead of `Generic`.
These operations guarantee a `Var + Const` result structure. Explicitly marking them as `AddConst` allows the compiler to track variable offsets more accurately, enabling potential constant folding and AP tracking optimizations that `Generic` obscures.


orizi:
Additionally - This specific case requires internal heavy auditing for approval, so rather not handle it, since the relevant libfuncs are not even active.

Closing.

PR #9517, opened 2026-01-22
Vastly removed the clones in sierra, bt not cloning type info.

orizi:
## Summary

Optimize memory usage by changing `TypeSpecializationContext::try_get_type_info` to return a reference to `TypeInfo` instead of cloning it. This change propagates throughout the codebase, updating all implementations and callers to work with references rather than owned values.

---

## Type of change

Please check **one**:

- [ ] New feature
- [x] Performance improvement

---

## Why is this change needed?

The current implementation of `try_get_type_info` returns a cloned `TypeInfo` object, which creates unnecessary copies of data structures. By changing the method to return a reference instead, we avoid these allocations and copies, improving performance especially in code paths that frequently access type information.

---

Previously, `try_get_type_info` returned an owned `TypeInfo` object, requiring a clone operation each time type information was accessed. This was inefficient, especially in hot code paths that frequently query type information.

---

Now `try_get_type_info` returns a reference (`&TypeInfo`), avoiding unnecessary cloning. All callers have been updated to work with references rather than owned values, and generic type implementations now take `&TypeInfo` parameters instead of owned values.

---

## Additional context

This change is part of ongoing performance optimizations in the Sierra compiler. It reduces memory allocations and copying, which should improve compilation speed, especially for large programs with many type references.

### DIFF
--- a/crates/cairo-lang-sierra-generator/src/specialization_context.rs
+++ b/crates/cairo-lang-sierra-generator/src/specialization_context.rs
@@ -16,8 +16,8 @@ impl TypeSpecializationContext for SierraSignatureSpecializationContext<'_> {
     fn try_get_type_info(
         &self,
         id: &cairo_lang_sierra::ids::ConcreteTypeId,
-    ) -> Option<cairo_lang_sierra::extensions::types::TypeInfo> {
-        self.0.get_type_info(id.clone()).cloned().to_option()
+    ) -> Option<&cairo_lang_sierra::extensions::types::TypeInfo> {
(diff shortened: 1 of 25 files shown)


orizi:
* **#9517** <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/9517?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/9517?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #9538, opened 2026-01-26
perf: Optimize felt252 add/sub with constants using deferred references

AaronAjose:
## Summary
Optimizes `felt252` addition and subtraction with constants by utilizing `DeferredOutputKind::AddConst`. This allows the compiler to represent operations like `a + 5` or `a - 10` as reference offsets (e.g., `[ap] + 5`) instead of generating CASM arithmetic instructions, where applicable.

---
## Type of change
Please check **one**:
- [ ] New feature
- [x] Performance improvement
> ⚠️ Note:
---
## Why is this change needed?
Currently, `felt252` operations with constants (e.g., `x + 1`) often generate explicit CASM instructions, consuming trace cells and gas. However, the Sierra compiler infrastructure supports "deferred" outputs where a value is represented as a reference plus a constant offset (`OutputVarReferenceInfo::Deferred(DeferredOutputKind::AddConst)`).
This optimization is already used for other types (e.g., `u128` operations, struct member access, and dictionary operations) but was missing for the native `felt252` type. Adding this significantly reduces the cost of simple arithmetic found in loops and offsets.

---
`felt252_add_const` and `felt252_sub_const` libfuncs were specialized with `DeferredOutputKind::Generic`, forcing the `sierra-to-casm` compiler to generate instructions like `[ap] = [ap] + c` for the result.

---
These libfuncs now use `DeferredOutputKind::AddConst`. The `sierra-to-casm` compiler's `build_felt252_op_with_const` has been updated to produce a `ReferenceExpression` containing a `CellExpression::BinOp` (base + offset) instead of emitting instructions, whenever the base reference allows it.
For example, `let y = x + 5;` where `x` is `[ap]`, will now result in `y` being tracked as `[ap] + 5` at compile time, costing 0 instructions.

---
N/A

---
## Additional context
Verified that negative constants (for subtraction) are correctly handled by `BigIntAsHex` and `CellExpression` logic. Fallback mechanisms ensure that if the expression cannot be represented as a simple offset (e.g. `DoubleDeref`), it reverts to instruction generation.

### DIFF
--- a/crates/cairo-lang-sierra-to-casm/src/invocations/felt252.rs
+++ b/crates/cairo-lang-sierra-to-casm/src/invocations/felt252.rs
@@ -1,6 +1,7 @@
 use cairo_lang_casm::builder::{CasmBuilder, Var};
 use cairo_lang_casm::casm_build_extend;
 use cairo_lang_casm::cell_expression::{CellExpression, CellOperator};
+use cairo_lang_casm::operand::DerefOrImmediate;
 use cairo_lang_sierra::extensions::felt252::{
     Felt252BinaryOpConcreteLibfunc, Felt252BinaryOperationConcrete, Felt252BinaryOperator,
     Felt252Concrete, Felt252OperationWithConstConcreteLibfunc,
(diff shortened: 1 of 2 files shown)


orizi:
PRs regarding performance improvements for sierra-to-casm products are not accepted currently, without very specific reasons, and with proof of not changing any existing sierra-to-casm contract compilation.

orizi:
Closing.

PR #9542, opened 2026-01-26
refactor: derive const type from ConstValueId

lisenokdonbassenok:
## Summary

This change makes the Const generator derive the variable type directly from ConstValueId::ty(ctx.db)
and removes the redundant ty field. All call sites of generators::Const in lowering are updated to
stop passing a type explicitly and rely on the type embedded in the const value.

---

## Type of change

- [x] New feature

---

## Why is this change needed?

Previously the lowering Const generator duplicated type information: it stored both a ConstValueId and an explicit semantic::TypeId. This opened the door for potential divergence between the value's actual type and the manually passed type.

---


orizi:
Closing.

PR #9561, opened 2026-01-29
bugfix(semantic): Fixed const-impl-item type-diagnostics span.

orizi:
## Summary

Fixed the error location in trait constant type mismatch diagnostics. The error now points to the type annotation itself rather than the entire constant declaration.

---

## Type of change

Please check **one**:

- [x] Bug fix (fixes incorrect behavior)
- [ ] New feature
- [ ] Performance improvement
- [ ] Documentation change with concrete technical impact
- [ ] Style, wording, formatting, or typo-only change

---

## Why is this change needed?

When a trait constant implementation had a type mismatch, the error diagnostic was pointing to the entire constant declaration rather than specifically to the type annotation. This made it harder for users to identify exactly where the type mismatch occurred.

---

## What was the behavior or documentation before?

Previously, when there was a type mismatch in a trait constant implementation, the error would point to the entire constant declaration line. For example, in the test case, the error pointed to `const X: i32 = 9;` rather than just the `i32` type.

---

## What is the behavior or documentation after?

Now the error points specifically to the type annotation (`i32` in the test case), making it clearer to users exactly what needs to be changed to fix the type mismatch.

PR #9562, opened 2026-01-29
bugfix(semantic): Fix mishandling of `Self::` in impl-const-items.

orizi:
## Summary

Fixed a bug in trait constant type validation by properly substituting the concrete trait constant type with the implementation definition substitution. This ensures that trait constants using `Self::SIZE` in their type definition resolve correctly.

---

## Type of change

Please check **one**:

- [ ] New feature
- [ ] Performance improvement

---

## Why is this change needed?

When implementing a trait with constants that reference other constants in their type definition (like array sizes using `Self::SIZE`), the type validation was failing because it wasn't properly substituting the implementation-specific values.

---

Previously, when validating an implementation of a trait constant, the code was using the concrete trait constant type directly without applying the implementation's substitutions. This would cause validation errors when trait constants used other trait constants in their type definitions.

---

Now, the implementation's substitution is applied to the concrete trait constant type during validation. This allows trait constants to correctly reference other constants like `Self::SIZE` in their type definitions, as demonstrated by the added test case.

---

## Additional context

The added test case demonstrates a common pattern where a trait defines a size constant and then uses it in the type of another constant (in this case, an array size). This pattern is now properly supported.

### DIFF
--- a/crates/cairo-lang-semantic/src/items/imp.rs
+++ b/crates/cairo-lang-semantic/src/items/imp.rs
@@ -2684,7 +2684,9 @@ fn validate_impl_item_constant<'db>(
         })?;
     let concrete_trait_constant =
         ConcreteTraitConstantId::new_from_data(db, concrete_trait_id, trait_constant_id);
-    let concrete_trait_constant_ty = db.concrete_trait_constant_type(concrete_trait_constant)?;
+    let impl_def_substitution = db.impl_def_substitution(impl_def_id)?;
+    let concrete_trait_constant_ty = impl_def_substitution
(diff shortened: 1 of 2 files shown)


orizi:
* **#9562** <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/9562?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/9562?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#9561** <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/9561?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

PR #9565, opened 2026-01-29
feature(semantic): Allow immediate consumption of `ref self` of tempraries.

orizi:
## Summary

Added support for calling `ref self` methods on temporary expressions, allowing code like `make_counter().increment()` where `increment` takes `ref self`. This enables method chaining on temporary values that modify self.

---

## Type of change

Please check **one**:

- [x] New feature
- [ ] Performance improvement

---

## Why is this change needed?

Previously, methods that take `ref self` could only be called on variables, not on temporary expressions. This limitation prevented method chaining for methods that modify self, requiring users to create intermediate variables. This change allows for more fluent API design where methods can modify state and return self for chaining.

---

Code like `make_counter().increment()` would fail to compile if `increment` took `ref self`, with an error message indicating that ref arguments must be variables. This forced developers to use intermediate variables even for simple method chains.

---

Methods that take `ref self` can now be called on temporary expressions, enabling patterns like:
```cairo
make_counter().increment().increment()
```

This works by introducing a new `TempReference` variant for function call arguments that handles passing temporary expressions by reference.

---

## Additional context

The implementation:
1. Adds a new `TempReference` variant to `ExprFunctionCallArg`
2. Restructures `NamedArg` to track when temporary references are allowed
3. Updates the lowering pass to handle the new argument type
4. Adds test cases to verify the functionality

This change only applies to `ref self` parameters in method calls, not to regular `ref` parameters in function calls, which still require variables.

### DIFF
--- a/crates/cairo-lang-debug/src/debug.rs
+++ b/crates/cairo-lang-debug/src/debug.rs
@@ -4,7 +4,6 @@ mod test;
 
 // Mostly taken from https://github.com/salsa-rs/salsa/blob/fd715619813f634fa07952f0d1b3d3a18b68fd65/components/salsa-2022/src/debug.rs
 use std::collections::{BTreeMap, BTreeSet, HashMap, HashSet};
-use std::hash::Hash;
 use std::rc::Rc;
 use std::sync::Arc;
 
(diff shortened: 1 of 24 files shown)


orizi:
* **#9569** <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/9569?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#9565** <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/9565?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/9565?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #9572, opened 2026-01-30
fix: simulation of array_pop_front_consume outputs

Bashmunta:
## Summary

Adjust the ArrayPopFrontConsume simulation to return the updated array and front element on the success branch and no values on the empty branch, aligning the simulator behavior with the libfunc definition, corelib API, and existing e2e tests.

---

## Type of change

---

## Why is this change needed?

The Sierra simulator returned mismatching outputs for array_pop_front_consume: it produced only the element on the non-empty branch and returned the array on the empty branch, violating the libfunc signature and corelib semantics and potentially breaking follow-up variable usage.


orizi:
Simulation in the cairo repo is no longer supported, and would be moved soon.

Closing PR.

PR #9586, opened 2026-02-01
refactor: simplify OrderedHashSet::PartialEq implementation

splinter012:
Simplified the PartialEq implementation for OrderedHashSet by replacing manual length check and zip_eq with the standard Iterator::eq method.


orizi:
no response. closing.

PR #9588, opened 2026-02-02
fix: prevent variable shadowing in event derive macro

Princetimix69:
## Summary

Fixes a variable shadowing issue in the `derive(starknet::Event)` macro where struct members named `keys` or `data` would collide with the generated `deserialize` function arguments.

---

## Type of change

Please check **one**:

- [ ] New feature
- [ ] Performance improvement

> ⚠️ Note:

---

## Why is this change needed?

When a user defines a Starknet Event struct with members named `keys` or `data`, the generated `deserialize` function creates local variables with the same names. These local variables shadow the `ref keys` and `ref data` arguments required for the deserialization process, causing compilation errors or incorrect behavior.

By prefixing the local variables for deserialized values (e.g., using `__val_keys` instead of `keys`), this PR ensures that the function arguments remain accessible throughout the function body.

---

The `derive(starknet::Event)` macro generated variable names identical to their struct member names. If a member was named `keys` or `data`, it would shadow the deserialization buffer arguments `keys` and `data`.

---

The macro now generates local variables with a `__val_` prefix (e.g., `__val_membername`). This prevents any collision with the fixed argument names `keys` and `data`, allowing Event structs to safely use these names as member fields.

---

---

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -123,30 +123,19 @@ jobs:
       - uses: Swatinem/rust-cache@v2
       - run: scripts/cairo_fmt.sh --check
 
-  parallel-cairo-tests:
+  # Checks that all Cairo code tests run correctly.
+  cairotest:
     runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        cmd:
-          - corelib/
-          - tests/bug_samples --starknet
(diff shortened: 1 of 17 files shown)


orizi:
no response.
Closing.

PR #9590, opened 2026-02-03
docs: add comprehensive path expressions documentation

Evelina331:
## Summary

Expanded the path expressions documentation to explain local variable and constant usage, including code examples and clearer formatting.

---

## Type of change

Please check **one**:

- [ ] New feature
- [ ] Performance improvement

> ⚠️ Note:

---

## Why is this change needed?

The previous documentation was a bare bullet list that lacked examples and context on how paths are used as expressions.

---

The file contained only a four-line list of item types that a path could resolve to.

---

The file now includes dedicated sections for local variables and constants with code examples showing their usage in functions and expressions.

---

---


Evelina331:
@orizi Done. Updated the generic const parameter note and replaced with a qualified path example.

PR #9608, opened 2026-02-04
refactor(lowering): remove unused dataflow analysis core module

Aleksandr1732:
Removes the unused `analysis::core` module containing `DataflowAnalyzer` trait and related types that were never utilized in the codebase.


orizi:
module is currently WIP please do not contribute here.

PR #9611, opened 2026-02-04
fix(parser): correct skip_until TextSpan range

MoNyAvA:
## Summary

adjusted skip_until to derive the span directly from the token text range (excluding trivia) and added a focused unit test that checks the returned TextSpan covers all skipped tokens and starts at the first non-trivia character.

---

## Type of change

Please check **one**:

---

## Why is this change needed?

Before this change, skip_until was computing SkippedElement spans from the raw parser offset, including leading trivia and stopping inside the token text. As a result, diagnostics for skipped tokens were slightly shifted and inconsistent with other span calculations.


orizi:
No response.
Closing.

PR #9613, opened 2026-02-05
fix: avoid redundant cloning in filtered patterns bindings

sashass1315:
## Summary

Made FilteredPatterns::add_bindings take a &[Bindings] instead of an owned Vec and updates create_node_for_patterns to pass &bindings. The behavior of the cache and the resulting flow control graph stays the same, but we eliminate the redundant deep cloning of all bindings on each callback call.

---

## Type of change

Please check **one**:

- [x] Performance improvement

## Why is this change needed?

Before this change FilteredPatterns::add_bindings took Vec<Bindings> by value, and create_node_for_patterns passed bindings.clone() into it. This caused every callback invocation to allocate a new Vec and then clone all bindings again when extending the existing ones, leading to unnecessary work and allocations in the match lowering flow control graph construction.

---

### DIFF
--- a/crates/cairo-lang-lowering/src/lower/flow_control/create_graph/filtered_patterns.rs
+++ b/crates/cairo-lang-lowering/src/lower/flow_control/create_graph/filtered_patterns.rs
@@ -66,10 +66,10 @@ impl FilteredPatterns {
         self.filter.into_iter().next()
     }
 
-    /// Adds bindings to the filter. The length of the vector must be equal to the number of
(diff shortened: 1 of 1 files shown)


orizi (approved):
:lgtm:

> @orizi reviewed 2 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/9613#-:-OkwBaS01iCT6xzweQvG:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @sashass1315).

PR #9623, opened 2026-02-09
perf(encoder): avoid unnecessary BigInt cloning in encode()

GarmashAlex:
## Summary

Changed `encode(&self)` to `encode(self)` to allow moving the `BigInt` value directly without cloning. Removed the `.clone()` call on `self.imm`. This eliminates the extra allocation while maintaining the same output behavior.

---

## Type of change

Please check **one**:

- [x] Performance improvement

---

## Why is this change needed?

The `encode()` method was taking `&self` and cloning `self.imm` (BigInt) when an immediate value was present. Since [InstructionRepr]https://github.com/starkware-libs/cairo/blob/fd9cc11559a81b5d1bc113e1360765b6b164c850/crates/cairo-lang-casm/src/assembler.rs#L68-L84 is always created temporarily and never reused after encoding, this cloning is unnecessary overhead, especially for large immediate values.

---

### DIFF
--- a/crates/cairo-lang-casm/src/encoder.rs
+++ b/crates/cairo-lang-casm/src/encoder.rs
@@ -32,7 +32,7 @@ const OPCODE_ASSERT_EQ_BIT: i32 = 14;
 const OPCODE_EXT_OFFSET: u64 = 63;
 
 impl InstructionRepr {
-    pub fn encode(&self) -> Vec<BigInt> {
+    pub fn encode(self) -> Vec<BigInt> {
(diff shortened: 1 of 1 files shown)


orizi (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/9623#-:-Ol0lVt4CJfq_UmbDh5W:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @GarmashAlex).

PR #9628, opened 2026-02-09
fix: remove redundant `collect_extra_allowed_attributes` call

ANtutov:
## Summary

Removed the redundant `collect_extra_allowed_attributes` call from `validate_attributes_flat` and use the passed `extra_allowed_attributes` parameter directly. This eliminates duplicate work while maintaining the same validation behavior.

---

## Type of change

---

## Why is this change needed?

The `validate_attributes_flat` function redundantly called `collect_extra_allowed_attributes` internally, even though the caller `validate_attributes` already collected these attributes and passed them as the `extra_allowed_attributes` parameter. This caused:
1. Double work for top-level items - attributes were collected twice
2. Unnecessary calls for nested items (trait functions, struct fields, enum variants) which cannot have `#[allow(...)]` attributes anyway

---


orizi:
No response.

Closing.

PR #9639, opened 2026-02-11
refactor(semantic): clone only keys in finalize_as_scope()

GarmashAlex:
## Summary

Changed all three loops (usage, snap_usage, changes) to collect only keys into a Vec before iteration. This eliminates unnecessary cloning of values and reduces memory allocations.

---

## Type of change

Please check **one**:

- [z] Performance improvement

---

## Why is this change needed?

The finalize_as_scope() method was cloning entire OrderedHashMap instances  (with all their values) just to iterate while modifying. This is wasteful  since we only need the keys - the values (ExprVarMemberPath) are never used in the loop body.


orizi:
Made irrelevant by #9641.

Closing.

PR #9658, opened 2026-02-14
fix: add bounds checking for cache size header reading

forkfury:
## Summary

Adds bounds checking and overflow protection when reading the cache size header in `load_cached_crate_modules`. The function now validates that the content buffer has sufficient bytes before reading the size field and checks for integer overflow when calculating data boundaries, preventing potential panics from out-of-bounds access or arithmetic overflow.

---

## Type of change

Please check **one**:

- [ ] New feature
- [ ] Performance improvement

> ⚠️ Note:

---

## Why is this change needed?

The `load_cached_crate_modules` function could panic when processing corrupted or malformed cache files. Specifically:

1. **Out-of-bounds access**: The code attempted to read 8 bytes from the content buffer without checking if it had at least 8 bytes, which could cause a panic with `try_into().unwrap()`.

2. **Integer overflow**: The calculation `size + 8` could overflow if `size` was close to `usize::MAX`, leading to undefined behavior or incorrect buffer slicing.

3. **Insufficient data**: Even if the size was read successfully, the code didn't verify that `content.len() >= size + 8` before slicing, potentially causing out-of-bounds access.

These issues could cause the compiler to crash when encountering corrupted cache files, rather than gracefully handling the error by falling back to rebuilding the cache.

---

Before this change, `load_cached_crate_modules` would:
- Attempt to read 8 bytes from the content buffer without checking its length
- Calculate `size + 8` without checking for integer overflow
- Slice the buffer without verifying sufficient data is available
- Panic if any of these operations failed (via `unwrap()` or out-of-bounds access)

This meant that corrupted cache files could crash the compiler instead of being handled gracefully.

---

After this change, `load_cached_crate_modules`:
- Checks that `content.len() >= 8` before attempting to read the size header
- Uses `checked_add(8)` to safely calculate the data end position, preventing overflow
- Verifies that `data_end <= content.len()` before slicing the buffer
- Returns `None` (which triggers cache rebuild) instead of panicking when encountering invalid cache data

The function now handles corrupted or malformed cache files gracefully by falling back to cache regeneration, improving robustness and preventing compiler crashes.

---

---

## Additional context

This is a defensive programming improvement that enhances the robustness of the cache loading mechanism. The changes are minimal and focused, maintaining the existing API contract (the function still returns `Option<ModuleDataCacheAndLoadingData<'db>>`). The error handling strategy (returning `None` to trigger cache rebuild) is consistent with the existing codebase pattern where missing or invalid cache data results in cache regeneration rather than failure.

The fix addresses a security and stability concern where malicious or corrupted cache files could potentially cause denial of service through panic, though in practice this is more likely to occur due to filesystem corruption or incomplete cache writes.

### DIFF
--- a/crates/cairo-lang-defs/src/cache/mod.rs
+++ b/crates/cairo-lang-defs/src/cache/mod.rs
@@ -94,9 +94,22 @@ pub fn load_cached_crate_modules<'db>(
         return Default::default();
     };
 
+    // Check that we have enough bytes to read the size header
+    if content.len() < 8 {
+        return Default::default();
+    }
+
     let size = usize::from_be_bytes(content[..8].try_into().unwrap());
 
-    let content = &content[8..size + 8];
+    // Check for overflow and that we have enough data
+    let Some(data_end) = size.checked_add(8) else {
+        return Default::default();
+    };
+    if data_end > content.len() {
+        return Default::default();
+    }
+
+    let content = &content[8..data_end];
 
     let (metadata, module_data, defs_lookups): DefCache<'_> = postcard::from_bytes(content)
         .unwrap_or_else(|e| {


orizi:
No response.

Closing.

PR #9670, opened 2026-02-15
refactor(diagnostics): optimize get_diagnostics_without_duplicates

Forostovec:
## Summary

- Removed unused `files_db` variable
- Changed from `sorted_by_cached_key` (cache ineffective due to unique `idx`) to `collect()` + `sort_by()`
- Pre-computed `user_span` for each diagnostic to avoid repeated `user_location()` calls
- Removed expensive `format(db)` from sort key (wasn't used for deduplication anyway)
- Simplified loop logic using pre-computed values

---

## Type of change

- [ ] Performance improvement

---

## Why is this change needed?

The previous implementation used `sorted_by_cached_key` which is intended for expensive key computations. However, the key included `(span, format(db), *idx)` where `*idx` from enumerate() guaranteed uniqueness, making the cache completely ineffective - it allocated memory without benefit.
 
Also, `format(db)`https://github.com/starkware-libs/cairo/blob/b012a898a06454c9742c33f0aa9c1fbe97151a32/crates/cairo-lang-diagnostics/src/diagnostics.rs#L326-L329 was called for every diagnostic just for sorting, but the actual deduplication logic only compared spans and called `is_same_kind()`https://github.com/starkware-libs/cairo/blob/b012a898a06454c9742c33f0aa9c1fbe97151a32/crates/cairo-lang-diagnostics/src/diagnostics_test.rs#L30-L32 This caused unnecessary string allocations.

### DIFF
--- a/crates/cairo-lang-diagnostics/src/diagnostics.rs
+++ b/crates/cairo-lang-diagnostics/src/diagnostics.rs
@@ -357,26 +357,28 @@ impl<'db, TEntry: DiagnosticEntry<'db> + salsa::Update> Diagnostics<'db, TEntry>
         if diagnostic_with_dup.is_empty() {
             return diagnostic_with_dup;
         }
-        let files_db: &'db dyn Database = db;
-        let mut indexed_dup_diagnostic =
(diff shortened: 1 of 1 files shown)


orizi:
No response.

Closing.

PR #9680, opened 2026-02-16
refactor: eliminate code duplication in extract_matches! macro

prestoalvarez:
Refactors the `extract_matches!` macro to eliminate duplicate `match` expressions. The macro previously contained two nearly identical `match` blocks that differed only in their panic message formatting, creating a maintenance risk where future changes could diverge between the two branches.


orizi:

Additionally - Only minor styling PR - and actually is much longer and less clear code.

Closing.

PR #9692, opened 2026-02-17
refactor(starknet): remove redundant clone calls in dispatcher

radik878:
## Summary

Changed `ret_decode` parameter from `String` to `&str` and updated call sites to pass references instead of cloned values, eliminating 7 unnecessary allocations per interface function processed.

---

## Type of change

---

## Why is this change needed?

The dispatcher plugin was making unnecessary `.clone()` calls when passing `ret_decode` (String), `serialization_code` (Vec), and `entry_point_selector` (RewriteNode) to the `declaration_method_impl` function. These values were either moved or not modified after being passed, making the clones redundant.

---


radik878:
> ```
>     ret_decode: String
> ```

`ret_decode: String` didn't compile

PR #9695, opened 2026-02-18
bugfix(semantic): Made enum variants stable even if there are redefinitions.

orizi:
## Summary

Fixed a bug in enum variant handling to properly detect and report duplicate variants. The PR ensures that when an enum has multiple variants with the same name, the compiler correctly identifies this as an error and prevents the second variant from being added to the semantic model. Added a test case to verify this behavior.

---

## Type of change

---

## Why is this change needed?

When an enum had multiple variants with the same name, the compiler would silently accept the duplicate variants, leading to unpredictable behavior. This could cause issues in pattern matching and other operations that rely on unique variant names.

---

Previously, when an enum had duplicate variant names, both variants would be added to the semantic model with different indices. This could lead to confusing behavior in match statements and other code that relied on unique variant names.

---

Now, the compiler properly detects duplicate variant names and reports them as errors. Only the first occurrence of a variant name is added to the semantic model, ensuring consistent behavior. A test case has been added to verify this functionality.

### DIFF
--- a/crates/cairo-lang-lowering/src/test_data/enums
+++ b/crates/cairo-lang-lowering/src/test_data/enums
@@ -42,3 +42,48 @@ Statements:
   (v5: test::MyEnum) <- MyEnum::B(v4)
 End:
   Return(v5)
+
+//! > ==========================================================================
+
+//! > Multiple variants with the same name.
+
+//! > test_runner_name
+test_function_lowering(expect_diagnostics: true)
+
+//! > module_code
+enum MyEnum {
+    Dup: felt252,
+    Dup: felt252,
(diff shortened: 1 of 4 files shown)


orizi:
* **#9695** <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/9695?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/9695?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #9702, opened 2026-02-20
fix: remove unreachable branch in Snapshot handling

phrwlk:
## Summary

The logic is now simplified to just use `stmt.input.var_id` in the else case, which produces identical behavior

---

## Type of change

- [x] Style, wording, formatting, or typo-only change

---

## Why is this change needed?

The else-if branch at lines 265-268 was dead code. It checked for `desnaps.is_empty()` but the subsequent loop for desnap in desnaps would never execute in that case, making the assigned value of `new_var` unused. 

---


orizi (commented):
> @orizi made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/9702)* status: 0 of 1 files reviewed, all discussions resolved.

___
*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/9702#-OmDew_7C671ZPl0LFBU:-OmDew_854tPOszsJDuj:bf901qd) (no related file):*
@eytan-starkware

PR #9703, opened 2026-02-20
fix: check `values.contains_key` before `get_type_info`

Bashmunta:
## Summary

Moved the `values.contains_key()` check before `get_type_info()` call to avoid redundant lookups, matching the pattern used in `parse_circuit_inputs.`

---

## Type of change

Please check **one**:

- [x] Performance improvement

---

## Why is this change needed?

 In `get_circuit_info`, `context.get_type_info()` was called before checking if the value was already processed. Since the same gate can appear multiple times in the stack (shared inputs), this caused unnecessary lookups when the value already existed in `values`.

### DIFF
--- a/crates/cairo-lang-sierra/src/extensions/modules/circuit.rs
+++ b/crates/cairo-lang-sierra/src/extensions/modules/circuit.rs
@@ -1063,12 +1063,11 @@ fn get_circuit_info(
         .collect();
 
     while let Some((ty, first_visit)) = stack.pop() {
-        let long_id = &context.get_type_info(&ty)?.long_id;
-
         if values.contains_key(&ty) {
-            // The value was already processed.
             continue;
         }
+
+        let long_id = &context.get_type_info(&ty)?.long_id;
 
         let gate_inputs = long_id
             .generic_args


orizi (changes requested):
> @orizi made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/9703)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on Bashmunta).

___
*[`crates/cairo-lang-sierra/src/extensions/modules/circuit.rs` line 1069 at r1](https://reviewable.io/reviews/starkware-libs/cairo/9703#-OmDfQ0m1Apuh_fDseAu:-OmDfQ0m1Apuh_fDseAv:br2j20l) ([raw file](https://github.com/starkware-libs/cairo/blob/e56055c87a9db4e3dbb91c82ccb2ea751a8dc617/crates/cairo-lang-sierra/src/extensions/modules/circuit.rs#L1069)):*
> ```XML
> 
>         if values.contains_key(&ty) {
>             // The value was already processed.
> ```

why remove the doc?

PR #9691, opened 2026-02-17
fix: remove unnecessary vector clone in closure capture

Galoretka:
## Summary

Removed the `.clone()` call since the vector is moved into `StructConstruct::add` and never used after that point.

---

## Type of change

- [x] Style, wording, formatting, or typo-only change

---

## Why is this change needed?

In the capture function, the inputs vector was being cloned when passed to `StructConstruct`, even though it was not used afterwards - causing an unnecessary heap allocation.

PR #9712, opened 2026-02-26
fix: avoid clone in lower_expr_desnap by matching LoweredExpr by value

Princetimix69:
## Summary

  In `lower_expr_desnap`, when the inner expression is already a `LoweredExpr::Snapshot`, the previous code matched by reference and called `expr.as_ref().clone()` to return the inner value. This PR matches by value instead and uses `*expr` to move out of the `Box`, eliminating the clone entirely.

  ---

  ## Type of change

  - [x] Performance improvement

  ---

  ## Why is this change needed?

  `LoweredExpr` is a recursive enum with variants like `Tuple { exprs: Vec<LoweredExpr> }` and `FixedSizeArray { exprs: Vec<LoweredExpr> }`. Cloning it performs a full heap allocation and element-wise deep copy whose cost grows with nesting depth. The clone was never necessary — `expr` is not used after the early return, so ownership can be taken directly.

  ---

  `lower_expr_desnap` cloned the entire `LoweredExpr` tree on the desnap-of-snapshot fast path.

  ---

  The inner `LoweredExpr` is moved out of the `Box` with `*expr`. No allocation or copy occurs.

  ---

  Related to the pattern addressed in #9691.

  ---

  ## Additional context

  The desnap fast path (`@expr` immediately followed by desnap) is not uncommon in Cairo code. The fix is a one-line change with no semantic difference.

### DIFF
--- a/crates/cairo-lang-lowering/src/lower/mod.rs
+++ b/crates/cairo-lang-lowering/src/lower/mod.rs
@@ -1224,8 +1224,8 @@ fn lower_expr_desnap<'db>(
     log::trace!("Lowering a desnap: {:?}", expr.debug(&ctx.expr_formatter));
     let location = ctx.get_location(expr.stable_ptr.untyped());
     let expr = lower_expr(ctx, builder, expr.inner)?;
-    if let LoweredExpr::Snapshot { expr, .. } = &expr {
-        return Ok(expr.as_ref().clone());
+    if let LoweredExpr::Snapshot { expr, .. } = expr {
+        return Ok(*expr);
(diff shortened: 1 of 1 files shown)


orizi (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/9712#-:-OmPYde_5mBQ8DDplJAJ:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on Princetimix69).

PR #9713, opened 2026-02-26
perf: lazily evaluate is_of_interface in process_submodule_contract

esorense:
## Summary

  `is_of_interface` (a call to `impl_def_trait(...).has_attr(...)`) was computed unconditionally for every `impl_def` in the loop, but it is only needed when the impl has `#[external]`, `#[abi(embed_v0)]`, or `#[abi(per_item)]`. Contracts contain many auto-generated helper impls (`ContractStateDrop`, `HasComponent`, etc.) that carry none of these attributes and hit the implicit `else` branch — paying the query cost for nothing. This PR moves the computation into each branch so it is only evaluated when actually needed.

  ---

  ## Type of change

  - [x] Performance improvement

  ---

  ## Why is this change needed?

  `impl_def_trait` is a salsa query that, when not yet cached, triggers a full semantic resolution with a `Resolver`. As noted in #9699, avoiding unnecessary calls to such queries is important for compilation performance. The number of no-attribute helper impls grows with the number of embedded components, so the wasted work scales with contract complexity.

  ---

  `impl_def_trait` was called once per impl regardless of whether the result would be used.

  ---

  `impl_def_trait` is only called for impls that have a relevant ABI attribute.

  ---

  Related to #9699.

---

> [!NOTE]
> **Low Risk**
> Only adds a comment/whitespace in `AbiBuilder::process_submodule_contract`; no functional behavior changes are introduced.
> 
> **Overview**
> Adds a TODO comment in `AbiBuilder::process_submodule_contract` noting the planned deprecation of the `#[external]` attribute when processing impls for ABI generation.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 6be8192400224fcf961de31c23ebcf88794fb026. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


esorense:
> > @orizi made 1 comment.
> > _[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/9713)_ status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on esorense).
> 
> _[`crates/cairo-lang-starknet/src/abi.rs` line 233 at r1](https://reviewable.io/reviews/starkware-libs/cairo/9713#-OmPYuts4txTu7VO55-l:-OmPYuts4txTu7VO55-m:bdj096f) ([raw file](https://github.com/starkware-libs/cairo/blob/a908db53b583b4f878b2c36b1627156bb7d575da/crates/cairo-lang-starknet/src/abi.rs#L233)):_
> 
> > ```
> >             let source = Source::Impl(impl_def);
> >             // TODO(v3): deprecate the external attribute.
> >             if impl_def.has_attr(self.db, EXTERNAL_ATTR)? {
> > ```
> 
> doesn't seem to be worth it - as external attr is mostly deprecated, and now we just duplicate code for that case.

@orizi  Good point. I've refactored it to use a lazy-evaluated closure with caching — the query logic is now written once, but still only executes when an impl has a relevant attribute. This eliminates the duplication while keeping the performance benefit.

orizi (changes requested):
> @orizi made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/9713)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on esorense).

___
*[`crates/cairo-lang-starknet/src/abi.rs` line 233 at r2](https://reviewable.io/reviews/starkware-libs/cairo/9713#-OmV-xTWCck99ONe8FQ2:-OmV-xTWCck99ONe8FQ3:b-d9a51l) ([raw file](https://github.com/starkware-libs/cairo/blob/184f5e54d92e4e1aea83e328b7f0166c7f6114ae/crates/cairo-lang-starknet/src/abi.rs#L233)):*
> ```XML
>             let source = Source::Impl(impl_def);
>             let mut is_of_interface = None;
>             let check_is_of_interface = || -> Maybe<bool> {
> ```

the var is never used twice and this is a piece of needlessly complicated code for the likely zero gain from it.

PR #9819, opened 2026-04-05
fix: replace recursive expression parser with iterative stack to fix stack overflow on deep nesting

orizi:
## Summary

Refactored the expression parser to use an explicit heap-allocated frame stack instead of recursion to prevent stack overflows when parsing deeply nested expressions. The parser now uses a two-phase state machine (Atom/Reduce) with different frame types for unary operators, binary operators, parentheses, and indexing operations.

---

## Type of change

Please check **one**:

- [x] Performance improvement
- [ ] New feature

---

## Why is this change needed?

The previous recursive implementation of `try_parse_expr_limited` could cause stack overflows when parsing deeply nested expressions with many levels of unary operators, parentheses, or binary operations. This is a concrete problem that would crash the parser on valid but deeply nested Cairo code.

---

The expression parser used recursive calls to `try_parse_expr_limited`, `try_parse_atom_or_unary`, and `expect_parenthesized_expr`, which could exhaust the call stack on deeply nested expressions.

---

The parser now uses an iterative approach with explicit frame management on the heap, allowing it to handle arbitrarily deep expression nesting without stack overflow. The parsing logic and AST output remain identical, but the implementation is now stack-safe.

---

N/A

---

## Additional context

The refactoring introduces several frame types (`ExprFrame::Unary`, `ExprFrame::Binary`, `ExprFrame::Paren`, `ExprFrame::Index`) that track pending computations, and a two-phase state machine that alternates between parsing atoms and reducing completed expressions. Test cases were added to verify that deeply nested unary operators and parentheses no longer cause stack overflows.


orizi:
* **#9819** <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/9819?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/9819?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #9829, opened 2026-04-09
Skip modules without priv uses when computing unused imports diagnostics

integraledelebesgue:
## Summary

A little optimization for further gain of diagnostic performance in CairoLS

---

## Type of change

Please check **one**:

- [ ] Bug fix (fixes incorrect behavior)
- [ ] New feature
- [x] Performance improvement
- [ ] Documentation change with concrete technical impact
- [ ] Style, wording, formatting, or typo-only change

PR #9828, opened 2026-04-09
Optimize diagnostics flattening in `get_all`

integraledelebesgue:
## Summary

Collecting the diagnostic trees into a pre-allocated vector significantly reduces the amount of intermediate allocations.
Results in a huge gain of performance in CairoLS.

---

## Type of change

Please check **one**:

- [ ] Bug fix (fixes incorrect behavior)
- [ ] New feature
- [x] Performance improvement
- [ ] Documentation change with concrete technical impact
- [ ] Style, wording, formatting, or typo-only change

PR #9830, opened 2026-04-09
Replace singleton file overrides with granular inputs and views

integraledelebesgue:
## Summary

A new architecture of file inputs, which significantly improves the memory performance of CairoLS.

## Changes
- File overrides don't exist anymore. Instead, all file contents (physical and generated) are stored in a shared map
- The shared map is registered as a Salsa View, which allows for granular, incremental computation for single files

---

## Type of change

Please check **one**:

- [ ] New feature
- [x] Performance improvement

### DIFF
--- a/crates/cairo-lang-compiler/src/db.rs
+++ b/crates/cairo-lang-compiler/src/db.rs
@@ -89,8 +89,7 @@ impl CloneableDatabase for RootDatabase {
 
 impl RootDatabase {
     fn new(default_plugin_suite: PluginSuite, optimizations: Optimizations) -> Self {
-        let mut res =
-            Self { storage: Default::default(), file_contents: Default::default() };
+        let mut res = Self { storage: Default::default(), file_contents: Default::default() };
         register_files_group_view(&res);
         init_external_files(&mut res);
         init_files_group(&mut res);
--- a/crates/cairo-lang-compiler/src/db.rs
+++ b/crates/cairo-lang-compiler/src/db.rs
@@ -1,3 +1,5 @@
+use std::sync::{Arc, RwLock};
+
 use anyhow::{Result, anyhow, bail};
 use cairo_lang_defs::db::{init_defs_group, init_external_files};
 use cairo_lang_diagnostics::Maybe;
@@ -77,8 +79,8 @@ pub struct RootDatabase {
 #[salsa::db]
 impl salsa::Database for RootDatabase {}
 impl FileContentView for RootDatabase {
-    fn file_content_storage(&self) -> Option<&FileContentStorage> {
-        Some(&self.file_contents)
+    fn file_content_storage(&self) -> &FileContentStorage {
+        &self.file_contents
     }
 }
 impl CloneableDatabase for RootDatabase {
@@ -113,7 +115,8 @@ impl RootDatabase {
 
     /// Snapshots the db for read only.
     pub fn snapshot(&self) -> RootDatabase {
-        RootDatabase { storage: self.storage.clone(), file_contents: self.file_contents.clone() }
+        let file_contents = Arc::new(RwLock::new(self.file_contents.read().unwrap().clone()));
+        RootDatabase { storage: self.storage.clone(), file_contents }
     }
 }
 
--- a/crates/cairo-lang-compiler/src/db.rs
+++ b/crates/cairo-lang-compiler/src/db.rs
@@ -2,7 +2,10 @@ use anyhow::{Result, anyhow, bail};
 use cairo_lang_defs::db::{init_defs_group, init_external_files};
 use cairo_lang_diagnostics::Maybe;
 use cairo_lang_filesystem::cfg::CfgSet;
-use cairo_lang_filesystem::db::{CORELIB_VERSION, FilesGroup, init_dev_corelib, init_files_group};
+use cairo_lang_filesystem::db::{
+    CORELIB_VERSION, FileContentStorage, FileContentView, FilesGroup, init_dev_corelib,
+    init_files_group, new_file_content_storage, register_files_group_view,
+};
 use cairo_lang_filesystem::detect::detect_corelib;
 use cairo_lang_filesystem::flag::{Flag, FlagsGroup};
 use cairo_lang_filesystem::ids::{CrateId, FlagLongId};
@@ -69,9 +72,15 @@ fn estimate_code_size(
 #[derive(Clone)]
 pub struct RootDatabase {
     storage: salsa::Storage<RootDatabase>,
+    file_contents: FileContentStorage,
 }
 #[salsa::db]
 impl salsa::Database for RootDatabase {}
+impl FileContentView for RootDatabase {
+    fn file_content_storage(&self) -> Option<&FileContentStorage> {
+        Some(&self.file_contents)
+    }
+}
 impl CloneableDatabase for RootDatabase {
     fn dyn_clone(&self) -> Box<dyn CloneableDatabase> {
         Box::new(self.clone())
@@ -80,7 +89,9 @@ impl CloneableDatabase for RootDatabase {
 
 impl RootDatabase {
     fn new(default_plugin_suite: PluginSuite, optimizations: Optimizations) -> Self {
-        let mut res = Self { storage: Default::default() };
+        let mut res =
+            Self { storage: Default::default(), file_contents: new_file_content_storage() };
+        register_files_group_view(&res);
         init_external_files(&mut res);
         init_files_group(&mut res);
         init_lowering_group(&mut res, optimizations, Some(estimate_code_size));
@@ -103,7 +114,7 @@ impl RootDatabase {
 
     /// Snapshots the db for read only.
     pub fn snapshot(&self) -> RootDatabase {
-        RootDatabase { storage: self.storage.clone() }
+        RootDatabase { storage: self.storage.clone(), file_contents: self.file_contents.clone() }
     }
 }
 
--- a/crates/cairo-lang-compiler/src/db.rs
+++ b/crates/cairo-lang-compiler/src/db.rs
@@ -4,7 +4,7 @@ use cairo_lang_diagnostics::Maybe;
 use cairo_lang_filesystem::cfg::CfgSet;
 use cairo_lang_filesystem::db::{
     CORELIB_VERSION, FileContentStorage, FileContentView, FilesGroup, init_dev_corelib,
-    init_files_group, new_file_content_storage, register_files_group_view,
+    init_files_group, register_files_group_view,
 };
 use cairo_lang_filesystem::detect::detect_corelib;
 use cairo_lang_filesystem::flag::{Flag, FlagsGroup};
@@ -90,7 +90,7 @@ impl CloneableDatabase for RootDatabase {
 impl RootDatabase {
     fn new(default_plugin_suite: PluginSuite, optimizations: Optimizations) -> Self {
         let mut res =
-            Self { storage: Default::default(), file_contents: new_file_content_storage() };
+            Self { storage: Default::default(), file_contents: Default::default() };
         register_files_group_view(&res);
         init_external_files(&mut res);
         init_files_group(&mut res);
--- a/crates/cairo-lang-compiler/src/db.rs
+++ b/crates/cairo-lang-compiler/src/db.rs
@@ -114,6 +114,13 @@ impl RootDatabase {
     }
 
     /// Snapshots the db for read only.
+    ///
+    /// [`FileContentStorage`] is deep-cloned into a fresh `Arc<RwLock<...>>` rather than
+    /// Arc-shared, so the snapshot is fully isolated from subsequent file-content changes to the
+    /// original. The `Arc` and `RwLock` wrappers are kept because [`FileContentStorage`] always
+    /// carries them: a live database may be shallow-cloned for parallel compilation workers (via
+    /// [`CloneableDatabase`]), and those clones share the same `Arc`; the `RwLock` provides the
+    /// thread-safe concurrent access required across those shared clones.
     pub fn snapshot(&self) -> RootDatabase {
         let file_contents = Arc::new(RwLock::new(self.file_contents.read().unwrap().clone()));
         RootDatabase { storage: self.storage.clone(), file_contents }
--- a/crates/cairo-lang-compiler/src/project.rs
+++ b/crates/cairo-lang-compiler/src/project.rs
@@ -56,27 +56,26 @@ pub fn setup_single_file_project(
         Ok(crate_id.long(db).clone().into_crate_input(db))
     } else {
         // If file_stem is not lib, create a fake lib file.
-        {
-            let crate_id = CrateId::plain(db, SmolStrId::from(db, file_stem));
-            set_crate_config!(
-                db,
-                crate_id,
-                Some(CrateConfiguration::default_for_root(Directory::Real(file_dir.to_path_buf())))
-            );
-        }
-        let file_input = {
+        let crate_id = CrateId::plain(db, SmolStrId::from(db, file_stem));
+        set_crate_config!(
+            db,
+            crate_id,
+            Some(CrateConfiguration::default_for_root(Directory::Real(file_dir.to_path_buf())))
+        );
+        let (file_input, crate_input) = {
             let crate_id = CrateId::plain(db, SmolStrId::from(db, file_stem));
             let module_id = ModuleId::CrateRoot(crate_id);
             let file_id = db.module_main_file(module_id).unwrap();
-            db.file_input(file_id).clone()
+            let file_input = db.file_input(file_id).clone();
+            let crate_input = crate_id.long(db).clone().into_crate_input(db);
+            (file_input, crate_input)
         };
         set_generated_file_content_for_input(
             db,
             file_input,
             Some(format!("mod {file_stem};").into()),
         );
-        let crate_id = CrateId::plain(db, SmolStrId::from(db, file_stem));
-        Ok(crate_id.long(db).clone().into_crate_input(db))
+        Ok(crate_input)
     }
 }
 
--- a/crates/cairo-lang-compiler/src/project.rs
+++ b/crates/cairo-lang-compiler/src/project.rs
@@ -5,9 +5,10 @@ use cairo_lang_defs::db::DefsGroup;
 use cairo_lang_defs::ids::ModuleId;
 use cairo_lang_filesystem::db::{
     CORELIB_CRATE_NAME, CrateConfiguration, CrateIdentifier, CrateSettings, FilesGroup,
+    set_generated_file_content_for_input,
 };
 use cairo_lang_filesystem::ids::{CrateId, CrateInput, CrateLongId, Directory, SmolStrId};
-use cairo_lang_filesystem::{override_file_content, set_crate_config};
+use cairo_lang_filesystem::set_crate_config;
 pub use cairo_lang_project::*;
 use cairo_lang_utils::Intern;
 use salsa::Database;
@@ -55,16 +56,25 @@ pub fn setup_single_file_project(
         Ok(crate_id.long(db).clone().into_crate_input(db))
     } else {
         // If file_stem is not lib, create a fake lib file.
-        let crate_id = CrateId::plain(db, SmolStrId::from(db, file_stem));
-        set_crate_config!(
+        {
+            let crate_id = CrateId::plain(db, SmolStrId::from(db, file_stem));
+            set_crate_config!(
+                db,
+                crate_id,
+                Some(CrateConfiguration::default_for_root(Directory::Real(file_dir.to_path_buf())))
+            );
+        }
+        let file_input = {
+            let crate_id = CrateId::plain(db, SmolStrId::from(db, file_stem));
+            let module_id = ModuleId::CrateRoot(crate_id);
+            let file_id = db.module_main_file(module_id).unwrap();
+            db.file_input(file_id).clone()
+        };
+        set_generated_file_content_for_input(
             db,
-            crate_id,
-            Some(CrateConfiguration::default_for_root(Directory::Real(file_dir.to_path_buf())))
+            file_input,
+            Some(format!("mod {file_stem};").into()),
         );
-        let crate_id = CrateId::plain(db, SmolStrId::from(db, file_stem));
-        let module_id = ModuleId::CrateRoot(crate_id);
-        let file_id = db.module_main_file(module_id).unwrap();
-        override_file_content!(db, file_id, Some(format!("mod {file_stem};").into()));
         let crate_id = CrateId::plain(db, SmolStrId::from(db, file_stem));
         Ok(crate_id.long(db).clone().into_crate_input(db))
     }
--- a/crates/cairo-lang-defs/src/test.rs
+++ b/crates/cairo-lang-defs/src/test.rs
@@ -43,8 +43,7 @@ impl FileContentView for DatabaseForTesting {
 
 impl Default for DatabaseForTesting {
     fn default() -> Self {
-        let mut res =
-            Self { storage: Default::default(), file_contents: Default::default() };
+        let mut res = Self { storage: Default::default(), file_contents: Default::default() };
         register_files_group_view(&res);
         init_external_files(&mut res);
         init_files_group(&mut res);
--- a/crates/cairo-lang-defs/src/test.rs
+++ b/crates/cairo-lang-defs/src/test.rs
@@ -4,7 +4,7 @@ use std::sync::Arc;
 use cairo_lang_debug::debug::DebugWithDb;
 use cairo_lang_filesystem::db::{
     CrateConfiguration, FileContentStorage, FileContentView, FilesGroup, init_files_group,
-    new_file_content_storage, register_files_group_view, set_on_disk_file_content_for_input,
+    override_file_content_for_input, register_files_group_view,
 };
 use cairo_lang_filesystem::ids::{CrateId, Directory, FileLongId, SmolStrId};
 use cairo_lang_filesystem::set_crate_config;
@@ -44,7 +44,7 @@ impl FileContentView for DatabaseForTesting {
 impl Default for DatabaseForTesting {
     fn default() -> Self {
         let mut res =
-            Self { storage: Default::default(), file_contents: new_file_content_storage() };
+            Self { storage: Default::default(), file_contents: Default::default() };
         register_files_group_view(&res);
         init_external_files(&mut res);
         init_files_group(&mut res);
@@ -125,7 +125,7 @@ pub fn setup_test_module(db: &mut dyn Database, content: &str) {
         let file = db.module_main_file(ModuleId::CrateRoot(crate_id)).unwrap();
         db.file_input(file).clone()
     };
-    set_on_disk_file_content_for_input(db, file_input, Some(content.into()));
+    override_file_content_for_input(db, file_input, Some(content.into()));
     let crate_id = get_crate_id(db);
     let file = db.module_main_file(ModuleId::CrateRoot(crate_id)).unwrap();
     let syntax_diagnostics = db.file_syntax_diagnostics(file).format(db);
@@ -166,7 +166,7 @@ macro_rules! set_file_content {
             let file_id = FileLongId::OnDisk($path.into()).intern($db);
             $db.file_input(file_id).clone()
         };
-        set_on_disk_file_content_for_input($db, file_input, Some($content.into()));
+        override_file_content_for_input($db, file_input, Some($content.into()));
     };
 }
 
--- a/crates/cairo-lang-defs/src/test.rs
+++ b/crates/cairo-lang-defs/src/test.rs
@@ -36,8 +36,8 @@ pub struct DatabaseForTesting {
 #[salsa::db]
 impl salsa::Database for DatabaseForTesting {}
 impl FileContentView for DatabaseForTesting {
-    fn file_content_storage(&self) -> Option<&FileContentStorage> {
-        Some(&self.file_contents)
+    fn file_content_storage(&self) -> &FileContentStorage {
+        &self.file_contents
     }
 }
 
--- a/crates/cairo-lang-defs/src/test.rs
+++ b/crates/cairo-lang-defs/src/test.rs
@@ -2,9 +2,12 @@ use std::fmt::Write as _;
 use std::sync::Arc;
 
 use cairo_lang_debug::debug::DebugWithDb;
-use cairo_lang_filesystem::db::{CrateConfiguration, FilesGroup, init_files_group};
+use cairo_lang_filesystem::db::{
+    CrateConfiguration, FileContentStorage, FileContentView, FilesGroup, init_files_group,
+    new_file_content_storage, register_files_group_view, set_on_disk_file_content_for_input,
+};
 use cairo_lang_filesystem::ids::{CrateId, Directory, FileLongId, SmolStrId};
-use cairo_lang_filesystem::{override_file_content, set_crate_config};
+use cairo_lang_filesystem::set_crate_config;
 use cairo_lang_parser::db::ParserGroup;
 use cairo_lang_syntax::node::helpers::QueryAttrs;
 use cairo_lang_syntax::node::kind::SyntaxKind;
@@ -28,13 +31,21 @@ use crate::plugin::{
 #[derive(Clone)]
 pub struct DatabaseForTesting {
     storage: salsa::Storage<DatabaseForTesting>,
+    file_contents: FileContentStorage,
 }
 #[salsa::db]
 impl salsa::Database for DatabaseForTesting {}
+impl FileContentView for DatabaseForTesting {
+    fn file_content_storage(&self) -> Option<&FileContentStorage> {
+        Some(&self.file_contents)
+    }
+}
 
 impl Default for DatabaseForTesting {
     fn default() -> Self {
-        let mut res = Self { storage: Default::default() };
+        let mut res =
+            Self { storage: Default::default(), file_contents: new_file_content_storage() };
+        register_files_group_view(&res);
         init_external_files(&mut res);
         init_files_group(&mut res);
         init_defs_group(&mut res);
@@ -109,9 +120,12 @@ pub fn setup_test_module(db: &mut dyn Database, content: &str) {
     let crate_id = get_crate_id(db);
     let directory = Directory::Real("src".into());
     set_crate_config!(db, crate_id, Some(CrateConfiguration::default_for_root(directory)));
-    let crate_id = get_crate_id(db);
-    let file = db.module_main_file(ModuleId::CrateRoot(crate_id)).unwrap();
-    override_file_content!(db, file, Some(content.into()));
+    let file_input = {
+        let crate_id = get_crate_id(db);
+        let file = db.module_main_file(ModuleId::CrateRoot(crate_id)).unwrap();
+        db.file_input(file).clone()
+    };
+    set_on_disk_file_content_for_input(db, file_input, Some(content.into()));
     let crate_id = get_crate_id(db);
     let file = db.module_main_file(ModuleId::CrateRoot(crate_id)).unwrap();
     let syntax_diagnostics = db.file_syntax_diagnostics(file).format(db);
@@ -148,8 +162,11 @@ fn test_module_file() {
 
 macro_rules! set_file_content {
     ($db:expr, $path:expr, $content:expr) => {
-        let file_id = FileLongId::OnDisk($path.into()).intern($db);
-        override_file_content!($db, file_id, Some($content.into()));
+        let file_input = {
+            let file_id = FileLongId::OnDisk($path.into()).intern($db);
+            $db.file_input(file_id).clone()
+        };
+        set_on_disk_file_content_for_input($db, file_input, Some($content.into()));
     };
 }
 
--- a/crates/cairo-lang-doc/src/tests/test_utils.rs
+++ b/crates/cairo-lang-doc/src/tests/test_utils.rs
@@ -3,8 +3,7 @@ use cairo_lang_defs::db::{DefsGroup, init_defs_group};
 use cairo_lang_defs::ids::ModuleId;
 use cairo_lang_filesystem::db::{
     CrateConfiguration, FileContentStorage, FileContentView, FilesGroup, init_dev_corelib,
-    init_files_group, new_file_content_storage, register_files_group_view,
-    set_on_disk_file_content_for_input,
+    init_files_group, override_file_content_for_input, register_files_group_view,
 };
 use cairo_lang_filesystem::detect::detect_corelib;
 use cairo_lang_filesystem::ids::{CrateId, Directory, FileLongId, SmolStrId};
@@ -32,7 +31,7 @@ impl FileContentView for TestDatabase {
 impl Default for TestDatabase {
     fn default() -> Self {
         let mut res =
-            Self { storage: Default::default(), file_contents: new_file_content_storage() };
+            Self { storage: Default::default(), file_contents: Default::default() };
         register_files_group_view(&res);
         init_files_group(&mut res);
         init_defs_group(&mut res);
@@ -64,7 +63,7 @@ pub fn setup_test_module(db: &mut dyn Database, content: &str) {
         let file = db.module_main_file(ModuleId::CrateRoot(crate_id)).unwrap();
         db.file_input(file).clone()
     };
-    set_on_disk_file_content_for_input(db, file_input, Some(content.into()));
+    override_file_content_for_input(db, file_input, Some(content.into()));
     let crate_id = test_crate_id(db);
     let file = db.module_main_file(ModuleId::CrateRoot(crate_id)).unwrap();
     let syntax_diagnostics = db.file_syntax_diagnostics(file).format(db);
@@ -84,7 +83,7 @@ pub fn setup_test_module_without_syntax_diagnostics(db: &mut dyn Database, conte
         let file = db.module_main_file(ModuleId::CrateRoot(crate_id)).unwrap();
         db.file_input(file).clone()
     };
-    set_on_disk_file_content_for_input(db, file_input, Some(content.into()));
+    override_file_content_for_input(db, file_input, Some(content.into()));
 }
 
 pub fn set_file_content(db: &mut TestDatabase, path: &str, content: &str) {
@@ -92,5 +91,5 @@ pub fn set_file_content(db: &mut TestDatabase, path: &str, content: &str) {
         let file_id = FileLongId::OnDisk(path.into()).intern(db);
         db.file_input(file_id).clone()
     };
-    set_on_disk_file_content_for_input(db, file_input, Some(content.into()));
+    override_file_content_for_input(db, file_input, Some(content.into()));
 }
--- a/crates/cairo-lang-doc/src/tests/test_utils.rs
+++ b/crates/cairo-lang-doc/src/tests/test_utils.rs
@@ -23,8 +23,8 @@ pub struct TestDatabase {
 #[salsa::db]
 impl salsa::Database for TestDatabase {}
 impl FileContentView for TestDatabase {
-    fn file_content_storage(&self) -> Option<&FileContentStorage> {
-        Some(&self.file_contents)
+    fn file_content_storage(&self) -> &FileContentStorage {
+        &self.file_contents
     }
 }
 
--- a/crates/cairo-lang-doc/src/tests/test_utils.rs
+++ b/crates/cairo-lang-doc/src/tests/test_utils.rs
@@ -30,8 +30,7 @@ impl FileContentView for TestDatabase {
 
 impl Default for TestDatabase {
     fn default() -> Self {
-        let mut res =
-            Self { storage: Default::default(), file_contents: Default::default() };
+        let mut res = Self { storage: Default::default(), file_contents: Default::default() };
         register_files_group_view(&res);
         init_files_group(&mut res);
         init_defs_group(&mut res);
--- a/crates/cairo-lang-doc/src/tests/test_utils.rs
+++ b/crates/cairo-lang-doc/src/tests/test_utils.rs
@@ -2,11 +2,13 @@ use anyhow::{Result, anyhow};
 use cairo_lang_defs::db::{DefsGroup, init_defs_group};
 use cairo_lang_defs::ids::ModuleId;
 use cairo_lang_filesystem::db::{
-    CrateConfiguration, FilesGroup, init_dev_corelib, init_files_group,
+    CrateConfiguration, FileContentStorage, FileContentView, FilesGroup, init_dev_corelib,
+    init_files_group, new_file_content_storage, register_files_group_view,
+    set_on_disk_file_content_for_input,
 };
 use cairo_lang_filesystem::detect::detect_corelib;
 use cairo_lang_filesystem::ids::{CrateId, Directory, FileLongId, SmolStrId};
-use cairo_lang_filesystem::{override_file_content, set_crate_config};
+use cairo_lang_filesystem::set_crate_config;
 use cairo_lang_parser::db::ParserGroup;
 use cairo_lang_semantic::db::{PluginSuiteInput, init_semantic_group};
 use cairo_lang_semantic::plugin::PluginSuite;
@@ -17,13 +19,21 @@ use salsa::Database;
 #[derive(Clone)]
 pub struct TestDatabase {
     storage: salsa::Storage<TestDatabase>,
+    file_contents: FileContentStorage,
 }
 #[salsa::db]
 impl salsa::Database for TestDatabase {}
+impl FileContentView for TestDatabase {
+    fn file_content_storage(&self) -> Option<&FileContentStorage> {
+        Some(&self.file_contents)
+    }
+}
 
 impl Default for TestDatabase {
     fn default() -> Self {
-        let mut res = Self { storage: Default::default() };
+        let mut res =
+            Self { storage: Default::default(), file_contents: new_file_content_storage() };
+        register_files_group_view(&res);
         init_files_group(&mut res);
         init_defs_group(&mut res);
         init_semantic_group(&mut res);
@@ -49,9 +59,12 @@ pub fn setup_test_module(db: &mut dyn Database, content: &str) {
     let crate_id = test_crate_id(db);
     let directory = Directory::Real("src".into());
     set_crate_config!(db, crate_id, Some(CrateConfiguration::default_for_root(directory)));
-    let crate_id = test_crate_id(db);
-    let file = db.module_main_file(ModuleId::CrateRoot(crate_id)).unwrap();
-    override_file_content!(db, file, Some(content.into()));
+    let file_input = {
+        let crate_id = test_crate_id(db);
+        let file = db.module_main_file(ModuleId::CrateRoot(crate_id)).unwrap();
+        db.file_input(file).clone()
+    };
+    set_on_disk_file_content_for_input(db, file_input, Some(content.into()));
     let crate_id = test_crate_id(db);
     let file = db.module_main_file(ModuleId::CrateRoot(crate_id)).unwrap();
     let syntax_diagnostics = db.file_syntax_diagnostics(file).format(db);
@@ -66,12 +79,18 @@ pub fn setup_test_module_without_syntax_diagnostics(db: &mut dyn Database, conte
     let crate_id = test_crate_id(db);
     let directory = Directory::Real("src".into());
     set_crate_config!(db, crate_id, Some(CrateConfiguration::default_for_root(directory)));
-    let crate_id = test_crate_id(db);
-    let file = db.module_main_file(ModuleId::CrateRoot(crate_id)).unwrap();
-    override_file_content!(db, file, Some(content.into()));
+    let file_input = {
+        let crate_id = test_crate_id(db);
+        let file = db.module_main_file(ModuleId::CrateRoot(crate_id)).unwrap();
+        db.file_input(file).clone()
+    };
+    set_on_disk_file_content_for_input(db, file_input, Some(content.into()));
 }
 
 pub fn set_file_content(db: &mut TestDatabase, path: &str, content: &str) {
-    let file_id = FileLongId::OnDisk(path.into()).intern(db);
-    override_file_content!(db, file_id, Some(content.into()));
+    let file_input = {
+        let file_id = FileLongId::OnDisk(path.into()).intern(db);
+        db.file_input(file_id).clone()
+    };
+    set_on_disk_file_content_for_input(db, file_input, Some(content.into()));
 }
--- a/crates/cairo-lang-filesystem/src/db.rs
+++ b/crates/cairo-lang-filesystem/src/db.rs
@@ -234,8 +234,10 @@ pub struct FilesGroupInput {
     /// Main input of the project. Lists all the crates configurations.
     #[returns(ref)]
     pub crate_configs: Option<OrderedHashMap<CrateInput, CrateConfigurationInput>>,
-    /// Structural revision for databases that source file content from granular per-file slots
-    /// instead of only the raw-file path.
+    /// Structural revision bumped each time a new [`FileContents`] handle is first created for a
+    /// file. The [`file_content`] tracked function reads this field so that it re-executes when a
+    /// previously-unregistered file gets a handle — at which point it records the per-file
+    /// dependency for future fine-grained incremental updates.
     pub file_contents_revision: u64,
     // TODO(yuval): consider moving this to a separate crate, or rename this crate.
     /// The compilation flags.
@@ -253,21 +255,16 @@ pub fn files_group_input(db: &dyn Database) -> FilesGroupInput {
     FilesGroupInput::new(db, None, 0, None, None, None)
 }
 
-/// Per-file content slots used by long-running tools (like CairoLS) that need granular
-/// invalidation.
+/// Per-file content slot used by long-running tools (like CairoLS) that need granular
+/// invalidation. A file is either an on-disk override or a generated file — never both — so a
+/// single field is sufficient.
 #[salsa::input]
 pub struct FileContents {
     #[returns(ref)]
-    pub on_disk_content: Option<ArcStr>,
-    #[returns(ref)]
-    pub generated_content: Option<ArcStr>,
+    pub content: Option<ArcStr>,
 }
 
 pub type FileContentStorage = Arc<RwLock<HashMap<FileInput, FileContents>>>;
-
-pub fn new_file_content_storage() -> FileContentStorage {
-    Default::default()
-}
 
 /// Database view used by [`FilesGroup::file_content`] to retrieve per-file editor/generated
 /// content.
@@ -280,29 +277,11 @@ pub trait FileContentView: Database {
         self.file_content_storage()?.read().unwrap().get(file_input).copied()
     }
 
-    /// Returns a content handle for the given file, if one exists.
-    fn file_contents<'db>(&'db self, file_id: FileId<'db>) -> Option<FileContents> {
+    /// Returns the overridden content for the given file, if one exists.
+    fn file_content_override<'db>(&'db self, file_id: FileId<'db>) -> Option<&'db ArcStr> {
         let file_input = self.file_input(file_id).clone();
-        self.file_contents_for_input(&file_input)
+        self.file_contents_for_input(&file_input)?.content(self).as_ref()
     }
-
-    /// Returns on-disk file content, if one exists.
-    fn on_disk_file_content<'db>(&'db self, file_id: FileId<'db>) -> Option<&'db ArcStr> {
-        self.file_contents(file_id).and_then(|contents| contents.on_disk_content(self).as_ref())
-    }
-
-    /// Returns generated file content, if one exists.
-    fn generated_file_content<'db>(&'db self, file_id: FileId<'db>) -> Option<&'db ArcStr> {
-        self.file_contents(file_id).and_then(|contents| contents.generated_content(self).as_ref())
-    }
-}
-
-fn cast_file_content_view<Db: FileContentView + 'static>(
-    db: NonNull<Db>,
-) -> NonNull<dyn FileContentView> {
-    // SAFETY: `db` points to a live `Db` of the registered concrete database type.
-    let db_ref = unsafe { db.as_ref() };
-    NonNull::from(db_ref as &dyn FileContentView)
 }
 
 fn file_content_view(db: &dyn Database) -> &dyn FileContentView {
@@ -322,7 +301,10 @@ pub fn register_files_group_view<Db>(db: &Db)
 where
     Db: Database + FileContentView + 'static,
 {
-    views(db).add::<Db, dyn FileContentView>(cast_file_content_view::<Db>);
+    views(db).add::<Db, dyn FileContentView>(|db: NonNull<Db>| {
+        // SAFETY: `db` points to a live `Db` of the registered concrete database type.
+        NonNull::from(unsafe { db.as_ref() } as &dyn FileContentView)
+    });
 }
 
 /// Queries over the files group.
@@ -496,10 +478,6 @@ macro_rules! set_crate_config {
     };
 }
 
-fn file_contents_storage(db: &dyn Database) -> &FileContentStorage {
-    file_content_view(db).file_content_storage().expect("file content storage is not registered")
-}
-
 fn ensure_file_contents_handle_for_input(
     db: &mut dyn Database,
     file_input: FileInput,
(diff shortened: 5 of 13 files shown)


integraledelebesgue:
This change is part of the following stack:

- #9828
    - #9829
        - #9830 ◀
            - #9831
                - #9835

<sub>Change managed by [git-spice](https://abhinav.github.io/git-spice/).</sub>

piotmag769 (changes requested):
> @piotmag769 reviewed 13 files and all commit messages, and made 7 comments.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/9830)* status: all files reviewed, 7 unresolved discussions (waiting on integraledelebesgue and orizi).

___
*[`crates/cairo-lang-filesystem/src/db.rs` line 237 at r2](https://reviewable.io/reviews/starkware-libs/cairo/9830#-Oq5jton8HzHu-AnpBPE:-Oq5jton8HzHu-AnpBPF:b-ak9mo6) ([raw file](https://github.com/starkware-libs/cairo/blob/c93308053a7fd8c6c3ea9b4a703a4857553c02b7/crates/cairo-lang-filesystem/src/db.rs#L237)):*
> ```XML
>     #[returns(ref)]
>     pub crate_configs: Option<OrderedHashMap<CrateInput, CrateConfigurationInput>>,
>     /// Structural revision for databases that source file content from granular per-file slots
> ```

Doesn't explain why we need this
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 268 at r2](https://reviewable.io/reviews/starkware-libs/cairo/9830#-Oq5gPlm0r_T690FUodu:-Oq5gPlm0r_T690FUodv:bigbmav) ([raw file](https://github.com/starkware-libs/cairo/blob/c93308053a7fd8c6c3ea9b4a703a4857553c02b7/crates/cairo-lang-filesystem/src/db.rs#L268)):*
> ```XML
> pub type FileContentStorage = Arc<RwLock<HashMap<FileInput, FileContents>>>;
> 
> pub fn new_file_content_storage() -> FileContentStorage {
> ```

This is useless
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 259 at r2](https://reviewable.io/reviews/starkware-libs/cairo/9830#-Oq5gY-h6Oigvs2zUuZx:-Oq5gY-i2ymYyJVrIqWD:b-x9wi1m) ([raw file](https://github.com/starkware-libs/cairo/blob/c93308053a7fd8c6c3ea9b4a703a4857553c02b7/crates/cairo-lang-filesystem/src/db.rs#L259)):*
> ```XML
> /// invalidation.
> #[salsa::input]
> pub struct FileContents {
> ```

Do we need two fields? Only one is set at a time anyways, the other has to be `None`
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 499 at r2](https://reviewable.io/reviews/starkware-libs/cairo/9830#-Oq5iTEo2ddHgHc3WKv-:-Oq5iTEo2ddHgHc3WKv0:bukk0la) ([raw file](https://github.com/starkware-libs/cairo/blob/c93308053a7fd8c6c3ea9b4a703a4857553c02b7/crates/cairo-lang-filesystem/src/db.rs#L499)):*
> ```XML
> }
> 
> fn file_contents_storage(db: &dyn Database) -> &FileContentStorage {
> ```

Inline
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 555 at r2](https://reviewable.io/reviews/starkware-libs/cairo/9830#-Oq5ji6DCjFj3WYYo578:-Oq5ji6DCjFj3WYYo579:b4g8xea) ([raw file](https://github.com/starkware-libs/cairo/blob/c93308053a7fd8c6c3ea9b4a703a4857553c02b7/crates/cairo-lang-filesystem/src/db.rs#L555)):*
> ```XML
> pub type FileContentSnapshot = OrderedHashMap<FileInput, (Option<Arc<str>>, Option<Arc<str>>)>;
> 
> pub fn snapshot_file_contents(db: &dyn Database) -> FileContentSnapshot {
> ```

Where is it needed? Also doc
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 300 at r2](https://reviewable.io/reviews/starkware-libs/cairo/9830#-Oq5h7cgAK2SeIZFZmVU:-Oq5h7cgAK2SeIZFZmVV:bl0wm2v) ([raw file](https://github.com/starkware-libs/cairo/blob/c93308053a7fd8c6c3ea9b4a703a4857553c02b7/crates/cairo-lang-filesystem/src/db.rs#L300)):*
> ```XML
> }
> 
> fn cast_file_content_view<Db: FileContentView + 'static>(
> ```

Put this next to `register_files_group_view` or preferably declare inside it
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 657 at r2](https://reviewable.io/reviews/starkware-libs/cairo/9830#-Oq5iYHgEWyJzSrKXGG7:-Oq5iYHgEWyJzSrKXGG8:b3z43rj) ([raw file](https://github.com/starkware-libs/cairo/blob/c93308053a7fd8c6c3ea9b4a703a4857553c02b7/crates/cairo-lang-filesystem/src/db.rs#L657)):*
> ```XML
> #[salsa::tracked(returns(ref))]
> fn file_content<'db>(db: &'db dyn Database, file_id: FileId<'db>) -> Option<Arc<str>> {
>     let _ = files_group_input(db).file_contents_revision(db);
> ```

Please doc why

orizi (changes requested):
> @orizi made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/9830)* status: all files reviewed, 8 unresolved discussions (waiting on integraledelebesgue).

___
*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/9830#-Oq9lULx0io4g8Wzxr9e:-Oq9lULx0io4g8Wzxr9f:b-lk8ypy) (no related file):*
without a full documentation of what you are trying to do it and how and why it helps - would definitely not approve the PR.

Arcticae (changes requested):
> @Arcticae reviewed 9 files and all commit messages, and made 6 comments.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/9830)* status: all files reviewed, 12 unresolved discussions (waiting on integraledelebesgue and piotmag769).

___
*[`crates/cairo-lang-filesystem/src/db.rs` line 268 at r2](https://reviewable.io/reviews/starkware-libs/cairo/9830#-Oq5gPlm0r_T690FUodu:-OqAOOV04Ux7eZtKbhkN:bvbuu6b) ([raw file](https://github.com/starkware-libs/cairo/blob/c93308053a7fd8c6c3ea9b4a703a4857553c02b7/crates/cairo-lang-filesystem/src/db.rs#L268)):*
<details><summary><i>Previously, piotmag769 (Piotr Magiera) wrote…</i></summary><blockquote>

This is useless
</blockquote></details>

Also `Default` should be derivable here 
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 657 at r2](https://reviewable.io/reviews/starkware-libs/cairo/9830#-Oq5iYHgEWyJzSrKXGG7:-OqBNpzgAmM3_OsdwEhB:bvabe6u) ([raw file](https://github.com/starkware-libs/cairo/blob/c93308053a7fd8c6c3ea9b4a703a4857553c02b7/crates/cairo-lang-filesystem/src/db.rs#L657)):*
<details><summary><i>Previously, piotmag769 (Piotr Magiera) wrote…</i></summary><blockquote>

Please doc why
</blockquote></details>

Mechanism of invalidation should be described here (on fist file edit vs later edits)
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 274 at r2](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OqBO5PAApXpCDcHtS04:-OqBO5PAApXpCDcHtS05:b-x2x09t) ([raw file](https://github.com/starkware-libs/cairo/blob/c93308053a7fd8c6c3ea9b4a703a4857553c02b7/crates/cairo-lang-filesystem/src/db.rs#L274)):*
> ```XML
> /// Database view used by [`FilesGroup::file_content`] to retrieve per-file editor/generated
> /// content.
> pub trait FileContentView: Database {
> ```

Describe why a view is used
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 517 at r2](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OqBNkLp3WM7aXSWp9eo:-OqBNkLp3WM7aXSWp9ep:bj0gwpc) ([raw file](https://github.com/starkware-libs/cairo/blob/c93308053a7fd8c6c3ea9b4a703a4857553c02b7/crates/cairo-lang-filesystem/src/db.rs#L517)):*
> ```XML
> }
> 
> pub fn set_on_disk_file_content(
> ```

This one is unused i think
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 526 at r2](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OqBNySr7KuXctm6mqSs:-OqBNySr7KuXctm6mqSt:b-x1ipwl) ([raw file](https://github.com/starkware-libs/cairo/blob/c93308053a7fd8c6c3ea9b4a703a4857553c02b7/crates/cairo-lang-filesystem/src/db.rs#L526)):*
> ```XML
> }
> 
> pub fn set_on_disk_file_content_for_input(
> ```

override terminology was pretty good one and i think it still describes what we want to achieve here
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 550 at r2](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OqATYGj7IcO2_NUr0b6:-OqATYGj7IcO2_NUr0b7:b-u4b6px) ([raw file](https://github.com/starkware-libs/cairo/blob/c93308053a7fd8c6c3ea9b4a703a4857553c02b7/crates/cairo-lang-filesystem/src/db.rs#L550)):*
> ```XML
> ) {
>     let handle = ensure_file_contents_handle_for_input(db, file_input);
>     handle.set_generated_content(db).with_durability(Durability::HIGH).to(content.map(ArcStr::new));
> ```

Why is durability value different here?

piotmag769 (changes requested):
> @piotmag769 made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/9830)* status: all files reviewed, 12 unresolved discussions (waiting on Arcticae and integraledelebesgue).

___
*[`crates/cairo-lang-filesystem/src/db.rs` line 268 at r2](https://reviewable.io/reviews/starkware-libs/cairo/9830#-Oq5gPlm0r_T690FUodu:-OqFbEXB5JcbXGoeTfjk:b-udumzg) ([raw file](https://github.com/starkware-libs/cairo/blob/c93308053a7fd8c6c3ea9b4a703a4857553c02b7/crates/cairo-lang-filesystem/src/db.rs#L268)):*
<details><summary><i>Previously, Arcticae (Tomasz Rejowski) wrote…</i></summary><blockquote>

Also `Default` should be derivable here
</blockquote></details>

Wdym? The default is there since `FileContentStorage` is a only type alias
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 274 at r2](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OqBO5PAApXpCDcHtS04:-OqFbD8aBoyUVvgjb6cU:b-5ute97) ([raw file](https://github.com/starkware-libs/cairo/blob/c93308053a7fd8c6c3ea9b4a703a4857553c02b7/crates/cairo-lang-filesystem/src/db.rs#L274)):*
<details><summary><i>Previously, Arcticae (Tomasz Rejowski) wrote…</i></summary><blockquote>

Describe why a view is used
</blockquote></details>

I am not sure wdym by that - it is the same as other traits that have `Database` as supertrait. There is nothing salsa view specific in this trait
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 550 at r2](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OqATYGj7IcO2_NUr0b6:-OqFb2zn0GN549-ZU6Bf:b-8a75uq) ([raw file](https://github.com/starkware-libs/cairo/blob/c93308053a7fd8c6c3ea9b4a703a4857553c02b7/crates/cairo-lang-filesystem/src/db.rs#L550)):*
<details><summary><i>Previously, Arcticae (Tomasz Rejowski) wrote…</i></summary><blockquote>

Why is durability value different here?
</blockquote></details>

Generated content is set rarely - e.g. we do that when there are integration test in Scarb package without `tests/lib.cairo` file - in that case we set generated content for `tests/lib.cairo`. I still believe there shouldn't se a separate field for that in `FileContents`, but a helper with different durability is definitely useful

wawel37 (commented):
> @wawel37 made 10 comments.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/9830)* status: 3 of 13 files reviewed, 12 unresolved discussions (waiting on Arcticae, integraledelebesgue, and piotmag769).

___
*[`crates/cairo-lang-filesystem/src/db.rs` line 237 at r2](https://reviewable.io/reviews/starkware-libs/cairo/9830#-Oq5jton8HzHu-AnpBPE:-OqKM3Mv8XzP_zZ3oM1b:bb6un1j) ([raw file](https://github.com/starkware-libs/cairo/blob/66e3d47796da5d2dbb6982bc0be3d281bbe10d01/crates/cairo-lang-filesystem/src/db.rs#L237)):*
<details><summary><i>Previously, piotmag769 (Piotr Magiera) wrote…</i></summary><blockquote>

Doesn't explain why we need this
</blockquote></details>

Done
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 259 at r2](https://reviewable.io/reviews/starkware-libs/cairo/9830#-Oq5gY-h6Oigvs2zUuZx:-OqKM2hjEw6Zc5X2Gj9f:bb6un1j) ([raw file](https://github.com/starkware-libs/cairo/blob/66e3d47796da5d2dbb6982bc0be3d281bbe10d01/crates/cairo-lang-filesystem/src/db.rs#L259)):*
<details><summary><i>Previously, piotmag769 (Piotr Magiera) wrote…</i></summary><blockquote>

Do we need two fields? Only one is set at a time anyways, the other has to be `None`
</blockquote></details>

Done
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 268 at r2](https://reviewable.io/reviews/starkware-libs/cairo/9830#-Oq5gPlm0r_T690FUodu:-OqKM1oFBd7kKJ4k7F3g:bb6un1j) ([raw file](https://github.com/starkware-libs/cairo/blob/66e3d47796da5d2dbb6982bc0be3d281bbe10d01/crates/cairo-lang-filesystem/src/db.rs#L268)):*
<details><summary><i>Previously, piotmag769 (Piotr Magiera) wrote…</i></summary><blockquote>

Wdym? The default is there since `FileContentStorage` is a only type alias
</blockquote></details>

Done
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 300 at r2](https://reviewable.io/reviews/starkware-libs/cairo/9830#-Oq5h7cgAK2SeIZFZmVU:-OqKM-E55ETmgzbPbtDI:bb6un1j) ([raw file](https://github.com/starkware-libs/cairo/blob/66e3d47796da5d2dbb6982bc0be3d281bbe10d01/crates/cairo-lang-filesystem/src/db.rs#L300)):*
<details><summary><i>Previously, piotmag769 (Piotr Magiera) wrote…</i></summary><blockquote>

Put this next to `register_files_group_view` or preferably declare inside it
</blockquote></details>

Done
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 499 at r2](https://reviewable.io/reviews/starkware-libs/cairo/9830#-Oq5iTEo2ddHgHc3WKv-:-OqKLzMjDkaYsMzY82JF:bb6un1j) ([raw file](https://github.com/starkware-libs/cairo/blob/66e3d47796da5d2dbb6982bc0be3d281bbe10d01/crates/cairo-lang-filesystem/src/db.rs#L499)):*
<details><summary><i>Previously, piotmag769 (Piotr Magiera) wrote…</i></summary><blockquote>

Inline
</blockquote></details>

Done
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 517 at r2](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OqBNkLp3WM7aXSWp9eo:-OqKLyI69Ll3LrqreWSn:bb6un1j) ([raw file](https://github.com/starkware-libs/cairo/blob/66e3d47796da5d2dbb6982bc0be3d281bbe10d01/crates/cairo-lang-filesystem/src/db.rs#L517)):*
<details><summary><i>Previously, Arcticae (Tomasz Rejowski) wrote…</i></summary><blockquote>

This one is unused i think
</blockquote></details>

Done
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 526 at r2](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OqBNySr7KuXctm6mqSs:-OqKLxle841CzG1T7h1D:bb6un1j) ([raw file](https://github.com/starkware-libs/cairo/blob/66e3d47796da5d2dbb6982bc0be3d281bbe10d01/crates/cairo-lang-filesystem/src/db.rs#L526)):*
<details><summary><i>Previously, Arcticae (Tomasz Rejowski) wrote…</i></summary><blockquote>

override terminology was pretty good one and i think it still describes what we want to achieve here
</blockquote></details>

Done
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 550 at r2](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OqATYGj7IcO2_NUr0b6:-OqKKDlC7DnCya3mWTUy:bbu482y) ([raw file](https://github.com/starkware-libs/cairo/blob/66e3d47796da5d2dbb6982bc0be3d281bbe10d01/crates/cairo-lang-filesystem/src/db.rs#L550)):*
<details><summary><i>Previously, piotmag769 (Piotr Magiera) wrote…</i></summary><blockquote>

Generated content is set rarely - e.g. we do that when there are integration test in Scarb package without `tests/lib.cairo` file - in that case we set generated content for `tests/lib.cairo`. I still believe there shouldn't se a separate field for that in `FileContents`, but a helper with different durability is definitely useful
</blockquote></details>

Added proper comment for that
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 555 at r2](https://reviewable.io/reviews/starkware-libs/cairo/9830#-Oq5ji6DCjFj3WYYo578:-OqK24o16ViJg-XMRY6L:b3xxr3k) ([raw file](https://github.com/starkware-libs/cairo/blob/66e3d47796da5d2dbb6982bc0be3d281bbe10d01/crates/cairo-lang-filesystem/src/db.rs#L555)):*
<details><summary><i>Previously, piotmag769 (Piotr Magiera) wrote…</i></summary><blockquote>

Where is it needed? Also doc
</blockquote></details>

It's needed for linter fixer
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 657 at r2](https://reviewable.io/reviews/starkware-libs/cairo/9830#-Oq5iYHgEWyJzSrKXGG7:-OqKLwUh9PdsjFFP2Jt3:bb6un1j) ([raw file](https://github.com/starkware-libs/cairo/blob/66e3d47796da5d2dbb6982bc0be3d281bbe10d01/crates/cairo-lang-filesystem/src/db.rs#L657)):*
<details><summary><i>Previously, Arcticae (Tomasz Rejowski) wrote…</i></summary><blockquote>

Mechanism of invalidation should be described here (on fist file edit vs later edits)
</blockquote></details>

Done

piotmag769 (changes requested):
> @piotmag769 reviewed 10 files and all commit messages, made 7 comments, and resolved 6 discussions.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/9830)* status: all files reviewed, 12 unresolved discussions (waiting on Arcticae and integraledelebesgue).

___
*[`crates/cairo-lang-filesystem/src/db.rs` line 550 at r2](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OqATYGj7IcO2_NUr0b6:-OqL-w-n0eIQavF_QKOg:b-kgn8u7) ([raw file](https://github.com/starkware-libs/cairo/blob/66e3d47796da5d2dbb6982bc0be3d281bbe10d01/crates/cairo-lang-filesystem/src/db.rs#L550)):*
<details><summary><i>Previously, wawel37 (Mateusz Kowalski) wrote…</i></summary><blockquote>

Added proper comment for that
</blockquote></details>

The comment does not reflect what I wrote ;p Your current comment is wrong, this is not used for plugin generated content. Please change
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 240 at r4](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OqKw0A01Cvmse15nmrG:-OqKw0A14Hmpizloac1f:b-3y9vsw) ([raw file](https://github.com/starkware-libs/cairo/blob/31c2d753cd378123b1a4c810a2db0709bf26ffe1/crates/cairo-lang-filesystem/src/db.rs#L240)):*
> ```XML
>     /// file. The [`file_content`] tracked function reads this field so that it re-executes when a
>     /// previously-unregistered file gets a handle — at which point it records the per-file
>     /// dependency for future fine-grained incremental updates.
> ```

Imprecise.Dependency on particular `FileContents` is registered by `file_content` when the tracked function reads the field of this `FileContents`



_Suggestion:_
```XML
    /// previously-unregistered file gets a handle.
```
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 260 at r4](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OqKx1Gu3I7eFcY8aU6_:-OqKx1Gu3I7eFcY8aU6a:b-7mvi4o) ([raw file](https://github.com/starkware-libs/cairo/blob/31c2d753cd378123b1a4c810a2db0709bf26ffe1/crates/cairo-lang-filesystem/src/db.rs#L260)):*
> ```XML
> /// Per-file content slot used by long-running tools (like CairoLS) that need granular
> /// invalidation. A file is either an on-disk override or a generated file — never both — so a
> /// single field is sufficient.
> ```

Sth like that is more precise



_Suggestion:_
```XML
/// Input used for overriding a content of a single file.
/// Useful for tools like CairoLS to manage contents of files that are opened in the editor.
///
/// Notice that this structure holds content override for a single file only - not the whole hashmap of overrides for all overriden files.
/// This is done to ensure granular invalidation of results of `file_content` tracked function - so that we don't invalidate everything when a single override changes.
```
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 262 at r4](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OqKxaVa8DmzQKW0pLjD:-OqKxaVa8DmzQKW0pLjE:b-9ml3ob) ([raw file](https://github.com/starkware-libs/cairo/blob/31c2d753cd378123b1a4c810a2db0709bf26ffe1/crates/cairo-lang-filesystem/src/db.rs#L262)):*
> ```XML
> /// single field is sufficient.
> #[salsa::input]
> pub struct FileContents {
> ```

More precise name imo - I think all the names should be changes accordingly



_Suggestion:_
```XML
FileContentOverride
```
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 500 at r4](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OqL0NzUE1WaHaJh3PDC:-OqL0NzUE1WaHaJh3PDD:b45h) ([raw file](https://github.com/starkware-libs/cairo/blob/31c2d753cd378123b1a4c810a2db0709bf26ffe1/crates/cairo-lang-filesystem/src/db.rs#L500)):*
> ```XML
> }
> 
> /// Overrides the on-disk content of a file as seen by the compiler. Used by editors to show
> ```





_Suggestion:_
```XML
Used by CairoLS to make compiler aware of
```
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 508 at r4](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OqL-ONdCTlkuXqeIWT2:-OqL-ONdCTlkuXqeIWT3:b45h) ([raw file](https://github.com/starkware-libs/cairo/blob/31c2d753cd378123b1a4c810a2db0709bf26ffe1/crates/cairo-lang-filesystem/src/db.rs#L508)):*
> ```XML
> ) {
>     let handle = ensure_file_contents_handle_for_input(db, file_input);
>     // LOW durability: on-disk overrides change on any even minor file changes.
> ```





_Suggestion:_
```XML
// LOW durability: on-disk overrides change frequently for files opened in editor in CairoLS.
// For regular compiling use cases the durability does not matter, as the inputs do not change during compilation.
```
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 631 at r4](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OqL22Ye0maADUql8jw1:-OqL22Ye0maADUql8jw2:bnr4qrg) ([raw file](https://github.com/starkware-libs/cairo/blob/31c2d753cd378123b1a4c810a2db0709bf26ffe1/crates/cairo-lang-filesystem/src/db.rs#L631)):*
> ```XML
>     let _ = files_group_input(db).file_contents_revision(db);
>     maybe_file_content_view(db)
>         .and_then(|view| view.file_content_override(file_id))
> ```

WDYT? This saves invalidation of this tracked function for `FileId` for which there already exists `FileContents` in the `FileContentStorage`. 

Explanation: we need the dependency on `file_contents_revision` only for `FileId` which is **not** in the `FileContentStorage` map. Such a file has to depend on `file_content_revision` that we bump on each new `FileContents` creation - the newly created input (content override) may have been for the aforementioned file. For files which already have their `FileContents` - we do not care about creation of new `FileContents` - they already depend on their own `FileContents`.



_Suggestion:_
```XML
    maybe_file_content_view(db)
        .and_then(|db| {
            let file_input = db.file_input(file_id);
            if let Some(file_contents) = db.file_contents_for_input(file_input) else {
                file_contents.content(db)
            } else {
                let _ = files_group_input(db).file_contents_revision(db);
                None
            }
        })
```

wawel37 (commented):
> @wawel37 made 7 comments.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/9830)* status: 12 of 13 files reviewed, 12 unresolved discussions (waiting on Arcticae, integraledelebesgue, orizi, and piotmag769).

___
*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/9830#-Oq9lULx0io4g8Wzxr9e:-OqLPkzr4Yr4jS2BsPeD:b-vuu1r6) (no related file):*

without a full documentation of what you are trying to do it and how and why it helps - would definitely not approve the PR.
</blockquote></details>

Added extended docs to the Code. Lmk if I should extend the docs more
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 550 at r2](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OqATYGj7IcO2_NUr0b6:-OqLbWCp8SGit6PLyRqZ:b-txeouz) ([raw file](https://github.com/starkware-libs/cairo/blob/66e3d47796da5d2dbb6982bc0be3d281bbe10d01/crates/cairo-lang-filesystem/src/db.rs#L550)):*
<details><summary><i>Previously, piotmag769 (Piotr Magiera) wrote…</i></summary><blockquote>

The comment does not reflect what I wrote ;p Your current comment is wrong, this is not used for plugin generated content. Please change
</blockquote></details>

Done, sorry, replied to wrong comment. Anyway, changed the comments here also. 
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 240 at r4](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OqKw0A01Cvmse15nmrG:-OqL_jAE7xl7rmzS_N2q:bb6un1j) ([raw file](https://github.com/starkware-libs/cairo/blob/31c2d753cd378123b1a4c810a2db0709bf26ffe1/crates/cairo-lang-filesystem/src/db.rs#L240)):*
<details><summary><i>Previously, piotmag769 (Piotr Magiera) wrote…</i></summary><blockquote>

Imprecise.Dependency on particular `FileContents` is registered by `file_content` when the tracked function reads the field of this `FileContents`
</blockquote></details>

Done
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 260 at r4](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OqKx1Gu3I7eFcY8aU6_:-OqL_wDHAsnovvV4gwEF:bb6un1j) ([raw file](https://github.com/starkware-libs/cairo/blob/31c2d753cd378123b1a4c810a2db0709bf26ffe1/crates/cairo-lang-filesystem/src/db.rs#L260)):*
<details><summary><i>Previously, piotmag769 (Piotr Magiera) wrote…</i></summary><blockquote>

Sth like that is more precise
</blockquote></details>

Done
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 262 at r4](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OqKxaVa8DmzQKW0pLjD:-OqL_vb8AZYDMxVKQv-V:bb6un1j) ([raw file](https://github.com/starkware-libs/cairo/blob/31c2d753cd378123b1a4c810a2db0709bf26ffe1/crates/cairo-lang-filesystem/src/db.rs#L262)):*
<details><summary><i>Previously, piotmag769 (Piotr Magiera) wrote…</i></summary><blockquote>

More precise name imo - I think all the names should be changes accordingly
</blockquote></details>

Done
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 500 at r4](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OqL0NzUE1WaHaJh3PDC:-OqL_u592MKego-mgbZL:bb6un1j) ([raw file](https://github.com/starkware-libs/cairo/blob/31c2d753cd378123b1a4c810a2db0709bf26ffe1/crates/cairo-lang-filesystem/src/db.rs#L500)):*
> ```XML
> }
> 
> /// Overrides the on-disk content of a file as seen by the compiler. Used by editors to show
> ```

Done
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 508 at r4](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OqL-ONdCTlkuXqeIWT2:-OqLa3O83mjYOui3n_ya:bb6un1j) ([raw file](https://github.com/starkware-libs/cairo/blob/31c2d753cd378123b1a4c810a2db0709bf26ffe1/crates/cairo-lang-filesystem/src/db.rs#L508)):*
> ```XML
> ) {
>     let handle = ensure_file_contents_handle_for_input(db, file_input);
>     // LOW durability: on-disk overrides change on any even minor file changes.
> ```

Done

piotmag769 (changes requested):
> @piotmag769 reviewed 1 file and all commit messages, made 1 comment, and resolved 6 discussions.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/9830)* status: all files reviewed, 6 unresolved discussions (waiting on Arcticae, integraledelebesgue, and orizi).

___
*[`crates/cairo-lang-filesystem/src/db.rs` line 550 at r2](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OqATYGj7IcO2_NUr0b6:-OqLmYqG4slG_zO_PVKr:b78i40q) ([raw file](https://github.com/starkware-libs/cairo/blob/66e3d47796da5d2dbb6982bc0be3d281bbe10d01/crates/cairo-lang-filesystem/src/db.rs#L550)):*
<details><summary><i>Previously, wawel37 (Mateusz Kowalski) wrote…</i></summary><blockquote>

Done, sorry, replied to wrong comment. Anyway, changed the comments here also.
</blockquote></details>

This is still wrong - in LS it can change more frequently than this. I would just say that it is set rarely - to make compiler act like there exists a cairo file (usually `lib.cairo`) while this file does not actually exists on the disk.

integraledelebesgue (commented):
> @integraledelebesgue made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/9830)* status: all files reviewed, 6 unresolved discussions (waiting on Arcticae, orizi, and piotmag769).

___
*[`crates/cairo-lang-filesystem/src/db.rs` line 274 at r2](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OqBO5PAApXpCDcHtS04:-OqjP4cM-B76iMfakulo:b-id54ur) ([raw file](https://github.com/starkware-libs/cairo/blob/66e3d47796da5d2dbb6982bc0be3d281bbe10d01/crates/cairo-lang-filesystem/src/db.rs#L274)):*
<details><summary><i>Previously, piotmag769 (Piotr Magiera) wrote…</i></summary><blockquote>

I am not sure wdym by that - it is the same as other traits that have `Database` as supertrait. There is nothing salsa view specific in this trait
</blockquote></details>

Right, nothing related to Salsa view. This trait extracts the common logic of access to the *storage*, which is db-specific (each db has its own field). We simply have to implement the getter and get the rest of the logic implemented
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 631 at r4](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OqL22Ye0maADUql8jw1:-OqjQKJz2BnjjDok15MB:bdokuzt) ([raw file](https://github.com/starkware-libs/cairo/blob/31c2d753cd378123b1a4c810a2db0709bf26ffe1/crates/cairo-lang-filesystem/src/db.rs#L631)):*
<details><summary><i>Previously, piotmag769 (Piotr Magiera) wrote…</i></summary><blockquote>

WDYT? This saves invalidation of this tracked function for `FileId` for which there already exists `FileContents` in the `FileContentStorage`. 

Explanation: we need the dependency on `file_contents_revision` only for `FileId` which is **not** in the `FileContentStorage` map. Such a file has to depend on `file_content_revision` that we bump on each new `FileContents` creation - the newly created input (content override) may have been for the aforementioned file. For files which already have their `FileContents` - we do not care about creation of new `FileContents` - they already depend on their own `FileContents`.
</blockquote></details>

I'm not sure if Salsa handles the dependency on `file_contents_revision` per-entry or per-query. The conditional dependency only makes sense if the former is true

piotmag769 (commented):
> @piotmag769 made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/9830)* status: all files reviewed, 6 unresolved discussions (waiting on Arcticae, integraledelebesgue, and orizi).

___
*[`crates/cairo-lang-filesystem/src/db.rs` line 631 at r4](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OqL22Ye0maADUql8jw1:-OqjVBWe82bSQkTy6hPq:bwhvky0) ([raw file](https://github.com/starkware-libs/cairo/blob/31c2d753cd378123b1a4c810a2db0709bf26ffe1/crates/cairo-lang-filesystem/src/db.rs#L631)):*
It is per entry - it would be weird for it to be otherwise

https://salsa-rs.github.io/salsa/reference/algorithm.html#basic-rule-when-inputs-change-re-execute

>When you invoke a tracked function, in addition to storing the value that was returned, we also track what _other_ tracked functions it depends on, and the revisions when their value last changed. When you invoke the function again, if the database is in a new revision, then we check whether any of the inputs to this function have changed in that new revision. If not, we can just return our cached value.

integraledelebesgue (commented):
> @integraledelebesgue made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/9830)* status: all files reviewed, 6 unresolved discussions (waiting on Arcticae, orizi, and piotmag769).

___
*[`crates/cairo-lang-filesystem/src/db.rs` line 631 at r4](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OqL22Ye0maADUql8jw1:-OqjbXXO34L9ck45dDfy:blth693) ([raw file](https://github.com/starkware-libs/cairo/blob/31c2d753cd378123b1a4c810a2db0709bf26ffe1/crates/cairo-lang-filesystem/src/db.rs#L631)):*
<details><summary><i>Previously, piotmag769 (Piotr Magiera) wrote…</i></summary><blockquote>

It is per entry - it would be weird for it to be otherwise

https://salsa-rs.github.io/salsa/reference/algorithm.html#basic-rule-when-inputs-change-re-execute

>When you invoke a tracked function, in addition to storing the value that was returned, we also track what _other_ tracked functions it depends on, and the revisions when their value last changed. When you invoke the function again, if the database is in a new revision, then we check whether any of the inputs to this function have changed in that new revision. If not, we can just return our cached value.
</blockquote></details>

So I think it's correct to do so

wawel37 (commented):
> @wawel37 made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/9830)* status: 12 of 13 files reviewed, 6 unresolved discussions (waiting on Arcticae, orizi, and piotmag769).

___
*[`crates/cairo-lang-filesystem/src/db.rs` line 550 at r2](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OqATYGj7IcO2_NUr0b6:-Oqk_hnyCmIFdfN7CFMT:bb6un1j) ([raw file](https://github.com/starkware-libs/cairo/blob/66e3d47796da5d2dbb6982bc0be3d281bbe10d01/crates/cairo-lang-filesystem/src/db.rs#L550)):*
<details><summary><i>Previously, piotmag769 (Piotr Magiera) wrote…</i></summary><blockquote>

This is still wrong - in LS it can change more frequently than this. I would just say that it is set rarely - to make compiler act like there exists a cairo file (usually `lib.cairo`) while this file does not actually exists on the disk.
</blockquote></details>

Done
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 631 at r4](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OqL22Ye0maADUql8jw1:-Oqk_jCm6VOUIqBPaU7s:bb6un1j) ([raw file](https://github.com/starkware-libs/cairo/blob/31c2d753cd378123b1a4c810a2db0709bf26ffe1/crates/cairo-lang-filesystem/src/db.rs#L631)):*
<details><summary><i>Previously, integraledelebesgue (Jan Smółka) wrote…</i></summary><blockquote>

So I think it's correct to do so
</blockquote></details>

Done

piotmag769 (approved):
:lgtm:

> @piotmag769 reviewed 1 file and all commit messages, made 1 comment, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/9830#-:-OqkbpBfDm5g7Yb0QvOW:bnfp4nl)* status: all files reviewed, 5 unresolved discussions (waiting on Arcticae and orizi).

orizi (changes requested):
> @orizi reviewed all commit messages, made 7 comments, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/9830)* status: all files reviewed, 10 unresolved discussions (waiting on Arcticae and integraledelebesgue).

___
*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OqtUDswDJXVeEopVpgo:-OqtUDswDJXVeEopVpgp:bf901qd) (no related file):*
@eytan-starkware 
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 239 at r6](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OqtToqK1FygOHJBq4gw:-OqtToqK1FygOHJBq4gx:b-xataih) ([raw file](https://github.com/starkware-libs/cairo/blob/9048a53e55e84e2b0d345ce84e2190bf2c673fde/crates/cairo-lang-filesystem/src/db.rs#L239)):*
> ```XML
>     /// Structural revision bumped each time a new [`FileContentOverride`] handle is first created
>     /// for a previously-unregistered file.
>     pub file_contents_revision: u64,
> ```

why isn't salsa any internal revision any better?

add this somewhere at the doc.
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 279 at r6](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OqtTUXl2aFs47LZkUfC:-OqtTUXl2aFs47LZkUfD:b-rnzpcb) ([raw file](https://github.com/starkware-libs/cairo/blob/9048a53e55e84e2b0d345ce84e2190bf2c673fde/crates/cairo-lang-filesystem/src/db.rs#L279)):*
> ```XML
> /// Implement this on the concrete DB struct and register it with [`register_files_group_view`].
> pub trait FileContentView: Database {
>     fn file_content_storage(&self) -> Option<&FileContentStorage> {
> ```

doc fn
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 283 at r6](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OqtTVTG6X2USLcJzlIM:-OqtTVTG6X2USLcJzlIN:b-rnzpcb) ([raw file](https://github.com/starkware-libs/cairo/blob/9048a53e55e84e2b0d345ce84e2190bf2c673fde/crates/cairo-lang-filesystem/src/db.rs#L283)):*
> ```XML
>     }
> 
>     fn file_contents_for_input(&self, file_input: &FileInput) -> Option<FileContentOverride> {
> ```

doc fn
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 294 at r6](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OqtTW6H5ip73_trAXSj:-OqtTW6H5ip73_trAXSk:b-rnzpcb) ([raw file](https://github.com/starkware-libs/cairo/blob/9048a53e55e84e2b0d345ce84e2190bf2c673fde/crates/cairo-lang-filesystem/src/db.rs#L294)):*
> ```XML
> }
> 
> fn file_content_view(db: &dyn Database) -> &dyn FileContentView {
> ```

doc fn
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 303 at r6](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OqtTXBY0uGuEn5yQDrf:-OqtTXBY0uGuEn5yQDrg:b-rnzpcb) ([raw file](https://github.com/starkware-libs/cairo/blob/9048a53e55e84e2b0d345ce84e2190bf2c673fde/crates/cairo-lang-filesystem/src/db.rs#L303)):*
> ```XML
> }
> 
> fn maybe_file_content_view(db: &dyn Database) -> Option<&dyn FileContentView> {
> ```

doc fn
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 391 at r6](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OqtT_4j1X17bymWlB1U:-OqtT_4j1X17bymWlB1V:b-rnzpcb) ([raw file](https://github.com/starkware-libs/cairo/blob/9048a53e55e84e2b0d345ce84e2190bf2c673fde/crates/cairo-lang-filesystem/src/db.rs#L391)):*
> ```XML
> }
> 
> fn bump_file_contents_revision(db: &mut dyn Database) {
> ```

doc fn

wawel37 (commented):
> @wawel37 made 6 comments.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/9830)* status: 12 of 13 files reviewed, 10 unresolved discussions (waiting on Arcticae, orizi, and piotmag769).

___
*[`crates/cairo-lang-filesystem/src/db.rs` line 239 at r6](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OqtToqK1FygOHJBq4gw:-Oqu9hpm0MSSi-SjH9Sl:b-uixapw) ([raw file](https://github.com/starkware-libs/cairo/blob/9048a53e55e84e2b0d345ce84e2190bf2c673fde/crates/cairo-lang-filesystem/src/db.rs#L239)):*

why isn't salsa any internal revision any better?

add this somewhere at the doc.
</blockquote></details>

Done, added docs explaining why we do it this way
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 279 at r6](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OqtTUXl2aFs47LZkUfC:-Oqu9liy2juHPlCH4UWI:bb6un1j) ([raw file](https://github.com/starkware-libs/cairo/blob/9048a53e55e84e2b0d345ce84e2190bf2c673fde/crates/cairo-lang-filesystem/src/db.rs#L279)):*

doc fn
</blockquote></details>

Done
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 283 at r6](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OqtTVTG6X2USLcJzlIM:-Oqu9mKk-ptshfz2w1b7:bb6un1j) ([raw file](https://github.com/starkware-libs/cairo/blob/9048a53e55e84e2b0d345ce84e2190bf2c673fde/crates/cairo-lang-filesystem/src/db.rs#L283)):*

doc fn
</blockquote></details>

Done
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 294 at r6](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OqtTW6H5ip73_trAXSj:-Oqu9nO04dxNI_DT90PB:bb6un1j) ([raw file](https://github.com/starkware-libs/cairo/blob/9048a53e55e84e2b0d345ce84e2190bf2c673fde/crates/cairo-lang-filesystem/src/db.rs#L294)):*

doc fn
</blockquote></details>

Done
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 303 at r6](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OqtTXBY0uGuEn5yQDrf:-Oqu9nyd4dJoGld9j8Zz:bb6un1j) ([raw file](https://github.com/starkware-libs/cairo/blob/9048a53e55e84e2b0d345ce84e2190bf2c673fde/crates/cairo-lang-filesystem/src/db.rs#L303)):*

doc fn
</blockquote></details>

Done
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 391 at r6](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OqtT_4j1X17bymWlB1U:-Oqu9oxa8cVVwlCK3P5M:bb6un1j) ([raw file](https://github.com/starkware-libs/cairo/blob/9048a53e55e84e2b0d345ce84e2190bf2c673fde/crates/cairo-lang-filesystem/src/db.rs#L391)):*

doc fn
</blockquote></details>

Done

piotmag769 (approved):
> @piotmag769 reviewed 1 file and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/9830)* status: all files reviewed, 10 unresolved discussions (waiting on Arcticae and orizi).

eytan-starkware (commented):
Have you considered moving all file contents out of the DB, and having the DB "start" from the AST?
I am essentially suggesting "handle all parsing before the DB, and have the DB only track AST nodes" which I suppose might improve stability, but will mostly improve code.

> @eytan-starkware made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/9830#-:-Or8hQzR7OGfXVHxyQEV:b641pqo)* status: all files reviewed, 10 unresolved discussions (waiting on Arcticae and orizi).

wawel37 (commented):
@eytan-starkware 

1. Storing the file content inside the DB is very convenient for cairols, as we change the file content really frequently. Having the parsing mechanism outside of the DB would make the API of the outside parser much more complicated and harder to maintain. So the code would not be cleaner this way.
2. The parsing of virtual files would be much more complicated, as we need the built macro code from virtual files to create a correct AST. So with parsing outside the DB, we would have to first parse the code, pass the AST to DB and later let it expand the macros. Then again we would go back to parsing again, as the code changed because of the virtual files. I got a feeling that might not be a good solution.
3. Also we would have to probably move some other things that the parsing depends on out of the DB, like: crate's edition and feature flags (as far as i know, they are stored inside the db).

> @wawel37 made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/9830#-:-OrDn1URBRKf6WxE4BGl:b-xc7dpw)* status: all files reviewed, 10 unresolved discussions (waiting on Arcticae and orizi).

eytan-starkware (changes requested):
@wawel37 
3. Details are wrong, but I will accept that it is out of scope at the moment. It is a much cleaner approach to pull parsing out of DB IMO and you should consider it. Changing a file without changing the ast form (comments, renames, and such) should be easier to deal with this way.

I am really not convinced about 1. You would always parse again, I see no real gain in having it inside the DB (as it is fast, and happens often).
2. Parsing virtual files would just mean that processing a macro as a query would include parsing. I am not seeing a problem with that.

> @eytan-starkware reviewed 3 files and all commit messages, and made 4 comments.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/9830#-:-OrDpWST2BmBe3VbfVZw:bruk23w)* status: all files reviewed, 13 unresolved discussions (waiting on Arcticae, integraledelebesgue, and orizi).

___
*[`crates/cairo-lang-compiler/src/db.rs` line 116 at r7](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OrET6GYBE5yA9YmAunz:-OrET6GYBE5yA9YmAuo-:b-t6wgrs) ([raw file](https://github.com/starkware-libs/cairo/blob/ebfe90af179c6853ddb60778fb9113950b6eb9d7/crates/cairo-lang-compiler/src/db.rs#L116)):*
> ```XML
>     /// Snapshots the db for read only.
>     pub fn snapshot(&self) -> RootDatabase {
>         RootDatabase { storage: self.storage.clone(), file_contents: self.file_contents.clone() }
> ```

Snapshot should be readonly, but now we are exposing a writable file_contents. What will happen if you mutate inputs on the snapshot/origianl while processing the other?
___
*[`crates/cairo-lang-compiler/src/project.rs` line 68 at r7](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OrEWl1N0ISDSOKFRvBa:-OrEWl1N0ISDSOKFRvBb:bef4ga3) ([raw file](https://github.com/starkware-libs/cairo/blob/ebfe90af179c6853ddb60778fb9113950b6eb9d7/crates/cairo-lang-compiler/src/project.rs#L68)):*
> ```XML
>         }
>         let file_input = {
>             let crate_id = CrateId::plain(db, SmolStrId::from(db, file_stem));
> ```

Create crate_id once
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 288 at r7](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OrESMu18ZsL9CYXu6oY:-OrESMu18ZsL9CYXu6oZ:b-o730kk) ([raw file](https://github.com/starkware-libs/cairo/blob/ebfe90af179c6853ddb60778fb9113950b6eb9d7/crates/cairo-lang-filesystem/src/db.rs#L288)):*
> ```XML
>     /// Returns the [`FileContentStorage`] side-table, or `None` if this database does not hold
>     /// file content overrides.
>     fn file_content_storage(&self) -> Option<&FileContentStorage> {
> ```

Why is this function returning an Option? All impls return Some

wawel37 (commented):
@eytan-starkware You might be right all along. I'm just curious if we could accept those changes right now as they are, and we will implement yours approach and tests all the performances later on, as we've already measured the gains from this changes and they are really beneficial for us. What do you think?

> @wawel37 made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/9830#-:-OrIXx0pBVstTCcU8hAB:b-joxlo9)* status: all files reviewed, 13 unresolved discussions (waiting on Arcticae, integraledelebesgue, and orizi).

eytan-starkware (commented):
That is essentially what I said, so yes.

> @eytan-starkware made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/9830#-:-OrIb4fa9hMuDStM7kwh:bvrb6wy)* status: all files reviewed, 13 unresolved discussions (waiting on Arcticae, integraledelebesgue, and orizi).

wawel37 (commented):
> @wawel37 made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/9830)* status: 2 of 13 files reviewed, 13 unresolved discussions (waiting on Arcticae, eytan-starkware, orizi, and piotmag769).

___
*[`crates/cairo-lang-compiler/src/db.rs` line 116 at r7](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OrET6GYBE5yA9YmAunz:-OrIyMm5C81LkYE7tRVw:bb6un1j) ([raw file](https://github.com/starkware-libs/cairo/blob/ebfe90af179c6853ddb60778fb9113950b6eb9d7/crates/cairo-lang-compiler/src/db.rs#L116)):*

Snapshot should be readonly, but now we are exposing a writable file_contents. What will happen if you mutate inputs on the snapshot/origianl while processing the other?
</blockquote></details>

Done
___
*[`crates/cairo-lang-compiler/src/project.rs` line 68 at r7](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OrEWl1N0ISDSOKFRvBa:-OrIyOwY5T3LGi2wNT6n:bb6un1j) ([raw file](https://github.com/starkware-libs/cairo/blob/ebfe90af179c6853ddb60778fb9113950b6eb9d7/crates/cairo-lang-compiler/src/project.rs#L68)):*

Create crate_id once
</blockquote></details>

Done
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 288 at r7](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OrESMu18ZsL9CYXu6oY:-OrIyRS84QjozAVwehcv:bb6un1j) ([raw file](https://github.com/starkware-libs/cairo/blob/ebfe90af179c6853ddb60778fb9113950b6eb9d7/crates/cairo-lang-filesystem/src/db.rs#L288)):*

Why is this function returning an Option? All impls return Some
</blockquote></details>

Done

eytan-starkware (changes requested):
> @eytan-starkware reviewed 3 files, made 4 comments, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/9830)* status: 5 of 13 files reviewed, 15 unresolved discussions (waiting on Arcticae, integraledelebesgue, orizi, and piotmag769).

___
*[`crates/cairo-lang-compiler/src/db.rs` line 116 at r7](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OrET6GYBE5yA9YmAunz:-OrJIsuD-EDpNvXaRo-f:bd2pqj8) ([raw file](https://github.com/starkware-libs/cairo/blob/ebfe90af179c6853ddb60778fb9113950b6eb9d7/crates/cairo-lang-compiler/src/db.rs#L116)):*
<details><summary><i>Previously, wawel37 (Mateusz Kowalski) wrote…</i></summary><blockquote>

Done
</blockquote></details>

Please explain why we need a lock
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 231 at r7](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OrJKh_2Bm70tMP7csGr:-OrJKh_2Bm70tMP7csGs:b-gf3bvc) ([raw file](https://github.com/starkware-libs/cairo/blob/ebfe90af179c6853ddb60778fb9113950b6eb9d7/crates/cairo-lang-filesystem/src/db.rs#L231)):*
> ```XML
> 
> #[salsa::input]
> // TODO(eytan-starkware): Change this mechanism to hold input handles on the db struct outside
> ```

You are essentially doing this todo, but differently then intended, so I want to understand the cause. Why do we need file_contents_revision instead of having the DB struct (e.g. `RootDatabase`) have input handles on it, and use the `set` functions to update the input? As I understand your main concern is a missing file, but that is solvable with things like `db.runtime().report_untracked_read()`? Or even better when you need an input you pass an input handle. 

It seems to me like you are reimplementing parts of salsa tracking, but it is unclear to me why. What is the limitation?
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 273 at r7](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OrJPjcu7qkX8RUB1Z65:-OrJPjcu7qkX8RUB1Z66:bpr3c1h) ([raw file](https://github.com/starkware-libs/cairo/blob/ebfe90af179c6853ddb60778fb9113950b6eb9d7/crates/cairo-lang-filesystem/src/db.rs#L273)):*
> ```XML
> pub struct FileContentOverride {
>     #[returns(ref)]
>     pub content: Option<ArcStr>,
> ```

Do we actually care if it is an override or not? I think this is transparent for the compiler, and a file can jsut be text. If you want option to support missing files that could be fine.
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 281 at r7](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OrJRv90BApyMB_rnfNg:-OrJRv90BApyMB_rnfNh:b-bbfszx) ([raw file](https://github.com/starkware-libs/cairo/blob/ebfe90af179c6853ddb60778fb9113950b6eb9d7/crates/cairo-lang-filesystem/src/db.rs#L281)):*
> ```XML
> /// created with `&mut dyn Database` while existing ones are read from within tracked functions
> /// via [`FileContentView`].
> pub type FileContentStorage = Arc<RwLock<HashMap<FileInput, FileContentOverride>>>;
> ```

Maybe you can remove the lock now that snapshot is deepcopy?

eytan-starkware (changes requested):
> @eytan-starkware made 2 comments and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/9830)* status: 5 of 13 files reviewed, 16 unresolved discussions (waiting on Arcticae, integraledelebesgue, orizi, and piotmag769).

___
*[`crates/cairo-lang-filesystem/src/db.rs` line 296 at r8](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OrJT80UEBuUFFD-bP-E:-OrJT80UEBuUFFD-bP-F:b-pdtrx6) ([raw file](https://github.com/starkware-libs/cairo/blob/0fe47a4389711b039d04e010a990f7d8549e9a1f/crates/cairo-lang-filesystem/src/db.rs#L296)):*
> ```XML
> 
>     /// Returns the overridden content for the given file, if one exists.
>     fn file_content_override<'db>(&'db self, file_id: FileId<'db>) -> Option<&'db ArcStr> {
> ```

Is `file_content_override` ever used?
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 657 at r8](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OrJSxQ37gK8QYVKA9bQ:-OrJSxQ37gK8QYVKA9bR:biqhxcw) ([raw file](https://github.com/starkware-libs/cairo/blob/0fe47a4389711b039d04e010a990f7d8549e9a1f/crates/cairo-lang-filesystem/src/db.rs#L657)):*
> ```XML
>         .and_then(|view| {
>             let file_input = view.file_input(file_id);
>             if let Some(file_contents) = view.file_contents_for_input(file_input) {
> ```

You already have a helper ffunction for reading content from file_input

eytan-starkware (changes requested):
> @eytan-starkware made 4 comments.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/9830)* status: 5 of 13 files reviewed, 20 unresolved discussions (waiting on Arcticae, integraledelebesgue, orizi, and piotmag769).

___
*[`crates/cairo-lang-filesystem/src/db.rs` line 306 at r8](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OrJUtIn3i93XapIzwQn:-OrJUtIn3i93XapIzwQo:b-jeorql) ([raw file](https://github.com/starkware-libs/cairo/blob/0fe47a4389711b039d04e010a990f7d8549e9a1f/crates/cairo-lang-filesystem/src/db.rs#L306)):*
> ```XML
> fn file_content_view(db: &dyn Database) -> &dyn FileContentView {
>     let caster =
>         catch_unwind(AssertUnwindSafe(|| *views(db).downcaster_for::<dyn FileContentView>())).ok();
> ```

You could just have this function return error/option instead of crashing? That way we could return nicer diagnostics (even if it is an internal error)
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 316 at r8](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OrJV3F90p-olKaxpuVX:-OrJV3F90p-olKaxpuVY:b1facv1) ([raw file](https://github.com/starkware-libs/cairo/blob/0fe47a4389711b039d04e010a990f7d8549e9a1f/crates/cairo-lang-filesystem/src/db.rs#L316)):*
> ```XML
> /// [`register_files_group_view`] was not called for this database type.
> fn maybe_file_content_view(db: &dyn Database) -> Option<&dyn FileContentView> {
>     catch_unwind(AssertUnwindSafe(|| file_content_view(db))).ok()
> ```

Wouldnt this be the function returning Option and the previous one using this one?
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 507 at r8](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OrJWIPH1qoe4z4MFQCk:-OrJWIPH1qoe4z4MFQCl:b-2cf3lg) ([raw file](https://github.com/starkware-libs/cairo/blob/0fe47a4389711b039d04e010a990f7d8549e9a1f/crates/cairo-lang-filesystem/src/db.rs#L507)):*
> ```XML
> }
> 
> fn ensure_file_contents_handle_for_input(
> ```

doc and maybe rename
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 544 at r8](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OrJWSq85Y5Sp1rV75-N:-OrJWSq85Y5Sp1rV75-O:b-dq67pc) ([raw file](https://github.com/starkware-libs/cairo/blob/0fe47a4389711b039d04e010a990f7d8549e9a1f/crates/cairo-lang-filesystem/src/db.rs#L544)):*
> ```XML
>     let handle = ensure_file_contents_handle_for_input(db, file_input);
>     // HIGH durability: set rarely.
>     handle.set_content(db).with_durability(Durability::HIGH).to(content.map(ArcStr::new));
> ```

Setting high durability wont miss when generated content changes due to changes in files with low durability? I am pretty sure it will

wawel37 (commented):
> @wawel37 made 10 comments.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/9830)* status: 3 of 13 files reviewed, 20 unresolved discussions (waiting on Arcticae, eytan-starkware, orizi, and piotmag769).

___
*[`crates/cairo-lang-compiler/src/db.rs` line 116 at r7](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OrET6GYBE5yA9YmAunz:-OrOPHLd2ZEcsswhXSmU:b5acplq) ([raw file](https://github.com/starkware-libs/cairo/blob/ebfe90af179c6853ddb60778fb9113950b6eb9d7/crates/cairo-lang-compiler/src/db.rs#L116)):*

Please explain why we need a lock
</blockquote></details>

Done. Added additional comments explaining that.
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 231 at r7](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OrJKh_2Bm70tMP7csGr:-OrOrI2t8XsjreZXIo53:bxt2z76) ([raw file](https://github.com/starkware-libs/cairo/blob/ebfe90af179c6853ddb60778fb9113950b6eb9d7/crates/cairo-lang-filesystem/src/db.rs#L231)):*

You are essentially doing this todo, but differently then intended, so I want to understand the cause. Why do we need file_contents_revision instead of having the DB struct (e.g. `RootDatabase`) have input handles on it, and use the `set` functions to update the input? As I understand your main concern is a missing file, but that is solvable with things like `db.runtime().report_untracked_read()`? Or even better when you need an input you pass an input handle. 

It seems to me like you are reimplementing parts of salsa tracking, but it is unclear to me why. What is the limitation?
</blockquote></details>

Done, used the `report_untracked_read` instead.
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 273 at r7](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OrJPjcu7qkX8RUB1Z65:-OrOrOJT5boyPoBEx5_y:btxksfe) ([raw file](https://github.com/starkware-libs/cairo/blob/ebfe90af179c6853ddb60778fb9113950b6eb9d7/crates/cairo-lang-filesystem/src/db.rs#L273)):*

Do we actually care if it is an override or not? I think this is transparent for the compiler, and a file can jsut be text. If you want option to support missing files that could be fine.
</blockquote></details>

Changed the name but the content stays optional due to some potential cairoLS application for that.
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 281 at r7](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OrJRv90BApyMB_rnfNg:-OrORCrDEo9AAKiFjK0L:b-4m2usx) ([raw file](https://github.com/starkware-libs/cairo/blob/ebfe90af179c6853ddb60778fb9113950b6eb9d7/crates/cairo-lang-filesystem/src/db.rs#L281)):*

Maybe you can remove the lock now that snapshot is deepcopy?
</blockquote></details>

The lock is for clone(), not snapshot() — parallel workers share the same Arc and may read while the main thread writes.
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 296 at r8](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OrJT80UEBuUFFD-bP-E:-OrOTXxcBqdlV5nKaxeA:b-sxka43) ([raw file](https://github.com/starkware-libs/cairo/blob/0fe47a4389711b039d04e010a990f7d8549e9a1f/crates/cairo-lang-filesystem/src/db.rs#L296)):*

Is `file_content_override` ever used?
</blockquote></details>

Renamed to `content_for_input` and it's now used inside `file_content`.
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 306 at r8](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OrJUtIn3i93XapIzwQn:-OrO_pCf9ERxcUUDitYz:b-534b5h) ([raw file](https://github.com/starkware-libs/cairo/blob/0fe47a4389711b039d04e010a990f7d8549e9a1f/crates/cairo-lang-filesystem/src/db.rs#L306)):*

You could just have this function return error/option instead of crashing? That way we could return nicer diagnostics (even if it is an internal error)
</blockquote></details>

I've added some context for the crash. I thought about adding some diagnostics but in my opinion showing it to the user is useless, as if it really crashes, it's 100% dev fault, not the user (forgot to call `register_files_group_view` during db init).
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 316 at r8](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OrJV3F90p-olKaxpuVX:-OrOb7Xg5NYlCdEdVpiE:bvkv70n) ([raw file](https://github.com/starkware-libs/cairo/blob/0fe47a4389711b039d04e010a990f7d8549e9a1f/crates/cairo-lang-filesystem/src/db.rs#L316)):*

Wouldnt this be the function returning Option and the previous one using this one?
</blockquote></details>

Done — `maybe_file_content_view` now does the actual downcast and returns `Option`, `file_content_view` just wraps it with `expect`.
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 507 at r8](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OrJWIPH1qoe4z4MFQCk:-OrOc4Y40y7IVbDxQX6o:bb6un1j) ([raw file](https://github.com/starkware-libs/cairo/blob/0fe47a4389711b039d04e010a990f7d8549e9a1f/crates/cairo-lang-filesystem/src/db.rs#L507)):*

doc and maybe rename
</blockquote></details>

Done
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 544 at r8](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OrJWSq85Y5Sp1rV75-N:-OrOgjEG5tZZku58KMI5:b-p5h43s) ([raw file](https://github.com/starkware-libs/cairo/blob/0fe47a4389711b039d04e010a990f7d8549e9a1f/crates/cairo-lang-filesystem/src/db.rs#L544)):*

Setting high durability wont miss when generated content changes due to changes in files with low durability? I am pretty sure it will
</blockquote></details>

No matter the durability, it's impossible for salsa to miss something due to anything being high or low durability.

From the Salsa source ([src/durability.rs](https://github.com/salsa-rs/salsa/blob/master/src/durability.rs)): the skip only fires when "the only changes were to inputs of low durability" — so when
`set_generated_file_content_for_input` mutates a HIGH durability input, Salsa always re-validates dependents. No correctness issue.
___
*[`crates/cairo-lang-filesystem/src/db.rs` line 657 at r8](https://reviewable.io/reviews/starkware-libs/cairo/9830#-OrJSxQ37gK8QYVKA9bQ:-OrOj8XvC-y41-Q1Piwe:b-896fix) ([raw file](https://github.com/starkware-libs/cairo/blob/0fe47a4389711b039d04e010a990f7d8549e9a1f/crates/cairo-lang-filesystem/src/db.rs#L657)):*

You already have a helper ffunction for reading content from file_input
</blockquote></details>

Done.

piotmag769 (commented):
> @piotmag769 reviewed 11 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/9830)* status: 12 of 13 files reviewed, 20 unresolved discussions (waiting on Arcticae, eytan-starkware, and orizi).

PR #9854, opened 2026-04-17
Filter Mapper by source inference

Keemosty12:
`Mapper` was mapping type/const/impl/negative-impl vars by local id alone, without checking `mapping.source_inference_id`.                          
  That is a correctness issue: vars from a different inference context can be treated as missing mappings (`FreeVariable`) or be mismapped when local ids overlap.

### DIFF
--- a/crates/cairo-lang-semantic/src/expr/inference/canonic.rs
+++ b/crates/cairo-lang-semantic/src/expr/inference/canonic.rs
@@ -639,3 +639,68 @@ impl<'db> SemanticRewriter<NegativeImplLongId<'db>, MapperError> for Mapper<'db,
         Ok(RewriteResult::Modified)
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use cairo_lang_utils::Intern;
+    use salsa::Database;
+
+    use super::{Mapper, VarMapping};
+    use crate::expr::inference::{
+        ConstVar, InferenceId, LocalConstVarId, LocalTypeVarId, TypeVar,
+    };
+    use crate::items::constant::ConstValue;
+    use crate::test_utils::SemanticDatabaseForTesting;
+    use crate::types::TypeLongId;
+
+    #[test]
+    fn mapper_maps_type_var_from_source_inference_context() {
+        let db = SemanticDatabaseForTesting::default();
+        let db: &dyn Database = &db;
+        let mut mapping = VarMapping::new_to_canonic(InferenceId::NoContext);
+        mapping.type_var_mapping.insert(LocalTypeVarId(0), LocalTypeVarId(7));
+
+        let original = TypeLongId::Var(TypeVar {
+            inference_id: InferenceId::NoContext,
+            id: LocalTypeVarId(0),
+        });
+
+        let mapped = Mapper::map(db, original, &mapping).unwrap();
+        assert_eq!(
+            mapped,
+            TypeLongId::Var(TypeVar { inference_id: InferenceId::Canonical, id: LocalTypeVarId(7) })
+        );
+    }
+
+    #[test]
+    fn mapper_ignores_type_var_from_other_inference_context() {
+        let db = SemanticDatabaseForTesting::default();
+        let db: &dyn Database = &db;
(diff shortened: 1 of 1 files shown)


orizi (changes requested):
> @orizi made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/9854)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on Keemosty12).

___
*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/9854#-OqULYyI0fbN0U2SzTFY:-OqULYyJB1hEVciSIA8m:b-i1x76y) (no related file):*
why? add an affected test if this does anything.

Keemosty12:
> why? add an affected test if this does anything.

Sure, I added regression tests for this.                                                                                                                                                                                 They cover exactly this case: Mapper now only rewrites vars from the current source inference context, and leaves vars from other inference contexts unchanged (even if local ids overlap).

orizi (changes requested):
> @orizi made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/9854)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on Keemosty12).

___
*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/9854#-OqULYyI0fbN0U2SzTFY:-Oqf318_2ceJgvADfHBo:b-gr1ujp) (no related file):*

why? add an affected test if this does anything.
</blockquote></details>

If no Cairo code can create this instance, this cannot happen and shouldn't be handled.
Making a case in rust by hand is meaningless.

Keemosty12:
> > @orizi made 1 comment.
> > _[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/9854)_ status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on Keemosty12).
> 
> _[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/9854#-OqULYyI0fbN0U2SzTFY:-Oqf318_2ceJgvADfHBo:b-gr1ujp) (no related file):_
> 
> Previously, orizi wrote…
> If no Cairo code can create this instance, this cannot happen and shouldn't be handled. Making a case in rust by hand is meaningless.

I updated the regression tests to use variables allocated through real inference flows (new_*_var_raw) with overlapping local IDs across different inference_ids, so this now validates the actual runtime path rather than a hand-crafted Rust-only case.

orizi (changes requested):
> @orizi made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/9854)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on Keemosty12).

___
*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/9854#-OqULYyI0fbN0U2SzTFY:-OqfUpVp3vCl6efV_CcN:bicbbny) (no related file):*
<details><summary><i>Previously, Keemosty12 (Keemo Styler) wrote…</i></summary><blockquote>

I updated the regression tests to use variables allocated through real inference flows (new_*_var_raw) with overlapping local IDs across different inference_ids, so this now validates the actual runtime path rather than a hand-crafted Rust-only case. 
</blockquote></details>

Thanks for updating the tests to use real inference machinery. But the core concern remains: this scenario is structurally impossible via real code paths.

Mapper::map is only invoked in two places:
1. CanonicalImpl::new → via mapping.to_canonic: the value being mapped comes from a single inference context by construction
2. CanonicalImpl::embed → via mapping.from_canonic: the value is in canonical form, so all vars have InferenceId::Canonical

Neither path can produce a value containing vars from a different inference context. Your tests bypass this by calling Mapper::map directly, which is not a real usage scenario.

Adding these checks introduces dead code and, worse, silently changes the behavior when the invariant is violated — instead of surfacing the bug (via MapperError), it quietly leaves a foreign var unmapped. The right handling of an invariant violation is a panic or an error, not a no-op passthrough.

The PR should be closed.

PR #9866, opened 2026-04-23
corelib: add byte_reverse for u16/u32/u64

### LINKED ISSUE #3589: feat: u64 Byte reverse
reverse endianness of u64 is useful in many cases when dealing with keccak 

I wanted to implement it based off this https://github.com/starkware-libs/cairo-lang/blob/3a32740d0a743af6f4bfe227d1b5ecab0fc2884b/src/starkware/cairo/common/uint256.cairo#L464-L510

with 3 steps instead of 4

```
fn u64_byte_reverse(word: u64) -> u64 {
    let word: u128 = word.into();
    let word = (word & 0x00ff00ff00ff00ff00ff00ff00ff00ff) * (65535) + word;
    let word = (word & 0x00ffff0000ffff0000ffff0000ffff00) * (4294967295) + word;
    let word = (word & 0x0fffffffffffffff0000000000000000) * (18446744073709551615) + word;

    return (word / (72057594037927936)).try_into().unwrap();
}
```

the problem is that & is not implemented on felt252 and u128 gives mul overflow 

is it possible to add a u64_byte_reverse in the integer core lib ?  

thanks!!

zacharyftw:
Closes #3589. Thin wrappers over u128_byte_reverse.

### DIFF
--- a/corelib/src/integer.cairo
+++ b/corelib/src/integer.cairo
@@ -306,6 +306,39 @@ pub(crate) extern const fn u128_is_zero(a: u128) -> IsZeroResult<u128> implicits
 
 pub extern fn u128_byte_reverse(input: u128) -> u128 implicits(Bitwise) nopanic;
 
+/// Reverses the byte order of a `u16` value.
+///
+/// # Examples
+///
+/// ```
+/// assert!(u16_byte_reverse(0x0102) == 0x0201);
+/// ```
+pub fn u16_byte_reverse(input: u16) -> u16 {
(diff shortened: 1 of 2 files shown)


orizi:
Thanks for the contribution! After review, we're closing this.

The current implementation wraps `u128_byte_reverse`, which means upcasting to `u128`, reversing all 16 bytes, then extracting the relevant portion. For `u16`/`u32`/`u64` this is wasteful. A correct and efficient implementation would likely require dedicated libfuncs, which is a more involved change on our end — not something that fits the scope of a corelib-only PR. Since we also don't have concrete use cases driving the need for these functions right now, we'd prefer to hold off.

PR #9899, opened 2026-05-06
feat(constant): Added support for const calculation for inner variants and lets.

### LINKED ISSUE #9887: bug: compiler ICE bug on `match` in `const` initializer (`sierra-generator/types.rs:166`)
# Bug Report
**[Note: I would be happy to help add a PR to fix this issue. Feel free to assign it to me and I will accordingly fix the problem]**

**Cairo version:**

sierra: 1.8.0
arch: aarch64-apple-darwin

**Current behavior:**

`scarb build` panics with a Rust `unreachable!()` when a `match` expression sits at the RHS of a `const` initializer.

```
thread 'scarb compile scratch-56pigveec6cka' (2961427) panicked at cairo-lang-sierra-generator-2.18.0/src/types.rs:166:13:
Types should be fully resolved at this point. Got: `<missing>`.

thread 'main' (2961416) panicked at scarb/src/ops/compile.rs:238:14:
```

**Expected behavior:**

The 3-line PoC program should compile cleanly, with `main()` returning `1`. 

**Steps to reproduce:**

`scarb build` directly errors:
```
thread 'scarb compile scratch-56pigveec6cka' (2961427) panicked at cairo-lang-sierra-generator-2.18.0/src/types.rs:166:13:
Types should be fully resolved at this point. Got: `<missing>`.

thread 'main' (2961416) panicked at scarb/src/ops/compile.rs:238:14:
```

**Related code:**

This is a PoC for the bug:

lib.cairo:
```
const X: felt252 = match 1_u8 { _ => 1 };

#[executable]
fn main() -> felt252 { X }```
```

Scarb.toml
```
[package]
name = "scratch"
version = "0.1.0"
edition = "2025_12"

[[target.executable]]

[cairo]
enable-gas = false

[dependencies]
cairo_execute = "2.18.0"
```

NOTE: The following variants also trigger the bug:
`const X = match 1_u8 { _ => 1 }`
`const X = match 1_felt252 { _ => 1 }`
`const X = match (1_u8, 2_u8) { (_, _) => 1 }` 
`const X = match (S { v: 1 }) { S { v: _ } => 1 }` 
`const X = match () { () => 1 }` 

**Other information:**

**Root Cause and Mitigation**

### Root cause

`cairo-lang-sierra-generator-2.18.0/src/types.rs:166`:

```rust
unreachable!("Types should be fully resolved at this point. Got: `{got}`.", ...)
```

This `unreachable!()` is reached because const-eval lowering for `match` arms whose scrutinee type is not an enum-with-discrete-variants does not run the same arm-result-type inference / unification step that the regular (non-const) match-lowering path runs. Bool / `Option` / user-enum cases avoid this because their lowering goes through a variant-tag dispatch that produces a fully-resolved type early. Scalar / tuple / struct / unit cases go through the general lowering path, which leaves the result type as `<missing>`. By the time Sierra generation reaches `types.rs:166`, the expected invariant *"types are fully resolved by this point"* no longer holds.

The clean compile of `let x = match 1_u8 { _ => 1 };` proves the inference *can* succeed for this exact scrutinee+arm shape; the bug is that const-eval skips or short-circuits the inference step.

### Mitigation

1. **Reuse the `let`-side arm-result-type inference at const-RHS.** The smallest fix. The clean-compile evidence shows inference works for this shape inside a function body — the const-eval lowering needs to make the same call. Likely a one-or-two-line addition in the const-eval lowering entry, gated by a `match` discriminator. ~5–20 LoC.

2. **Convert the `unreachable!()` into a graceful diagnostic.** Smallest defensive change: replace it with `error[E....]: cannot infer type for match arm in const initializer; consider \`if\`/\`else\` instead`. ~10 LoC.

3. **Run full type inference in const-eval.** Removes this bug and a class of related const-eval inference gaps. Multi-week effort.

(1) is the right ship fix; (2) is a worthwhile defensive companion (and applies to every `<missing>` reaching this site, not just the const-eval path).

orizi:
## Summary

Extends constant evaluation to support a broader set of pattern matching constructs. The `destructure_pattern` function now returns `Option<()>` to indicate whether a pattern matched, enabling proper handling of non-matching arms. Match expressions now iterate arms using `destructure_pattern` for all pattern types (including wildcards, literals, tuples, fixed-size arrays, and structs) rather than only enum variants. `let-else` statements in constant blocks are now evaluated: if the pattern fails to match, the else clause is executed. `if let` chains and `if let` conditions similarly use the generalized `destructure_pattern` path. Boolean condition evaluation in `if` expressions now compares against cached true/false const values directly instead of extracting the enum variant.

---

## Type of change

Please check **one**:

---

## Why is this change needed?

Constant evaluation previously only handled enum variant patterns in `match` expressions and `if let` conditions. Wildcard patterns, literal patterns, tuple patterns, array patterns, and struct patterns in `match` arms were not correctly evaluated, causing incorrect or missing constant folding. Additionally, `let-else` statements inside constant blocks were not evaluated at all — the else clause was never considered, so refutable patterns with a `let-else` would silently produce wrong results instead of executing the panic branch.

---

- `match` on non-enum types (e.g., `u8`, `felt252`, tuples, structs) with wildcard or literal patterns would fail to evaluate correctly.
- `let-else` in a constant block would ignore the else clause even when the pattern did not match.
- `if let` chains were restricted to enum variant patterns only.

---

- `match` expressions in constants correctly evaluate wildcard, literal, tuple, fixed-size array, struct, and enum variant patterns.
- `let-else` in constant blocks executes the else clause (typically a panic) when the pattern does not match, producing the appropriate `Failed to calculate constant` diagnostic.
- `if let` and `if let` chains work with the full set of supported patterns.

---

Fixes #9887.

---

## Additional context

A regression test (`tests/bug_samples/issue9887.cairo`) covers wildcard, literal, tuple, struct, and unit patterns in constant `match` expressions. New constant test cases in the semantic test data cover `match` with wildcards and literals, `let-else`, `if let`, and `if let` chains with multiple conditions.

### DIFF
--- a/crates/cairo-lang-semantic/src/expr/test_data/constant
+++ b/crates/cairo-lang-semantic/src/expr/test_data/constant
@@ -17,6 +17,10 @@ const MY_CONST: felt252 = 0x1234;
 
 const FELT_TUPLE: (felt252, felt252) = (1, 2);
 
+mod my_module {
+    const CONST_IN_MODULE: felt252 = 1;
+}
+
 const FELT_FIXED_SIZE_ARRAY: [felt252; 3] = [1, 2, 3];
 
 const OTHER_CONST_REF: felt252 = my_module::CONST_IN_MODULE;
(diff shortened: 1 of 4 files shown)


orizi:
* **#9901** <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/9901?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#9899** <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/9899?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/9899?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

TomerStarkware (approved):
:lgtm:

> @TomerStarkware reviewed 4 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/9899#-:-OsG7oQC-4Lky5gsYwhn:bnfp4nl)* status: all files reviewed, 3 unresolved discussions (waiting on orizi).

TomerStarkware (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/9899#-:-OsGRf8kAq6QFFFz8uPT:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on orizi).

orizi (commented):
> @orizi resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/9899)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on orizi).

PR #9901, opened 2026-05-06
bugfix(semantic): fix ICE on nested const fn in const fn.

orizi:
## Summary

Fixes an ICE (Internal Compiler Error) that occurred when a `const fn` was nested inside another `const fn`. The fix adds cycle recovery to `priv_free_function_body_data` and replaces a panicking `extract_matches!` call with a graceful fallback that returns a missing value when the generic function is not an `Impl` variant.

---

## Type of change

Please check **one**:

- [x] Bug fix (fixes incorrect behavior)
- [ ] New feature
- [ ] Performance improvement
- [ ] Documentation change with concrete technical impact
- [ ] Style, wording, formatting, or typo-only change

---

## Why is this change needed?

When a `const fn` was nested inside another `const fn`, the compiler would ICE instead of emitting proper parser diagnostics. Two root causes were addressed:

1. `priv_free_function_body_data` had no cycle recovery, causing a panic when a cyclic dependency was encountered during evaluation of nested `const fn` bodies.
2. In `ConstantEvaluateContext`, `extract_matches!` would panic if `concrete_function.generic_function` was not a `GenericFunctionId::Impl`, which could happen in the presence of malformed or unsupported constructs.

---

## What was the behavior or documentation before?

Writing a nested `const fn` inside a `const fn` caused an ICE rather than emitting parser diagnostics and continuing gracefully.

---

## What is the behavior or documentation after?

The compiler now emits the expected parser diagnostics (e.g., unexpected token, missing `=`, reserved identifier `fn`, etc.) without crashing. A new test case documents the expected diagnostic output for this pattern.

---

## Related issue or discussion (if any)

Fixes https://github.com/starkware-libs/cairo/issues/9889

---

## Additional context

The cycle recovery functions (`priv_free_function_body_data_cycle` and `priv_free_function_body_data_initial`) follow the Salsa cycle-handling pattern: the initial value is `Err(skip_diagnostic())` and the cycle function propagates the last provisional value.

PR #9907, opened 2026-05-07
fix(semantic): graceful diagnostic for impl-method calls in const fn …

### LINKED ISSUE #9906: bug: ICE: const fn returning Array/Box/Span ICEs in const initializer ('Unexpected function call in constant lowering' at constant.rs:964)
# Bug Report

**Cairo version:**

sierra: 1.8.0
arch: aarch64-apple-darwin

**Current behavior:**

**scarb build** produces this panic:
```

thread '<unnamed>' (156068) panicked at /Users/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/cairo-lang-semantic-2.18.0/src/items/constant.rs:964:17:
internal error: entered unreachable code: Unexpected function call in constant lowering: ExprFunctionCall { function: FunctionId(32c05), args: [], coupon_arg: None, ty: TypeId(1f80e), stable_ptr: ExprPtr(SyntaxStablePtrId(SyntaxNode(30d40))) }

thread 'main' (156027) panicked at scarb/src/ops/compile.rs:238:14:
```

**Expected behavior:**

`const X: Array<felt252> = f()` should produce the same `error[E2127]: This expression is not supported as constant.` Cairo already emits for the non-const-fn version of the same code. Or, if the team wants to support compile-time-evaluated heap types as a future feature, it should compile cleanly. Either way, never a Rust backtrace.

**Steps to reproduce:**

just run `scarb build` on the following code:

**Related code:**

```cairo
const fn f() -> Array<felt252> { array![1, 2, 3] }
const X: Array<felt252> = f();

#[executable]
fn main() {}
```

**Other information:**

Other variants tested:
| Variant | Triggers? |
|---|---|
| `const X: Array<T> = f()` (baseline) | ✅ |
| `const X: Box<T> = f()` | ✅ |
| `const X: Span<T> = f()` | ✅ |
| `const X: Box<Box<T>> = f()` (nested) | ✅ |
| `const X: (felt, Array<T>) = f()` (tuple containing Array) | ✅ |
| `const X: Option<Array<T>> = f()` (Option of Array) | ✅ |
| `const X: Holder { arr: Array<T> } = f()` (struct containing Array) | ✅ |
| `const X: felt252 = f()` (felt return) | ❌ clean |
| `const X: u256 = f()` | ❌ clean |
| `const X: Pt = f()` (struct of value-types) | ❌ clean |
| `fn main() { let _arr = f(); }` (call NOT in const) | ❌ E2127 graceful |
| `const X = ArrayTrait::new()` (direct call, no const-fn indirection) | ❌ E2127 graceful |

## Root Cause and Mitigation

### Root cause

`cairo-lang-semantic-2.18.0/src/items/constant.rs:964`:

```rust
_ => unreachable!("Unexpected function call in constant lowering: {:?}", expr),
```

The const-eval admission gate (`is_const_expr` / `is_supported_in_const_position`) accepts any user-defined `const fn` call without inspecting the return type. The lowering loop in `constant.rs` only handles value-typed call results, so heap-returning `const fn` calls fall through to this `unreachable!()` arm.

The same call shape outside a const initializer is correctly rejected by the more conservative outer `is_const_expr` check, producing E2127. So this is fundamentally an **inter-stage gate inconsistency** — the inner gate (around `const fn` admission) is laxer than the outer one.

A direct corelib call in const position (`const X = ArrayTrait::new();`) is also correctly rejected with E2127. Only the **`const fn` indirection** confuses the gate.

### Mitigation

1. **Replace the `unreachable!()` with a graceful diagnostic AND tighten the outer gate.** At line 964:
   ```rust
   _ => return Err(diagnostics.report(
       expr.stable_ptr().untyped(),
       SemanticDiagnosticKind::ConstFnReturnsHeapType,
   )),
   ```
   And add the heap-return check to the outer gate so this is caught before lowering, not at lowering. ~20 LoC.

2. **Implement compile-time evaluation for heap-allocated types.** This is a substantial feature, not a bug fix. Out of scope for an ICE-only patch; multi-week effort.

3. **Add a debug-build assertion that "outer-admitted ⊆ inner-lowered."** A defensive structural invariant — every shape admitted by the outer gate should be in the inner lowering's enumeration. Stating it explicitly catches future drift at CI time. ~20 LoC.

(1) is sufficient to ship; user-facing impact becomes identical to today's E2127 path.

anon-researchers-123:
Fixes #9906 

Replaces the `unreachable!()` in `evaluate_function_call`
(`crates/cairo-lang-semantic/src/items/constant.rs`) with a graceful
`UnsupportedConstant` (E2127) diagnostic, mirroring the existing extern
fallback in the same function. The arm is reachable when the inner const
evaluator encounters an impl-method call it has no lowering rule for —
e.g. `ArrayTrait::new` reached via `array![...]` desugared inside a
user-defined `const fn` body — and currently produces a Rust backtrace
instead of a compiler error. Adds a regression fixture in
`expr/test_data/constant`.

## Why is this change needed?

The arm at the bottom of `evaluate_function_call`'s `match imp.function`
was annotated `unreachable!("Unexpected function call in constant
lowering: ...")`, but it is in fact reachable: when a user-defined
`const fn` body uses `array![...]`, the desugared `ArrayTrait::new()`
call has no const-evaluation handler and falls through to this arm. The
result on `scarb 2.18.0` is a Rust panic with the user-facing message
"Compiler thread has panicked" — not a compiler diagnostic — for code
that should produce the same `error[E2127]` Cairo already emits when
`const X = ArrayTrait::new()` is written directly (without the
`const fn` indirection).

### DIFF
--- a/crates/cairo-lang-semantic/src/expr/test_data/constant
+++ b/crates/cairo-lang-semantic/src/expr/test_data/constant
@@ -850,3 +850,64 @@ error[E2127]: This expression is not supported as constant.
  --> lib.cairo:5:26
     const VALUE: bool = !F::VALUE;
                          ^^^^^^^^
+
+//! > ==========================================================================
+
+//! > Regression: user-defined const fn returning a heap type - graceful diagnostic, not ICE.
+
+//! > test_runner_name
+test_function_diagnostics(expect_diagnostics: true)
+
+//! > function_code
+fn foo() {}
+
+//! > function_name
+foo
+
+//! > module_code
+const fn f() -> Array<felt252> {
+    array![1, 2, 3]
+}
+const X: Array<felt252> = f();
+
+//! > expected_diagnostics
+error[E2127]: This expression is not supported as constant.
+ --> lib.cairo:2:5
+    array![1, 2, 3]
+    ^^^^^^^^^^^^^^^
+
+error[E2130]: This expression is not supported as constant.
+ --> lib.cairo:4:27
+const X: Array<felt252> = f();
+                          ^^^
+note: In `test::f`:
+  --> lib.cairo:2:5
+    array![1, 2, 3]
+    ^^^^^^^^^^^^^^^
+
+error[E2130]: This expression is not supported as constant.
+ --> lib.cairo:4:27
+const X: Array<felt252> = f();
+                          ^^^
+note: In `test::f`:
+  --> lib.cairo:2:5
+    array![1, 2, 3]
+    ^^^^^^^^^^^^^^^
+
+error[E2130]: This expression is not supported as constant.
+ --> lib.cairo:4:27
+const X: Array<felt252> = f();
+                          ^^^
+note: In `test::f`:
+  --> lib.cairo:2:5
+    array![1, 2, 3]
+    ^^^^^^^^^^^^^^^
+
+error[E2130]: This expression is not supported as constant.
+ --> lib.cairo:4:27
+const X: Array<felt252> = f();
+                          ^^^
+note: In `test::f`:
+  --> lib.cairo:2:5
+    array![1, 2, 3]
+    ^^^^^^^^^^^^^^^
--- a/crates/cairo-lang-semantic/src/items/constant.rs
+++ b/crates/cairo-lang-semantic/src/items/constant.rs
@@ -952,10 +952,12 @@ impl<'a, 'r, 'mt> ConstantEvaluateContext<'a, 'r, 'mt> {
                 .intern(db);
             }
             _ => {
-                unreachable!(
-                    "Unexpected function call in constant lowering: `{:?}`",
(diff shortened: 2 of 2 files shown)


orizi (changes requested):
> @orizi made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/9907)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on anon-researchers-123).

___
*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/9907#-OsGSzukDiItiX6K6pCe:-OsGSzul0sjt5bWCHkgs:b-qkx6a9) (no related file):*
already handled at #9901 without the extra diagnostics that are out of scope.

PR #9909, opened 2026-05-07
bugfix(semantic): const-eval no longer ICEs on free-function const fn…

### LINKED ISSUE #9908: bug: ICE: const fn { panic!("...") } in const initializer panics const-eval at `constant.rs:891`
# Bug Report

**Cairo version:**

sierra: 1.8.0
arch: aarch64-apple-darwin

**Current behavior:**

A compiler panic is observed:

 **scarb build** produces:
```

thread '<unnamed>' (316147) panicked at /Users/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/cairo-lang-semantic-2.18.0/src/items/constant.rs:891:19:
Variant extract failed: `Free(FreeFunctionId(e827))` is not of variant `GenericFunctionId::Impl`

thread 'main' (316115) panicked at scarb/src/ops/compile.rs:238:14:
```

**Expected behavior:**

The 4-line program should produce the same `error[E2130]: Failed to calculate constant: panic at runtime` diagnostic that `Option::None.unwrap()` and `1 / 0` already produce in const position. Or, alternatively, be rejected at the const-fn-call gate with a different graceful diagnostic. Never a Rust backtrace.

**Steps to reproduce:**

`scarb build` on the following code:

**Related code:**

```
const fn f() -> felt252 { panic!("nope") }
const X: felt252 = f();

#[executable]
fn main() -> felt252 { X }
```

**Other information:**

Some variant shapes:

| Variant | Triggers?| Else |
|---|---|---|
| `const fn f() -> felt252 { panic!("nope") }` (string-arg) | ✅ | | 
| `const fn f() -> felt252 { panic!() }` (no-arg) | ✅ | | 
| `helper() -> felt252 { panic!("h") }; f() -> felt252 { helper() }` | ✅ | | 
| `const X = f();` (declared but unused) | ✅ | | 
| `fn f() -> felt252 { panic!("nope") }` (NOT const fn) | || runtime panic — correct |
| `const fn f() -> felt252 { Option::<felt252>::None.unwrap() }` |  | **E2130 graceful** |
| `const fn f() -> u32 { 1 / 0 }` (compile-time div-0) | |  **E2130 graceful** |

### Root cause

`cairo-lang-semantic-2.18.0/src/items/constant.rs:891`:

```rust
let impl_function = extract_matches!(generic_function, GenericFunctionId::Impl);
```

This unconditionally extracts the `Impl(_)` variant from `GenericFunctionId`. The assumption is "if we reach this point, the call is a trait-impl method call." That holds for ordinary user-defined `const fn` calls. But `panic!("string-literal-arg")` desugars into a **free-function** call (`core::panics::panic_with_byte_array` or similar `core::panics::*`), which is a `Free(FreeFunctionId)` variant, not `Impl(_)`. The `extract_matches!` aborts.

### Mitigation

1. **Replace `extract_matches!` at line 891 with a `match`** that handles `Free(_)` gracefully:
   ```rust
   match generic_function {
       GenericFunctionId::Impl(impl_id) => { /* existing path */ }
       GenericFunctionId::Free(_) => return Err(diagnostics.report(
           expr.stable_ptr().untyped(),
           SemanticDiagnosticKind::ConstantCalculationFailed("evaluation panicked".into()),
       )),
       // …other variants…
   }
   ```
   ~10 LoC; removes the visible panic.

2. **Route `panic!`-bodied const fn into the existing E2130 diagnostic.** Treat `panic!()` (and other compile-time-unreachable expressions) uniformly with `Option::None.unwrap()` and `1/0`. The const-fn-with-panic-body pattern is a natural Rust-port idiom for "unreachable const stub"; making it diagnose like its siblings would be a UX win. ~20 LoC.

3. **Audit other `extract_matches!` calls in const-eval.** Each unconditional variant extraction is a brittle assumption about which variants flow into a path. A grep through `cairo-lang-semantic/src/items/constant.rs` for sibling `extract_matches!` calls — and demoting them in batch to graceful `match` — is a hygiene PR that prevents this exact bug class from re-appearing. ~1 day.

(1) alone fixes the ICE; (2) makes the diagnostic informative; (3) is preventive maintenance.

anon-researchers-123:
… body

Fixes #9908 

A `const fn` whose body is `panic!("...")` (string-literal arg) crashes the
compiler with a Rust `extract_matches!` assertion failure in
`cairo-lang-semantic/src/items/constant.rs`. This PR replaces the unconditional
`extract_matches!(GenericFunctionId::Impl)` in the const-eval call path with a
graceful `let-else`, so the compiler now emits a regular
`error[E2130]: Failed to calculate constant.` diagnostic instead of a Rust
backtrace.

---

## Type of change

---

## Why is this change needed?

`scarb build` produces a Rust backtrace on a 4-line program:

```cairo
const fn f() -> felt252 { panic!("nope") }
const X: felt252 = f();

#[executable]
fn main() -> felt252 { X }
```

This exposes a compiler `panic` to users, indistinguishable from a true
compiler crash. The `const fn` body of `panic!(...)` is a natural Rust-port
idiom for "unreachable const stub", so the surface is a real one, not contrived.

---

## What was the behavior before?

Compiler panic:

```
thread '<unnamed>' panicked at .../cairo-lang-semantic/src/items/constant.rs:891:19:
Variant extract failed: `Free(FreeFunctionId(...))` is not of variant `GenericFunctionId::Impl`

thread 'main' panicked at scarb/src/ops/compile.rs:238:14:
```

Notably, sister "compile-time-unreachable" const-fn bodies were already handled
gracefully:

```cairo
const fn f() -> felt252 { Option::<felt252>::None.unwrap() }   // → error[E2130] graceful
const fn f() -> u32 { 1 / 0 }                                  // → error[E2130] graceful
```

Only the `panic!("string")` shape (and `panic!()` with no arg) ICEd, because
those macros desugar into a free-function call to
`core::panics::panic_with_byte_array` rather than into the operator/extern
paths the const-eval pipeline already covered.

---

Regression test: A new case in crates/cairo-lang-semantic/src/expr/test_data/constant covers both panic!("nope") (string-arg) and panic!() (no-arg) const-fn bodies, asserting the wrapped E2130 diagnostic with a note: In <fn>: chain — the same shape assert(false) and my_unwrap(...) already produce in that file. It runs under the existing expr_diagnostics::constant suite (no new harness needed). The test was confirmed to be a genuine regression test by reverting the fix and observing the original ICE signature reproduce verbatim.

### DIFF
--- a/crates/cairo-lang-semantic/src/expr/test_data/constant
+++ b/crates/cairo-lang-semantic/src/expr/test_data/constant
@@ -850,3 +850,62 @@ error[E2127]: This expression is not supported as constant.
  --> lib.cairo:5:26
     const VALUE: bool = !F::VALUE;
                          ^^^^^^^^
+
+//! > ==========================================================================
+
+//! > Const fn whose body desugars to a free-function call (issue #9908).
+
+//! > test_runner_name
+test_function_diagnostics(expect_diagnostics: true)
+
+//! > function_code
+fn foo() {}
+
+//! > function_name
+foo
+
+//! > module_code
+// `panic!("string")` desugars to a call to `core::panics::panic_with_byte_array`,
+// a free function. Const-eval used to panic at `evaluate_function_call` when the
+// trait-impl extraction of `GenericFunctionId` met a `Free(_)` variant; it must
+// now produce a regular const-evaluation diagnostic instead.
+const fn panic_string_body() -> felt252 {
+    panic!("nope")
+}
+
+const PANIC_STRING_CALLED: felt252 = panic_string_body();
+
+const fn panic_no_arg_body() -> felt252 {
+    panic!()
+}
+
+const PANIC_NO_ARG_CALLED: felt252 = panic_no_arg_body();
+
+//! > expected_diagnostics
+error[E2127]: This expression is not supported as constant.
+ --> lib.cairo:6:5
+    panic!("nope")
+    ^^^^^^^^^^^^^^
+
+error[E2130]: Failed to calculate constant.
+ --> lib.cairo:9:38
+const PANIC_STRING_CALLED: felt252 = panic_string_body();
+                                     ^^^^^^^^^^^^^^^^^^^
+note: In `test::panic_string_body`:
+  --> lib.cairo:6:5
+    panic!("nope")
+    ^^^^^^^^^^^^^^
+
+error[E2127]: This expression is not supported as constant.
+ --> lib.cairo:12:5
+    panic!()
+    ^^^^^^^^
+
+error[E2130]: Failed to calculate constant.
+ --> lib.cairo:15:38
+const PANIC_NO_ARG_CALLED: felt252 = panic_no_arg_body();
+                                     ^^^^^^^^^^^^^^^^^^^
+note: In `test::panic_no_arg_body`:
+  --> lib.cairo:12:5
+    panic!()
+    ^^^^^^^^
--- a/crates/cairo-lang-semantic/src/items/constant.rs
+++ b/crates/cairo-lang-semantic/src/items/constant.rs
@@ -898,7 +898,17 @@ impl<'a, 'r, 'mt> ConstantEvaluateContext<'a, 'r, 'mt> {
             return calc_result;
         }
 
-        let imp = extract_matches!(concrete_function.generic_function, GenericFunctionId::Impl);
+        let GenericFunctionId::Impl(imp) = concrete_function.generic_function else {
+            // The const-eval pipeline only knows how to evaluate trait-impl operator calls past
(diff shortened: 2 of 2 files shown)


orizi (changes requested):
> @orizi made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/9909)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on anon-researchers-123).

___
*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/9909#-OsGTA9c31pJt4kk2zj0:-OsGTA9c31pJt4kk2zj1:b-5n9dzt) (no related file):*
handled at #9901 without the diagnostics that are out of scope.

PR #9916, opened 2026-05-10
refactor(lowering): Made function not return Maybe and possibly hide diags.

orizi:
## Summary

`function_with_body_lowering_diagnostics` now returns `Diagnostics` directly instead of `Maybe<Diagnostics>`. The `in_cycle` error case is handled inline using `matches!(..., Ok(true))` rather than propagating via `?`, allowing the function to be infallible. All call sites are updated to remove `.unwrap_or_default()` and `?` unwrapping accordingly.

---

## Type of change

Please check **one**:

- [ ] New feature
- [ ] Performance improvement

---

## Why is this change needed?

`function_with_body_lowering_diagnostics` was returning `Maybe<Diagnostics>` despite having no meaningful failure path — the only `?` usage was on `db.in_cycle(...)`, which on error would silently swallow all diagnostics collected up to that point. This meant a query failure could cause legitimate diagnostics to be lost rather than reported.

---

If `db.in_cycle(function_id, DependencyType::Cost)` returned an `Err`, the function would early-return `Err`, discarding any diagnostics already accumulated in the builder. Call sites used `.unwrap_or_default()` to recover, silently dropping those diagnostics.

---

The function is now infallible and always returns a `Diagnostics` value. The `in_cycle` result is checked with `matches!(..., Ok(true))`, so errors are treated as "not in cycle" without aborting diagnostic collection. All diagnostics gathered before the cycle check are preserved and returned regardless of whether `in_cycle` succeeds.

---

---

## Additional context

The test helper in `block_generator_test.rs` is also updated to remove the now-unnecessary `.unwrap()` call.

### DIFF
--- a/crates/cairo-lang-lowering/src/db.rs
+++ b/crates/cairo-lang-lowering/src/db.rs
@@ -136,7 +136,7 @@ pub trait LoweringGroup: Database {
     fn function_with_body_lowering_diagnostics<'db>(
         &'db self,
         function_id: ids::FunctionWithBodyId<'db>,
-    ) -> Maybe<Diagnostics<'db, LoweringDiagnostic<'db>>> {
+    ) -> Diagnostics<'db, LoweringDiagnostic<'db>> {
         function_with_body_lowering_diagnostics(self.as_dyn_database(), function_id)
     }
     /// Aggregates semantic function level lowering diagnostics - along with all its generated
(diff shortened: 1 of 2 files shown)


orizi:
* **#9916** <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/9916?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/9916?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

TomerStarkware (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/9916#-:-OsGRE490mLYyOTrL2nu:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on orizi).

PR #9921, opened 2026-05-12
fix: point implicit unit return mismatch at tail expr

### LINKED ISSUE #6486: bug: Incorrect diagnostic placement when returning wrong type
# Bug Report

**Cairo version:**

2.8.2

**Current behavior:**

```cairo
fn incorrect_return_type() {
    let x = 5;
    x
}
```

Diagnostic point to the variable definition and blame the impossibility of casting

```
error: Mismatched types. The type `()` cannot be created from a numeric literal.
 --> {SOME_PATH}:16:17
        let x = 5;
                ^
```

**Expected behavior:**

Imo it should show the actual root cause of the issue. That's how it looks in Rust:

```
error[E0308]: mismatched types
  --> crates/cairo-lang-language-server/src/lang/diagnostics/mod.rs:98:9
   |
96 |     fn incorrect_return_type() {
   |                               - expected `()` because of default return type
97 |         let x = 5;
98 |         x
   |         ^ expected `()`, found integer
```

Elena343-ai:
## Summary

Reports implicit unit-return mismatches on the inferred tail expression instead of forcing an unconstrained numeric literal to conform to `()`.

---

## Type of change

Please check **one**:

- [ ] New feature
- [ ] Performance improvement

---

## Why is this change needed?

For a function without an explicit return type, a trailing expression should be diagnosed as the value that does not match the implicit `()` return type. When the tail expression was an inferred numeric local, the compiler instead reported that `()` could not be created from the numeric literal at the variable definition.

---

In code like:

```cairo
fn foo() {
    let x = 5;
    x
}
```

the diagnostic pointed at `5` in `let x = 5`.

---

The diagnostic reports `Unexpected return type. Expected: "()", found: "core::felt252".` on the tail expression `x`, matching the actual source of the return-type mismatch.

---

Fixes #6486

---

## Additional context

Adds semantic regression coverage for the implicit unit-return case while preserving the existing diagnostic locations for explicit return types and already concrete tail expressions.

### DIFF
--- a/crates/cairo-lang-semantic/src/expr/compute.rs
+++ b/crates/cairo-lang-semantic/src/expr/compute.rs
@@ -1734,25 +1734,48 @@ pub fn compute_root_expr<'db>(
         ctx.db.generic_params_type_constraints(ctx.resolver.data.generic_params.clone());
     inference.conform_generic_params_type_constraints(constrains);
 
+    let db = ctx.db;
     let return_type = ctx.reduce_ty(return_type);
     let res = compute_expr_block_semantic(ctx, syntax)?;
     let res_ty = ctx.reduce_ty(res.ty());
     let res = ctx.arenas.exprs.alloc(res);
+    let implicit_unit_return = ctx
+        .signature
+        .map(|s| matches!(s.stable_ptr.lookup(db).ret_ty(db), OptionReturnTypeClause::Empty(_)))
+        .unwrap_or(false)
+        && return_type.is_unit(db);
+    let tail_expr_stable_ptr = if implicit_unit_return {
+        let (_, tail) = statements_and_tail(db, syntax.statements(db));
+        tail.map(|tail| tail.expr(db).stable_ptr(db).untyped())
+    } else {
+        None
+    };
+    let prefer_tail_return_diagnostic = tail_expr_stable_ptr.is_some() && !res_ty.is_var_free(db);
     let inference = &mut ctx.resolver.inference();
-    let db = ctx.db;
+    if prefer_tail_return_diagnostic {
+        let _ = inference.finalize_numeric_literals();
+    }
+    let res_ty = inference.rewrite(res_ty).no_err();
+    let tail_return_diag_stable_ptr = if prefer_tail_return_diagnostic && res_ty.is_var_free(db) {
+        tail_expr_stable_ptr
+    } else {
+        None
+    };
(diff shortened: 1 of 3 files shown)


orizi:
The approach here is wrong because it calls `finalize_numeric_literals()` eagerly, in the middle of type inference, before `conform_ty_for_diag`. Numeric literal finalization is supposed to happen exactly once, at the very end of inference (as part of `finalize_without_reporting` → `finalize`), after all type constraints from the entire function body have been gathered. Pulling it forward into `compute_root_expr` changes the order of operations for every implicit-unit-return function that has an unresolved inference variable in its tail expression.

Concretely: if the tail expression's type variable `?T` would be constrained to something concrete (e.g. `u32`) by a downstream constraint that hasn't fired yet, this early finalization locks it to `felt252` first, potentially masking a real type error or producing a wrong one.

The PR also splits `finalize_without_reporting` into `finalize_numeric_literals` + the rest purely to enable this premature call — a red flag that the abstraction is being twisted to serve a hack.

**The actual fix** is simpler and doesn't touch inference ordering: when the return type clause is `Empty(_)`, the diagnostic stable pointer should point at the tail expression (if one exists), not the whole block syntax. That one-line change to the existing `conform_ty_for_diag` closure fixes the diagnostic location without perturbing inference. The root cause of the *wrong diagnostic type* (`()` cannot be created from integer literal instead of `WrongReturnType`) is that `conform_ty_for_diag` propagates the `() = ?T` constraint backward — that's what needs to be addressed, not worked around by pre-resolving the variable.

Elena343-ai:
Thanks, that makes sense. I’ll rework this.

Elena343-ai:
@orizi Updated as request, please take another look, ty!

orizi (changes requested):
> @orizi made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/9921)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on Elena343-ai).

___
*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/9921#-Ot37OwBDTzTRqg2J1Li:-Ot37OwBDTzTRqg2J1Lj:b-2xku73) (no related file):*
i believe it was fixed in another set of PRs - please rebase to make sure.

orizi:
no response.
Closing.

PR #9956, opened 2026-05-18
Avoid panic on inline macro calls with missing arguments

### LINKED ISSUE #9938: bug: Compiler panics at `macro_declaration.rs:330` (`internal error: entered unreachable code`)
# Bug Report

**Cairo version:**

sierra: 1.8.0
arch: aarch64-apple-darwin

**Current behavior:**

When a user-defined macro is declared and then invoked at module/item scope **without** any of the three argument-bracket forms (`(...)`, `{...}`, or `[...]`), the parser correctly emits `E1006: Missing tokens. Expected an argument list wrapped in either parentheses, brackets, or braces.`, and then the semantic pass panics afterwards at `crates/cairo-lang-semantic/src/items/macro_declaration.rs:330:46`:

```
scarb build
error[E1006]: Missing tokens. Expected an argument list wrapped in either parentheses, brackets, or braces.
 --> src/lib.cairo:7:3
m!                       // <-- panic here
  ^

error[E1001]: Missing token ';'.
 --> src/lib.cairo:7:3
m!                       // <-- panic here
  ^


thread '<unnamed>' (495631) panicked at /Users/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/cairo-lang-semantic-2.18.0/src/items/macro_declaration.rs:330:46:
internal error: entered unreachable code

thread 'main' (495597) panicked at scarb/src/ops/compile.rs:238:14:
```

**Expected behavior:**

The item-scope expansion path should match the function-body-scope path: the parser's `E1006` is the only diagnostic the user should see. The compiler should not continue into `is_macro_rule_match` after the parser has emitted a recovery error on the macro arguments. The expected output is just the two parser errors with no panic.

**Steps to reproduce:**

run `scarb build` on the following PoC:

**Related code:**

lib.cairo:
```
macro m {
    ($x:ident) => { 1 };
}
m!                       // panics here

#[executable]
fn main() -> felt252 { 0 }
```

Scarb.toml:
```
[package]
name = "scratch"
version = "0.1.0"
edition = "2025_12"
experimental-features = ["user_defined_inline_macros"]

[[target.executable]]

[cairo]
enable-gas = false

[dependencies]
cairo_execute = "2.18.0"

```

**Other information:**

Five trigger variants all produce the same panic at `:330:46`:

| invocation | result |
|---|---|
| `m!` | PANIC `:330` |
| `m!foo` | PANIC `:330` |
| `m!42` | PANIC `:330` |
| `m! "str"` | PANIC `:330` |
| `m!:ident` | PANIC `:330` |

The corresponding **function-body-scope** invocation does NOT panic: it emits the same `E1006` and stops cleanly. So the bug is specific to the item-scope expansion path:

```cairo
macro m { ($x:ident) => { 1 }; }
fn f() { m! }   // <-- E1006 only, no panic
```

## Root Cause and Mitigation

### Root cause

`is_macro_rule_match` in `cairo-lang-semantic/src/items/macro_declaration.rs` asserts via `unreachable!()` that the input `TokenTreeNode`'s subtree is one of the three concrete bracketing variants. The assertion is wrong for the item-scope path: the parser already emits `E1006 Missing tokens` and constructs an `ItemInlineMacro` whose subtree is `WrappedTokenTree::Missing(_)`. The semantic pass then runs anyway and hits the unreachable.

### Mitigation

The smallest correct change is to replace the unreachable with `return None`, so the macro rule match cleanly fails on parser-recovered inputs and the caller's existing "no rule matched" path emits a normal diagnostic (which `E1006` already covers):

```rust
let mut input_iter = match input.subtree(db) {
    ast::WrappedTokenTree::Parenthesized(tt) => tt.tokens(db),
    ast::WrappedTokenTree::Braced(tt) => tt.tokens(db),
    ast::WrappedTokenTree::Bracketed(tt) => tt.tokens(db),
    ast::WrappedTokenTree::Missing(_) => return None,    // <-- defensive None
}
```

A structurally cleaner alternative is to gate the inline-macro expansion pass entirely: if the `ItemInlineMacro`'s `arguments.subtree(db)` is `Missing(_)`, skip the expansion attempt for that item — the parser already reported the error. Either fix is sound; the one-line `return None` is the minimum surface change.

A regression test would live next to the macro-declaration test data — a row that defines a one-rule macro and invokes it at item scope without brackets, expecting only `E1006` with no panic.

1sgtpepper:
Fixes #9938.

## Summary
This fixes a compiler panic for malformed item-level user-defined inline macro calls like `m!;`, where the parser recovers a missing argument list.

When the macro call has no parsed argument token tree, semantic analysis now stops before resolving or expanding the macro. The parser diagnostic is kept as the only user-facing error, and semantic recovery no longer reaches macro rule matching with a missing token tree.

The matcher also defensively treats a missing wrapped token tree as “no match” instead of panicking if that path is reached directly.

## Tests
Added regression coverage for:
- a declared item-level macro called as `m!;`
- an unresolved item-level macro called as `m!;`

Verified with:
- `cargo test -p cairo-lang-semantic inline_macros --lib`
- `./scripts/rust_fmt.sh`
- `./scripts/clippy.sh`
- `git diff --check`

### DIFF
--- a/crates/cairo-lang-semantic/src/expr/test_data/inline_macros
+++ b/crates/cairo-lang-semantic/src/expr/test_data/inline_macros
@@ -1905,6 +1905,69 @@ macro_item_level!();
 
 //! > ==========================================================================
 
+//! > Test item-level macro invocation with missing argument list.
+
+//! > test_runner_name
+test_function_diagnostics(expect_diagnostics: true)
+
+//! > crate_settings
+edition = "2024_07"
+[experimental_features]
+negative_impls = false
+associated_item_constraints = false
+coupons = false
+user_defined_inline_macros = true
+
+//! > function_code
+fn foo() {}
+
+//! > function_name
+foo
+
+//! > module_code
+macro m {
+    ($x:ident) => { 1 };
+}
+m!;
+
+//! > expected_diagnostics
+error[E1006]: Missing tokens. Expected an argument list wrapped in either parentheses, brackets, or braces.
+ --> lib.cairo:4:3
+m!;
+  ^
+
+//! > ==========================================================================
+
+//! > Test unresolved item-level macro invocation with missing argument list.
+
+//! > test_runner_name
+test_function_diagnostics(expect_diagnostics: true)
+
+//! > crate_settings
+edition = "2024_07"
+[experimental_features]
+negative_impls = false
+associated_item_constraints = false
+coupons = false
+user_defined_inline_macros = true
+
+//! > function_code
+fn foo() {}
+
+//! > function_name
+foo
+
+//! > module_code
+m!;
+
+//! > expected_diagnostics
+error[E1006]: Missing tokens. Expected an argument list wrapped in either parentheses, brackets, or braces.
+ --> lib.cairo:1:3
+m!;
+  ^
+
+//! > ==========================================================================
+
 //! > Test macro with use-star resolution.
 
 //! > test_runner_name
--- a/crates/cairo-lang-semantic/src/items/macro_call.rs
+++ b/crates/cairo-lang-semantic/src/items/macro_call.rs
@@ -43,16 +43,29 @@ fn priv_macro_call_data<'db>(
     let module_id = macro_call_id.parent_module(db);
     let mut resolver = Resolver::new(db, module_id, inference_id);
     let macro_call_syntax = db.module_macro_call_by_id(macro_call_id)?;
+    let callsite_module_id = module_id;
+    let arguments = macro_call_syntax.arguments(db);
+    // The parser already reports the missing argument list. Avoid resolving or expanding a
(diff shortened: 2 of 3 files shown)


orizi (changes requested):
> @orizi made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/9956)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on Kuhai9801).

___
*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/9956#-Ot1BCWm0CDU3oUSnWnz:-Ot1BCWm0CDU3oUSnWo-:bl5289x) (no related file):*
i believe this is already solved.

PR #9977, opened 2026-05-21
Avoid ICE for generic numeric literal suffixes

### LINKED ISSUE #9975: bug: Numeric literal suffix with generic type ICEs compiler
# Bug Report

**Cairo version:**

`6a3c423a9`

**Current behavior:**

`cairo-compile` panics with an itertools `.zip_eq()` invariant violation when a numeric literal uses a corelib `pub enum` / `pub struct` type with required generic parameters as its suffix. The four affected type names are `Span`, `Option`, `Result`, and `PanicResult`. The crash exits without a clean diagnostic E-code:

```
thread 'main' panicked at zip_eq_impl.rs:50:17:
itertools: .zip_eq() reached end of one iterator before the other
```

An empirical sweep over 27 corelib type names used as numeric-literal suffixes (`42_<TypeName>`) at the pinned commit produces a 5-category typology:

- **Cat A — ICE (4 types):** `Span`, `Option`, `Result`, `PanicResult` — itertools `.zip_eq()` panic, no clean E-code. All are corelib `pub enum` / `pub struct` types with required generic parameters.
- **Cat B — clean compile (12 types):** `u8`, `u16`, `u32`, `u64`, `u128`, `u256`, `i8`, `i16`, `i32`, `i64`, `i128`, `felt252` — the legitimate numeric-literal-suffix types.
- **Cat C — E2008 graceful (8 types):** `Array`, `Box`, `Felt252Dict`, `Nullable`, `bool`, `ByteArray`, `Pedersen`, `EcPoint` — mostly `pub extern type`; matcher emits `'A numeric literal of type <T> cannot be created.'`
- **Cat D — E2006 not found (2 types):** `Bytes31`, `EnumeratedIter`.
- **Cat E — trait/non-type (4 names):** `Default`, `Drop`, `Copy` → E2011; `NonZero` → E2030.

The inconsistency: Cat-C generic types declared as `pub extern type` (e.g., `pub extern type Array<T>;`) take a branch in the suffix matcher that emits E2008, while Cat-A `pub enum` / `pub struct` types with required generics fall into the `.zip_eq()` path where a length mismatch (e.g., 1 expected vs. 0 supplied generic args for `Option<T>`) panics.

PR #9810 (commit `a459dfa0e`, April 2026, included in `6a3c423a9`) emits a diagnostic for non-type identifier suffixes instead of panicking, but it does not cover the Cat-A `pub enum` / `pub struct`-with-required-generics case, which still ICEs.

**Expected behavior:**

`cairo-compile` should emit a graceful diagnostic (e.g., the same `error[E2008]: A numeric literal of type <T> cannot be created.` already produced for Cat-C extern types) instead of panicking with an internal `.zip_eq()` invariant violation. Suffix matching for `pub enum` / `pub struct` corelib types with required generics should be consistent with the extern-type path.

**Steps to reproduce:**

1. Build / use `cairo-compile` at commit `6a3c423a9`.
2. Save the snippet below as `probe.cairo`.
3. Run `cairo-compile --single-file probe.cairo`.
4. Observe the `.zip_eq()` panic instead of a diagnostic.

```bash
$ echo 'fn main() { let _x = 42_Span; }' > probe.cairo
$ cairo-compile --single-file probe.cairo
thread 'main' panicked at zip_eq_impl.rs:50:17:
itertools: .zip_eq() reached end of one iterator before the other
```

For comparison, the Cat-C path emits a clean diagnostic:

```cairo
fn main() { let _x = 42_Array; }
```

```
error[E2008]: A numeric literal of type core::array::Array cannot be created.
 --> probe.cairo:X:Y
    let _x = 42_Array;
             ^^^^^^^^
```

**Related code:**

All four Cat-A suffixes ICE the compiler:

```cairo
fn main() {
    let _a = 42_Span;         // ICE
    let _b = 42_Option;       // ICE
    let _c = 42_Result;       // ICE
    let _d = 42_PanicResult;  // ICE
}
```

The relevant corelib type definitions (each requires generic args):

```
pub enum Option<T> { Some(T), None }
pub enum Result<T, E> { Ok(T), Err(E) }
pub enum PanicResult<T> { Ok(T), Err: (Panic, Array<felt252>) }
pub struct Span<T> { snapshot: @Array<T> }
```

Contrast with the Cat-C extern types that handle the same situation gracefully:

```
pub extern type Array<T>;
pub extern type Box<T>;
pub extern type Nullable<T>;
pub extern type Felt252Dict<T>;
```

**Other information:**

- N/A

1sgtpepper:
## Summary

Reject generic core types used as numeric literal suffixes before carrying a malformed concrete type into later semantic phases.

---

## Type of change

Please check **one**:

- [ ] New feature
- [ ] Performance improvement

---

## Why is this change needed?

Numeric literal suffix syntax can name a type, but it cannot provide generic arguments. For a suffix like `42_Span`, semantic analysis resolved `Span` with an empty generic argument list. That malformed concrete type could later reach substitution/formatting and panic in `zip_eq`.

This should be reported as a normal invalid numeric literal type diagnostic instead of an internal compiler error.

---

Compiling this code could panic:

```cairo
fn main() {
    let _x = 42_Span;
}
```

---

The compiler reports an E2008-style diagnostic and does not carry the malformed generic type forward:

```text
error[E2008]: A numeric literal of type core::array::Span cannot be created.
```

---

Fixes #9975.

---

## Additional context

Validation performed locally:

- reproduced the original `42_Span` panic
- verified `42_Span` and `42_Option` now emit E2008 instead of panicking
- `cargo test -p cairo-lang-semantic expr_diagnostics::literal --profile dev`
- `scripts/rust_fmt.sh`
- clippy equivalent for `cairo-lang-semantic` with the repo's pinned nightly and warning flags

### DIFF
--- a/crates/cairo-lang-semantic/src/corelib.rs
+++ b/crates/cairo-lang-semantic/src/corelib.rs
@@ -813,6 +813,7 @@ pub fn numeric_upcastable_to_felt252(db: &dyn Database, ty: TypeId<'_>) -> bool 
 #[derive(Clone, Debug, Eq, Hash, PartialEq, salsa::Update)]
 pub enum LiteralError<'db> {
     InvalidTypeForLiteral(TypeId<'db>),
+    InvalidGenericTypeForLiteral(GenericTypeId<'db>),
     OutOfRange(TypeId<'db>),
 }
 impl<'db> LiteralError<'db> {
@@ -822,6 +823,9 @@ impl<'db> LiteralError<'db> {
                 format!("The value does not fit within the range of type {}.", ty.format(db))
             }
             Self::InvalidTypeForLiteral(ty) => {
+                format!("A numeric literal of type {} cannot be created.", ty.format(db))
+            }
+            Self::InvalidGenericTypeForLiteral(ty) => {
                 format!("A numeric literal of type {} cannot be created.", ty.format(db))
             }
         }
--- a/crates/cairo-lang-semantic/src/expr/compute.rs
+++ b/crates/cairo-lang-semantic/src/expr/compute.rs
@@ -55,7 +55,7 @@ use super::pattern::{
     PatternOtherwise, PatternTuple, PatternVariable, PatternWrappingInfo,
 };
 use crate::corelib::{
-    self, CorelibSemantic, core_binary_operator, core_bool_ty, core_unary_operator,
+    self, CorelibSemantic, LiteralError, core_binary_operator, core_bool_ty, core_unary_operator,
     false_literal_expr, get_usize_ty, never_ty, true_literal_expr, try_extract_box_inner_type,
     try_get_core_ty_by_name, unit_expr, unit_ty, unwrap_error_propagation_type, validate_literal,
 };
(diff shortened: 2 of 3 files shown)


1sgtpepper:
Closing since #9978 fixed the same issue with a smaller patch. Thanks for taking care of it.

Issue #9993, opened 2026-05-27
bug: Macro-call missing repetition operator crashes parser


researchzero-sec:
# Bug Report

**Cairo version:**

commit `6a3c423a9`

**Current behavior:**

A malformed macro call `foo!($(x))` — a `$(...)` token-tree repetition that is missing the required `?`, `+`, or `*` operator after the closing `)` — crashes `cairo-compile` with an internal compiler error before name resolution. The parser hits an `unreachable!()` branch at `crates/cairo-lang-parser/src/parser.rs:1801`.

Observed exit code: `101`.

Observed panic:

```text
thread 'main' (...) panicked at crates/cairo-lang-parser/src/parser.rs:1801:34:
internal error: entered unreachable code
```

**Expected behavior:**

The compiler should emit a clean diagnostic — the same `error[E1031]: Missing macro repetition operator. Expected '?', '+', or '*' after '$(...)'.` that the macro-definition parser already produces for the analogous shape — instead of panicking. There is a parser-context asymmetry: the macro-definition RHS handles this shape with `MissingMacroRepetitionOperator`, while macro-call token trees fall through to `unreachable!()`.

**Steps to reproduce:**

Compile the following single file with `cairo-compile`:

```cairo
fn main() -> u32 {
    foo!($(x))
}
```

Example invocation:

```bash
cairo-compile --single-file probe.cairo /dev/null
```

The process exits with code `101` and the panic shown above.

For contrast, the same missing-operator shape inside a macro definition does *not* crash and instead emits `E1031`:

```cairo
macro foo {
    ($x:expr) => { $($x) };
}

fn main() -> u32 {
    foo!(5)
}
```

**Related code:**

The macro-call token-tree parser has no recovery for the missing-operator case:

```rust
// crates/cairo-lang-parser/src/parser.rs:1795-1801
let operator = match self.peek().kind {
    SyntaxKind::TerminalQuestionMark => self.take::<TerminalQuestionMark<'_>>().into(),
    SyntaxKind::TerminalPlus => self.take::<TerminalPlus<'_>>().into(),
    SyntaxKind::TerminalMul => self.take::<TerminalMul<'_>>().into(),
    _ => unreachable!(),
};
```

The macro-definition parser handles the same shape safely:

```rust
// crates/cairo-lang-parser/src/parser.rs:792-800
let operator = match self.peek().kind {
    SyntaxKind::TerminalQuestionMark => self.take::<TerminalQuestionMark<'_>>().into(),
[...]

PR #10004, opened 2026-05-29
fix(parser): don't panic on $(...) without operator in macro calls (#9993)

orizi:
## Summary

In the parser, a missing repetition operator inside a macro call site (`$(...)` without `?`, `+`, or `*`) previously hit an `unreachable!()` panic. The operator arm now produces a `MacroRepetition​Operator::missing` node instead, deferring validation to the semantic layer. Macro *definitions* continue to emit a hard parse error (`E1031`) when the operator is absent.

---

## Type of change

Please check **one**:

- [ ] New feature
- [ ] Performance improvement

---

## Why is this change needed?

At a macro call site, `$(x)` without a trailing operator is syntactically ambiguous but not necessarily illegal — the semantic layer is responsible for checking whether the shape matches the macro declaration. Hitting `unreachable!()` in the parser caused a panic instead of a recoverable missing-node, preventing any downstream diagnostic or recovery.

---

Parsing a macro invocation containing `$(...)` with no repetition operator caused the parser to panic via `unreachable!()`.

---

The parser emits a `MacroRepetitionOperator::missing` node and continues without a diagnostic. The semantic layer is then responsible for deciding whether the missing operator is valid given the macro declaration. A new test case (`expect_diagnostics: false`) confirms that `foo!($(x))` parses cleanly. Macro definitions still produce `E1031` when the operator is missing, as validated by an additional test case (`expect_diagnostics: true`).

---

---

## Additional context

The distinction between call-site leniency and definition-site strictness is now documented with an inline comment in `parser.rs`.

### DIFF
--- a/crates/cairo-lang-parser/src/parser.rs
+++ b/crates/cairo-lang-parser/src/parser.rs
@@ -1792,13 +1792,16 @@ impl<'a, 'mt> Parser<'a, 'mt> {
                             SyntaxKind::TerminalComma => self.take::<TerminalComma<'_>>().into(),
                             _ => OptionTerminalCommaEmpty::new_green(self.db).into(),
                         };
+                        // Unlike macro definitions, a missing operator at a call site is not a
+                        // parse error: the semantic layer validates the call against the macro
+                        // declaration and decides whether the shape is legal.
                         let operator = match self.peek().kind {
                             SyntaxKind::TerminalQuestionMark => {
                                 self.take::<TerminalQuestionMark<'_>>().into()
                             }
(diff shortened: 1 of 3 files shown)


orizi:
* **#10004** <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/10004?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/10004?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #10011, opened 2026-05-31
fix(starknet): respect #[cfg] when generating deploy_for_test

orizi:
## Summary

When a `#[constructor]` function is annotated with `#[cfg(feature: 'xyz')]` and that feature is inactive, the constructor was still being picked up by `generate_constructor_deploy_function` and used to generate the `deploy_for_test` function. This happened because the body was iterated with `iter_items` instead of `iter_items_in_cfg`, ignoring cfg conditions entirely.

The fix passes the active `CfgSet` down to `generate_constructor_deploy_function` and `handle_contract_impl`, replacing `iter_items` with `iter_items_in_cfg` so that constructors (and impl items) behind inactive `#[cfg]` attributes are correctly excluded.

A new test case is added to verify that a constructor behind an inactive `#[cfg(feature: 'xyz')]` does not appear in the generated `__constructor` module or influence `deploy_for_test`.

---

## Type of change

Please check **one**:

- [ ] New feature
- [ ] Performance improvement

---

## Why is this change needed?

A constructor gated behind an inactive `#[cfg]` attribute was being included in the generated contract code because `iter_items` does not respect cfg conditions. This caused incorrect code generation — the inactive constructor's signature would influence the `deploy_for_test` helper, leading to a mismatch between what is actually compiled and what the test infrastructure expects.

---

`generate_constructor_deploy_function` iterated over all module body items unconditionally, so a `#[constructor]` behind an inactive `#[cfg]` would still be treated as the contract's constructor when generating test deployment code.

---

`generate_constructor_deploy_function` and `handle_contract_impl` now filter items through `iter_items_in_cfg`, so only constructors and impl items that are active under the current `CfgSet` are considered during code generation.

---

N/A

---

## Additional context

The `MacroPluginMetadata` argument to `handle_contract_impl` was narrowed to just `&CfgSet` since that is the only field used, which also makes the dependency explicit.

### DIFF
--- a/crates/cairo-lang-starknet/src/plugin/plugin_test_data/contracts/contract
+++ b/crates/cairo-lang-starknet/src/plugin/plugin_test_data/contracts/contract
@@ -460,3 +460,194 @@ impl StorageStorageBaseMutDrop<> of core::traits::Drop::<StorageStorageBaseMut>;
 impl StorageStorageBaseMutCopy<> of core::traits::Copy::<StorageStorageBaseMut>;
 
 //! > expected_diagnostics
+
+//! > ==========================================================================
+
+//! > Test that a constructor behind an inactive #[cfg] is not used for deploy_for_test.
+
+//! > test_runner_name
+ExpandContractTestRunner(expect_diagnostics: false)
+
+//! > cairo_code
+#[starknet::contract]
+mod test_contract {
+    #[storage]
+    struct Storage {}
+
+    #[cfg(feature: 'xyz')]
+    #[constructor]
+    fn constructor(ref self: ContractState, special_arg: u256) {}
+}
+
+//! > generated_cairo_code
+lib.cairo:
+
+#[starknet::contract]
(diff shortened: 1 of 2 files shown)


orizi:
* **#10011** <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/10011?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/10011?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

TomerStarkware (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/10011#-:-OtxBAfd2zyXZgzyn5YF:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on eytan-starkware).

PR #10033, opened 2026-06-04
fix(lowering): early_unsafe_panic must keep side-effecting statements

orizi:
## Summary

Fixes the `early_unsafe_panic` optimization so that the `unsafe_panic` insertion point is placed **after** the last side-effecting statement in a block, rather than at the statement itself. Previously, the loop iterated forward and pushed the fix at index `i` (overwriting the side-effecting statement), and the test harness ran against `PreOptimizations` lowering where inlined externs like `debug::print` were not yet visible. Now the loop iterates in reverse, finds the last side-effecting statement, and inserts the early panic at `i + 1` — preserving the side effect. The test stage is also updated to `PostBaseline` so that inlined side-effecting externs are present when the optimization runs.

---

## Type of change

Please check **one**:

- [ ] New feature
- [ ] Performance improvement

---

## Why is this change needed?

The previous forward-iterating loop pushed a fix at index `i` for the first side-effecting statement it found, which would replace that statement with an `unsafe_panic` call instead of inserting the panic after it. This caused side-effecting statements (e.g., `debug::print`) to be dropped from the lowered output. Additionally, the test harness used `PreOptimizations` lowering, which does not include inlined externs, so the bug was not caught by the existing test for the side-effect-preservation case.

---

When a block contained a side-effecting statement followed by dead code (no reachable `return`), the optimization replaced the side-effecting statement itself with `unsafe_panic`, discarding the side effect.

---

The optimization now inserts `unsafe_panic` immediately after the last side-effecting statement, preserving all side effects. A new test case (`Test that a side-effecting statement (debug::print) before the dead tail is preserved`) verifies this behavior using `PostBaseline` lowering.

---

N/A

---

## Additional context

The fix also simplifies the conditional logic: instead of checking `ReachableSideEffects::Reachable` to early-return and then re-checking `Unreachable` inside the loop, the new code destructures `Unreachable` upfront and returns early if the state is `Reachable`.

### DIFF
--- a/crates/cairo-lang-lowering/src/optimizations/early_unsafe_panic.rs
+++ b/crates/cairo-lang-lowering/src/optimizations/early_unsafe_panic.rs
@@ -106,14 +106,14 @@ impl<'db, 'a> DataflowAnalyzer<'db, 'a> for UnsafePanicContext<'db> {
         {
             self.fixes.push(((block_id, block.statements.len()), *match_info.location()));
         }
-        if ReachableSideEffects::Reachable == *info {
+        let ReachableSideEffects::Unreachable(location) = *info else {
             return;
-        }
-        for (i, stmt) in block.statements.iter().enumerate() {
-            if self.has_side_effects(stmt)
-                && let ReachableSideEffects::Unreachable(locations) = *info
-            {
-                self.fixes.push(((block_id, i), locations));
+        };
+        // Everything past the last side-effecting statement is dead (no `return` is reachable
+        // from here), so replace it with an early panic — but keep the side effects themselves.
+        for (i, stmt) in block.statements.iter().enumerate().rev() {
+            if self.has_side_effects(stmt) {
+                self.fixes.push(((block_id, i + 1), location));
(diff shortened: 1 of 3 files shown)


orizi:
* **#10033** <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/10033?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/10033?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

eytan-starkware (approved):
:lgtm:

> @eytan-starkware reviewed 3 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/10033#-:-OuGgEhgB5130PK6Z1Fn:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on TomerStarkware).

PR #10065, opened 2026-06-08
bugfix(plugins): Use all cfg args to extract diagnostics.

orizi:
## Summary

Introduces `parse_predicate_items`, a helper that collects parsed predicate arguments and returns `None` if any single argument fails to parse. This replaces the previous `filter_map` approach, which silently dropped malformed arguments from `and`/`or`/top-level predicate lists. The old behavior could cause incorrect item inclusion — for example, an `or` predicate containing only invalid arguments would evaluate to `false` (empty `or`) rather than being treated as an error.

---

## Type of change

Please check **one**:

- [ ] New feature
- [ ] Performance improvement

---

## Why is this change needed?

When a `#[cfg(or(...))]` or `#[cfg(and(...))]` attribute contained unparsable arguments, the previous implementation silently dropped those arguments via `filter_map`. This meant an `or` predicate with only invalid arguments would collapse to an empty `or`, evaluating to `false` and incorrectly excluding the item. The same issue applied to `and` predicates and top-level predicate lists.

---

Invalid arguments inside `or(...)`, `and(...)`, or top-level `cfg(...)` predicates were silently dropped. A diagnostic was emitted per invalid argument, but the remaining valid arguments were still evaluated, potentially producing incorrect inclusion/exclusion results.

---

If any argument in a predicate list fails to parse, the entire predicate returns `None`, preventing the item from being incorrectly included or excluded due to a partially-evaluated predicate. Diagnostics are still emitted for each invalid argument. A new test case verifies that items annotated with `#[cfg(or(0, 1))]` and `#[cfg(and(0, 1))]` are kept (not dropped) when their arguments are invalid.

---

N/A

---

## Additional context

The `parse_predicate_items` helper uses an explicit loop with `extend` rather than `filter_map`, tracking expected vs. actual parsed count to detect any parse failure and propagate `None` accordingly.

### DIFF
--- a/crates/cairo-lang-plugins/src/plugins/config.rs
+++ b/crates/cairo-lang-plugins/src/plugins/config.rs
@@ -203,12 +203,7 @@ fn parse_predicate<'a>(
     attr: Attribute<'a>,
     diagnostics: &mut Vec<PluginDiagnostic<'a>>,
 ) -> Option<PredicateTree> {
-    Some(PredicateTree::And(
-        attr.args
-            .into_iter()
-            .filter_map(|arg| parse_predicate_item(db, arg, diagnostics))
-            .collect(),
-    ))
+    Some(PredicateTree::And(parse_predicate_items(db, attr.args, diagnostics)?))
 }
 
 /// Parse single `#[cfg(...)]` attribute argument as a [`Cfg`] item.
@@ -252,11 +247,7 @@ fn parse_predicate_item<'a>(
                         ));
                         None
                     } else {
-                        Some(PredicateTree::And(
-                            args.into_iter()
-                                .filter_map(|arg| parse_predicate_item(db, arg, diagnostics))
-                                .collect(),
-                        ))
+                        Some(PredicateTree::And(parse_predicate_items(db, args, diagnostics)?))
                     }
                 }
                 "or" => {
@@ -267,11 +258,7 @@ fn parse_predicate_item<'a>(
                         ));
                         None
                     } else {
-                        Some(PredicateTree::Or(
-                            args.into_iter()
-                                .filter_map(|arg| parse_predicate_item(db, arg, diagnostics))
-                                .collect(),
-                        ))
+                        Some(PredicateTree::Or(parse_predicate_items(db, args, diagnostics)?))
                     }
                 }
                 _ => {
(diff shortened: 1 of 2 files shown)


orizi:
* **#10065** <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/10065?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/10065?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

eytan-starkware (changes requested):
> @eytan-starkware reviewed 2 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/10065)* status: all files reviewed, 1 unresolved discussion (waiting on orizi and TomerStarkware).

___
*[`crates/cairo-lang-plugins/src/test_data/config` line 499 at r1](https://reviewable.io/reviews/starkware-libs/cairo/10065#-Ouf8H317GMr256q3qm4:-Ouf8H317GMr256q3qm5:b-qkhvsj) ([raw file](https://github.com/starkware-libs/cairo/blob/9b3016ac56105571e1bc8a3d2b99d668e83860bc/crates/cairo-lang-plugins/src/test_data/config#L499)):*
> ```
> #[cfg(and(0, 1))]
>              ^
> ```

Add a nested test with an error on outer one and an error in the inner one

orizi (commented):
> @orizi made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/10065)* status: all files reviewed, 1 unresolved discussion (waiting on eytan-starkware and TomerStarkware).

___
*[`crates/cairo-lang-plugins/src/test_data/config` line 499 at r1](https://reviewable.io/reviews/starkware-libs/cairo/10065#-Ouf8H317GMr256q3qm4:-OufFbShAiukUFfWd4z7:b-896fix) ([raw file](https://github.com/starkware-libs/cairo/blob/9b3016ac56105571e1bc8a3d2b99d668e83860bc/crates/cairo-lang-plugins/src/test_data/config#L499)):*

Add a nested test with an error on outer one and an error in the inner one
</blockquote></details>

Done.

eytan-starkware (approved):
:lgtm:

> @eytan-starkware reviewed 1 file and all commit messages, made 1 comment, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/10065#-:-OufJBudCnj5oTu2ee1w:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on TomerStarkware).

PR #10069, opened 2026-06-08
fix(starknet): handle empty enum in `Store` derive

orizi:
## Summary

Fixes malformed Cairo code generated by the `#[derive(starknet::Store)]` macro for enums. Match arm strings were being joined with `,\n` at the join site while also having commas appended individually to each arm, resulting in double commas. The fix moves the trailing comma into each arm's format string and joins with `\n` only. Additionally, adds support for empty enums (`enum Foo {}`): the `size()` method previously emitted a dangling `+` operator when no variants were present; it now falls back to `0` in that case, matching the struct handler's behavior.

---

## Type of change

Please check **one**:

- [ ] New feature
- [ ] Performance improvement

---

## Why is this change needed?

The `Store` derive macro for enums was joining match arms with `,\n` at the `join` call while each arm string already lacked a trailing comma, causing the last arm to be missing its comma. Separately, when an enum had zero variants, `match_size` remained empty and was interpolated directly into `1_u8 + ` in `size()`, producing invalid Cairo syntax with a dangling `+`.

---

- Generated `write` and `write_at_offset` match arms were missing trailing commas on the last arm.
- Deriving `starknet::Store` on an empty enum caused a compiler error due to `1_u8 + ` with nothing following the operator in the generated `size()` implementation.

---

- Each match arm string now carries its own trailing comma, and arms are joined with `\n` alone, producing well-formed Cairo match expressions.
- Empty enums derive `starknet::Store` without a syntax error; `size()` emits `1_u8 + 0` instead of `1_u8 + `.

---

N/A

---

## Additional context

Test data snapshots for `new_storage_interface` and `user_defined_types` have been updated to reflect the corrected output, including a new `EmptyEnum` test case and its expected generated `Store` implementation and diagnostics.

### DIFF
--- a/crates/cairo-lang-starknet/src/plugin/derive/store.rs
+++ b/crates/cairo-lang-starknet/src/plugin/derive/store.rs
@@ -217,6 +217,41 @@ fn handle_enum<'db>(
 ) -> Option<RewriteNode<'db>> {
     let enum_name = &info.name;
     let full_name = &info.full_typename();
+
+    // A never-kind enum (no variants) is uninhabited: a value can never be constructed, so `write`
+    // is unreachable and `read` always fails. Emit a trivial impl instead of the per-variant match.
+    if info.members_info.is_empty() {
+        let store_impl = formatdoc!(
+            "
+            {header} {{
+                fn read(address_domain: u32, base: starknet::storage_access::StorageBaseAddress) \
+             -> starknet::SyscallResult<{full_name}> {{
+                    starknet::SyscallResult::Err(array!['Read of an uninhabited type'])
+                }}
+                fn write(address_domain: u32, base: starknet::storage_access::StorageBaseAddress, \
+             value: {full_name}) -> starknet::SyscallResult<()> {{
+                    match value {{}}
+                }}
+                fn read_at_offset(address_domain: u32, base: \
+             starknet::storage_access::StorageBaseAddress, offset: u8) -> \
+             starknet::SyscallResult<{full_name}> {{
+                    starknet::SyscallResult::Err(array!['Read of an uninhabited type'])
+                }}
+                fn write_at_offset(address_domain: u32, base: \
+             starknet::storage_access::StorageBaseAddress, offset: u8, value: {full_name}) -> \
+             starknet::SyscallResult<()> {{
+                    match value {{}}
+                }}
+                fn size() -> u8 {{
+                    0
+                }}
+            }}
+            ",
+            header = info.impl_header(STORE_TRAIT, &[STORE_TRAIT, "core::traits::Destruct"]),
+        );
+        return Some(RewriteNode::Text(store_impl));
+    }
+
     let mut match_idx = Vec::new();
     let mut match_idx_at_offset = Vec::new();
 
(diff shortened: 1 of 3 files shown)


orizi:
* **#10069** <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/10069?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/10069?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

orizi (commented):
> @orizi resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/10069)* status: 0 of 3 files reviewed, all discussions resolved (waiting on eytan-starkware and TomerStarkware).

eytan-starkware:
The empty-enum handling can be a lot smaller. For a variant-less enum the existing template only breaks in two spots, and the comma fix already covers one of them:

1. `size()` emits a dangling `1_u8 + ` — the exact case the struct handler already guards one function up (`if sizes.is_empty() { "0" }`). A 3-line fallback after the loop is enough:
   ```rust
   if match_size.is_empty() {
       match_size = "0".to_string();
   }
   ```
   → `1_u8 + 0`.
2. `match idx { , 0 | _ => … }` leading comma — already handled by moving the trailing comma into each arm (the comma fix in this PR).

With just those, the existing template generates valid code for empty enums:
```cairo
fn read(...)  { let idx = Store::<felt252>::read(...)?; match idx { 0 | _ => Err(...) } }
fn write(..., value) { match value {}; Ok(()) }   // uninhabited → exhaustive
fn size() -> u8 { 1_u8 + 0 }
```
so the separate ~35-line impl isn't needed. Net is +16/−10 in `store.rs` instead of +45/−10, verified compiling via the golden test (`EmptyEnum` gets only the usual "no default variant" warning, no error).

Minor: this reuses the normal `read` (one extra felt read before erroring) and `size` is `1` not `0` — both moot for an uninhabited type, and `1` is arguably more consistent since every enum reserves the discriminant byte.

If we'd rather not pretend an uninhabited type is storable, the even-smaller option is to reject it outright with a diagnostic at the top of `handle_enum` and `return None`.

orizi (commented):
this generates unreachable code - handling to create trivial code is more sensible.

> @orizi made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/10069#-:-OufyMX47rBblogByqGG:btwb5f7)* status: 0 of 3 files reviewed, all discussions resolved (waiting on eytan-starkware and TomerStarkware).

eytan-starkware (approved):
:lgtm:

> @eytan-starkware reviewed 3 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/10069#-:-Oug-UxU0rJGD3H1xQOE:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on TomerStarkware).

PR #10103, opened 2026-06-15
fix(parser): anchor literal-validation diagnostics at the token text.

orizi:
## Summary

Fixes incorrect diagnostic span anchoring for string/short-string/number literal validation errors when the literal appears with leading trivia (e.g., indented on its own line). Previously, diagnostic spans were computed relative to `self.offset`, which includes the terminal's leading trivia width, causing the error underline to point at the wrong position in the source. Now, the span is anchored at the start of the terminal's text (after leading trivia), so the caret and underline correctly point to the literal itself.

The three `take_terminal_*` methods are also consolidated into a single generic `take_validated_terminal` helper that accepts a validation function, eliminating the repeated peek/take/diagnostic pattern.

---

## Type of change

Please check **one**:

- [ ] New feature
- [ ] Performance improvement

---

## Why is this change needed?

When a string or short-string literal appeared as the first token on a new line (i.e., with leading whitespace trivia), validation diagnostics such as "Invalid string escaping" or "Short string literals can only include ASCII characters" would point to the wrong column — offset by the width of the leading whitespace — rather than pointing directly at the literal.

---

Diagnostic spans for literal validation errors were computed using `self.offset` and `self.current_width`, which included the terminal's leading trivia. This caused the error underline to be shifted left, pointing into the indentation whitespace rather than the literal text.

---

Diagnostic spans are now anchored at `self.offset + leading_trivia_width`, so the span starts at the first character of the literal text. The underline and caret in error messages correctly point to the literal, even when it is indented or preceded by other trivia.

New test cases cover `'\p'` and `'\u{1024}'` appearing with leading trivia (literal first on its line) to confirm correct span reporting.

---

---

## Additional context

The `trivia_total_width` helper is used to compute the leading trivia width from the peeked token before it is consumed by `self.take::<Terminal>()`.

### DIFF
--- a/crates/cairo-lang-parser/src/parser.rs
+++ b/crates/cairo-lang-parser/src/parser.rs
@@ -3144,44 +3144,45 @@ impl<'a, 'mt> Parser<'a, 'mt> {
 
     /// Takes and validates a TerminalLiteralNumber token.
     fn take_terminal_literal_number(&mut self) -> TerminalLiteralNumberGreen<'a> {
-        let diag = validate_literal_number(self.peek().text.long(self.db));
-        let green = self.take::<TerminalLiteralNumber<'_>>();
-        self.add_optional_diagnostic(diag);
-        green
+        self.take_validated_terminal::<TerminalLiteralNumber<'_>>(validate_literal_number)
     }
 
     /// Takes and validates a TerminalShortString token.
     fn take_terminal_short_string(&mut self) -> TerminalShortStringGreen<'a> {
-        let diag = validate_short_string(self.peek().text.long(self.db));
-        let green = self.take::<TerminalShortString<'_>>();
-        self.add_optional_diagnostic(diag);
-        green
+        self.take_validated_terminal::<TerminalShortString<'_>>(validate_short_string)
     }
 
     /// Takes and validates a TerminalString token.
     fn take_terminal_string(&mut self) -> TerminalStringGreen<'a> {
-        let diag = validate_string(self.peek().text.long(self.db));
-        let green = self.take::<TerminalString<'_>>();
-        self.add_optional_diagnostic(diag);
-        green
+        self.take_validated_terminal::<TerminalString<'_>>(validate_string)
     }
 
-    /// Adds a diagnostic of kind `kind` if provided, at the current offset.
-    fn add_optional_diagnostic(&mut self, err: Option<ValidationError>) {
-        if let Some(err) = err {
+    /// Takes a terminal of the given kind and validates its text with `validate`, reporting the
+    /// resulting diagnostic (if any). Validation offsets are relative to the terminal's text, so
+    /// the diagnostic span is anchored at the text and excludes the terminal's trivia.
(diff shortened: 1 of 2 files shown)


orizi:
* **#10103** <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/10103?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/10103?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

eytan-starkware (changes requested):
> @eytan-starkware reviewed 1 file and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/10103)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on orizi and TomerStarkware).

___
*[`crates/cairo-lang-parser/src/parser.rs` line 3170 at r1](https://reviewable.io/reviews/starkware-libs/cairo/10103#-OvAFHQKBnqeR1HlpxhN:-OvAFHQL5kXl4rLhG_8d:biz5lk0) ([raw file](https://github.com/starkware-libs/cairo/blob/4068f9060f8fcb212736961994944dc7091e154c/crates/cairo-lang-parser/src/parser.rs#L3170)):*
> ```XML
>         let text = peek.text.long(self.db);
>         let leading_width = trivia_total_width(self.db, &peek.leading_trivia);
>         let green = self.take::<Terminal>();
> ```

Cant we get text from green? Also we only need the leading width inside the error case so we should do that inside the if

orizi (commented):
> @orizi made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/10103)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on eytan-starkware and TomerStarkware).

___
*[`crates/cairo-lang-parser/src/parser.rs` line 3170 at r1](https://reviewable.io/reviews/starkware-libs/cairo/10103#-OvAFHQKBnqeR1HlpxhN:-OvAtqvZ53Nvs4xgGfAs:b-gd828w) ([raw file](https://github.com/starkware-libs/cairo/blob/4068f9060f8fcb212736961994944dc7091e154c/crates/cairo-lang-parser/src/parser.rs#L3170)):*

Cant we get text from green? Also we only need the leading width inside the error case so we should do that inside the if
</blockquote></details>

unfortunately - it seems we cannot. as we know nothing about the type of the terminal at this point, since it is generic.

eytan-starkware (approved):
:lgtm:

> @eytan-starkware reviewed 1 file, made 1 comment, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/10103#-:-OvEnkQb6VVuEMqDmXqP:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on TomerStarkware).

PR #10137, opened 2026-06-21
bug fix(starknet): Fully-qualify prelude types in contract codegen.

### LINKED ISSUE #10128: bug: Contract codegen bare prelude type names break compile
# Bug Report

**Cairo version:** 6a3c423a9

**Current behavior:**

The `#[starknet::contract]` plugin emits generated contract code that references prelude types by their **bare** names (`Array<felt252>`, `Option<...>`, `Span<felt252>`), relying on the prelude rather than emitting fully-qualified paths (`core::array::Array`, `core::option::Option`, `core::array::Span`). When a user defines a type with one of those names inside the contract module, Cairo's single flat namespace makes the user type shadow the prelude name, so the generated reference resolves to the user's struct instead. A user `struct Array {}` (0 generic params) where the generated code expects `Array<felt252>` produces:

```
error[E2163]: Expected 0 generic arguments, found 1
```

This happens even for a storage-only contract with no entry points — the contract-state/storage scaffolding alone references the bare prelude types. The diagnostic gives no hint that the user's type name is the cause.

**Expected behavior:**

The contract should compile. The generated codegen should fully-qualify every prelude type it references (`core::array::Array`, `core::option::Option`, `core::array::Span`) so that a same-module user type sharing the name cannot shadow it.

**Steps to reproduce:**

Compile the snippet below with the Starknet plugin enabled, e.g.:

```
cairo-test --single-file <file>.cairo --starknet
```

This yields `error[E2163]: Expected 0 generic arguments, found 1`. Replacing `Array` with `Option` or `Span` produces the same error.

Controls (both compile):
- The same `struct Array {}` declared at top level, outside any contract, compiles — shadowing a prelude type is allowed generally; these names are not reserved keywords.
- A non-prelude name such as `struct MyThing {}` inside the contract compiles.

**Related code:**

```
#[starknet::contract]
mod c {
    #[storage]
    struct Storage {}
    struct Array {}
}
```

**Other information:**

This is a contract-codegen hygiene gap on the prelude-**type** axis: the fix is to fully-qualify every prelude type the codegen emits. It is distinct from parameter-**name** collisions where generated bindings are shadowed (e.g. the dispatcher `__calldata__`, embeddable `component`, Serde `serialized`, and Hash `value` cases).

orizi:
## Summary

The `#[starknet::contract]` plugin now fully-qualifies the prelude types it
emits in generated code (`core::array::Array`, `core::array::Span`,
`core::option::Option`, plus `Option::Some`/`Option::None` and
`core::array::ArrayTrait::new`) instead of relying on the bare prelude names.

Touched codegen sites:
- `derive/event.rs` — struct & enum `Event` impls (`append_keys_and_data` /
  `deserialize`) and the flat/nested variant deserialization.
- `entry_point.rs` — entry-point wrapper signature (`Span`) and `ArrayTrait::new`.
- `dispatcher.rs` — forward-method wrapper signature (`Span`).

Adds a regression test (`shadowed_prelude_types`) covering a contract that
defines `Array`, `Span`, and `Option` user types, and regenerates the affected
`plugin_test_data` golden files.

---

## Type of change

---

## Why is this change needed?

Cairo has a single flat namespace, so a user type defined inside a contract
module that shares a name with a prelude type shadows that prelude name. Because
the generated code referenced the bare names (e.g. `Array<felt252>`), a user
`struct Array {}` (0 generic params) made the generated `Array<felt252>` resolve
to the user struct, producing:

```
error[E2163]: Expected 0 generic arguments, found 1
```

This happened even for a storage-only contract with no entry points — the
auto-derived empty `Event` impl alone references the bare prelude types. The
diagnostic gave no hint that the user's type name was the cause.

Fixes #10128.

---

## What was the behavior before?

```cairo
#[starknet::contract]
mod c {
    #[storage]
    struct Storage {}
    struct Array {}
}
```

failed to compile with `error[E2163]: Expected 0 generic arguments, found 1`
(same for `Option` and `Span`).

---

## What is the behavior after?

The contract compiles. Generated code fully-qualifies every prelude type it
references, so a same-module user type sharing the name can no longer shadow it.

---

## Additional context

Removing the bare reference also surfaced a now-redundant
`use core::array::ArrayTrait;` in `cairo_level_tests/l2_to_l1_messages.cairo`
(it was only being consumed by the previously-bare generated `ArrayTrait::new()`),
which has been dropped.

This is scoped to the prelude-**type** axis named in the issue and is distinct
from parameter-**name** collisions (e.g. `__calldata__`, `component`,
`serialized`, `value`).

### DIFF
--- a/crates/cairo-lang-starknet/cairo_level_tests/l2_to_l1_messages.cairo
+++ b/crates/cairo-lang-starknet/cairo_level_tests/l2_to_l1_messages.cairo
@@ -3,7 +3,6 @@ use starknet::{SyscallResultTrait, testing};
 
 #[starknet::contract]
 mod contract_with_messages_sent_to_l1 {
-    use core::array::ArrayTrait;
     use starknet::SyscallResultTrait;
     use starknet::storage::{StoragePointerReadAccess, StoragePointerWriteAccess};
     use super::generate_payload;
--- a/crates/cairo-lang-starknet/src/plugin/derive/event.rs
+++ b/crates/cairo-lang-starknet/src/plugin/derive/event.rs
@@ -75,13 +75,14 @@ fn handle_struct<'db>(
             "
             impl $struct_name$IsEvent of {EVENT_TRAIT}<$struct_name$> {{
                 fn append_keys_and_data(
-                    self: @$struct_name$, ref keys: Array<felt252>, ref data: Array<felt252>
+                    self: @$struct_name$, ref keys: core::array::Array<felt252>, ref data: \
+             core::array::Array<felt252>
                 ) {{$append_members$
                 }}
                 fn deserialize(
-                    ref keys: Span<felt252>, ref data: Span<felt252>,
-                ) -> Option<$struct_name$> {{
-                    Option::Some($struct_name$ {{$ctor$}})
+                    ref keys: core::array::Span<felt252>, ref data: core::array::Span<felt252>,
+                ) -> core::option::Option<$struct_name$> {{
+                    core::option::Option::Some($struct_name$ {{$ctor$}})
                 }}
             }}
             "
@@ -233,10 +234,10 @@ fn handle_enum<'db>(
             let mut keys = keys;
             let mut data = data;
             match {} {{
-                Option::Some(val) => {{
-                    return Option::Some($enum_name$::$variant_name$(val));
+                core::option::Option::Some(val) => {{
+                    return core::option::Option::Some($enum_name$::$variant_name$(val));
                 }},
-                Option::None => {{}},
+                core::option::Option::None => {{}},
             }};
         }}
         ",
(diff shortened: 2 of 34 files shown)


orizi:
* **#10137** <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/10137?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/10137?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

TomerStarkware (approved):
:lgtm:

> @TomerStarkware reviewed 35 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/10137#-:-OveLYzL9WbvvbiNERRR:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on eytan-starkware).

PR #10149, opened 2026-06-22
bugfix(parser): Make `macro` a valid recovery terminal and tree token.

### LINKED ISSUE #10144: bug: Block comment in macro args panics compiler
# Bug Report

**Cairo version:**

commit `6a3c423a9`

**Current behavior:**

Placing a `/* ... */` block comment inside an inline macro argument list (e.g. inside `assert!(...)`) crashes `cairo-compile` with an internal compiler error (exit code 101). The panic message is:

```
Unexpected token kind: TerminalMacro
```

originating from `crates/cairo-lang-parser/src/parser.rs:1965`.

A `//` line comment in the same position compiles cleanly, so a formatting-only change from a line comment to a block comment inside an inline macro argument can turn a clean compile into a compiler panic.

**Expected behavior:**

The compiler should either accept the block comment (treating it equivalently to a line comment in this context) or, at minimum, emit a structured diagnostic (such as the previously-observed `E2200` plugin diagnostic) rather than panicking. Ordinary user syntax should never escalate to an ICE.

**Steps to reproduce:**

1. Save the snippet below as `repro.cairo`.
2. Run `cairo-compile --single-file repro.cairo`.
3. Observe that the process exits with code 101 and prints `Unexpected token kind: TerminalMacro`.
4. As a control, replace the `/* ... */` block comment with a `//` line comment on its own line; the file then compiles cleanly under the same binary.

**Related code:**

Reproducer:

```cairo
fn main() {
    let x = 1_u32;
    assert!(
        /* block comment inside inline macro arg list */
        x == 1_u32
    );
}
```

Relevant compiler source:

```rust
// crates/cairo-lang-parser/src/macro_helpers.rs:21-42
let node_text = token_tree.as_syntax_node().get_text(db);
let mut parser = Parser::new(db, file_id, node_text, &mut diagnostics);
let wrapped_arg_list_green: WrappedArgListGreen<'a> = parser.parse_wrapped_arg_list();
...
if !diagnostics.is_empty() {
    return None;
}
```

```rust
// crates/cairo-lang-parser/src/parser.rs:1965
other => unreachable!("Unexpected token kind: {other:?}"),
```

`token_tree_as_wrapped_arg_list` reparses the macro token tree text. With a block comment inside an inline macro argument list, the reparse reaches the `unreachable!` arm at `parser.rs:1965` and panics on `TerminalMacro` instead of producing a diagnostic.

**Other information:**

- The comment-kind boundary is sharp: `//` line comments inside `assert!(...)` compile cleanly, while `/* ... */` block comments in the same position ICE. Operationally this is a diagnostic-to-ICE escalation on ordinary user syntax.
- Adjacent surfaces worth auditing for the same panic path: doc comments (`///`, `//!`) inside inline macro arguments, and block comments inside attribute argument lists.

orizi:
## Summary

Fixes an ICE (Internal Compiler Error) that occurred when the `macro` keyword was used as a token inside an inline macro call (e.g., `foo!(macro)`). The `TerminalMacro` syntax kind was missing from the parser's token-tree leaf handling, causing a panic. Additionally, `TerminalMacro` was not included in the error recovery set of module-level keywords, meaning a `macro` declaration following a malformed item would not correctly stop error recovery.

---

## Type of change

Please check **one**:

- [ ] New feature
- [ ] Performance improvement

---

## Why is this change needed?

`TerminalMacro` was absent from the parser's token-tree leaf dispatch in `parser.rs`, so encountering `macro` as a leaf token inside an inline macro invocation caused an ICE. It was also missing from the `module_item_kw!` recovery macro in `recovery.rs`, so error recovery would not stop at a `macro` declaration boundary when recovering from a prior parse error.

---

- Using `macro` as an argument to an inline macro call (e.g., `foo!(macro)`) caused an internal compiler error (panic).
- A `macro` declaration following a malformed item (e.g., `const X` with no type or value) would not act as a recovery boundary, potentially producing confusing cascading diagnostics.

---

- `foo!(macro)` is parsed correctly without panicking; `macro` is treated as a `TokenTreeLeaf` with kind `TokenMacro`.
- Error recovery now stops at a `macro` keyword at module level, consistent with other module-item keywords like `struct`, `fn`, `impl`, etc.

---

Regression fix for #10144.

---

## Additional context

Two new test cases are added: one verifying the inline macro regression no longer ICEs, and one verifying that error recovery halts correctly at a following `macro` declaration.

### DIFF
--- a/crates/cairo-lang-parser/src/parser.rs
+++ b/crates/cairo-lang-parser/src/parser.rs
@@ -1872,6 +1872,7 @@ impl<'a, 'mt> Parser<'a, 'mt> {
             SyntaxKind::TerminalImpl => self.take::<TerminalImpl<'_>>().into(),
             SyntaxKind::TerminalImplicits => self.take::<TerminalImplicits<'_>>().into(),
             SyntaxKind::TerminalLet => self.take::<TerminalLet<'_>>().into(),
+            SyntaxKind::TerminalMacro => self.take::<TerminalMacro<'_>>().into(),
             SyntaxKind::TerminalMatch => self.take::<TerminalMatch<'_>>().into(),
             SyntaxKind::TerminalModule => self.take::<TerminalModule<'_>>().into(),
             SyntaxKind::TerminalMut => self.take::<TerminalMut<'_>>().into(),
--- a/crates/cairo-lang-parser/src/parser_test_data/diagnostics/module_diagnostics
+++ b/crates/cairo-lang-parser/src/parser_test_data/diagnostics/module_diagnostics
@@ -66,3 +66,37 @@ error[E1029]: Expected either ';' or '{' after module name. Use ';' for an exter
  --> dummy_file.cairo:1:11
 mod my_mod const X: felt252 = 5;
           ^
+
+//! > ==========================================================================
+
+//! > Error recovery stops at a following `macro` declaration.
+
+//! > test_runner_name
+get_diagnostics
+
+//! > cairo_code
+const X
+macro m {
+    ($x: expr) => { $x };
+}
+
+//! > expected_diagnostics
+error[E1004]: Unexpected token, expected ':' followed by a type.
+ --> dummy_file.cairo:1:8
+const X
+       ^
+
+error[E1001]: Missing token '='.
+ --> dummy_file.cairo:1:8
+const X
+       ^
+
+error[E1002]: Missing tokens. Expected an expression.
+ --> dummy_file.cairo:1:8
+const X
+       ^
+
+error[E1001]: Missing token ';'.
+ --> dummy_file.cairo:1:8
+const X
+       ^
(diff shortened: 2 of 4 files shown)


orizi:
* **#10149** <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/10149?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/10149?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

TomerStarkware (approved):
:lgtm:

> @TomerStarkware reviewed 4 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/10149#-:-OvnV3bm4pseIj_sXT4U:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on eytan-starkware).

PR #10151, opened 2026-06-22
bugfix(doc): Made inline links with markup point to the url.

orizi:
## Summary

Fixes incorrect destination span computation for inline Markdown links whose labels contain markup (e.g. `[**bold**](path)`). The `](` boundary separating the label from the destination was assumed to be immediately adjacent to the label's end offset, but when the label contains emphasis or other inline markup, the parser-reported label end does not point directly at `](`. The fix searches forward in the raw content string from the label's end to locate the actual `](` before computing the destination range. The `content` string is now threaded through `location_from_link_fields` and `find_inline_destination_range` to enable this lookup.

---

## Type of change

Please check **one**:

- [ ] New feature
- [ ] Performance improvement

---

## Why is this change needed?

When a doc comment link label contained bold or italic markers such as `[**bold**](path)`, the computed `dest_span` pointed to the wrong byte range in the source text. This caused incorrect go-to-definition or hover behavior for those links because the destination text offset was shifted by the length of the markup characters inside the label.

---

`find_inline_destination_range` assumed the destination started exactly 2 bytes after the label's end offset (i.e., immediately after `](`). For plain labels this is correct, but for labels with inline markup the Markdown parser reports the label end at the closing `**` or `*`, not at `]`, so the computed destination range was wrong.

---

`find_inline_destination_range` now scans forward in the raw comment string from the label's end to find the `](` token, then computes the destination range relative to that position. Links like `[**bold**](path)` and `[*em*](path2)` now produce correct `dest_span` values, as verified by the new test cases.

---

---

## Additional context

A new test case covering `[**bold**](path)` and `[*em*](path2)` in the same doc comment is added to `documentation_comment_parser_links.txt` to prevent regressions.

### DIFF
--- a/crates/cairo-lang-doc/src/parser.rs
+++ b/crates/cairo-lang-doc/src/parser.rs
@@ -264,6 +264,7 @@ pub fn parse_documentation_comment(documentation_comment: &str) -> Vec<Documenta
                             .as_ref()
                             .and_then(|label_range| {
                                 location_from_link_fields(
+                                    documentation_comment,
                                     link.link_type,
                                     &link.destination,
                                     label_range,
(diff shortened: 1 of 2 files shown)


orizi:
* **#10151** <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/10151?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/10151?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

TomerStarkware (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/10151#-:-OvnOJ-HDfHMHUSdwloL:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on eytan-starkware).

PR #10156, opened 2026-06-24
feat(starknet): generate a CallBuilder for #[starknet::interface] traits

Yoni-Starkware:
## What

Adds a **`CallBuilder`** to the code generated for every `#[starknet::interface]` trait, sitting alongside the existing `Dispatcher` / `LibraryDispatcher` / `SafeDispatcher` family.

A dispatcher method does: serialize args → `selector!` → **syscall** → deserialize return. A call builder is the same method with the syscall/deserialize removed, returning a `starknet::account::Call` instead:

```cairo
#[feature("call_builder")]
fn build_transfer_call(token: ContractAddress, to: ContractAddress, amount: u256) -> Call {
    IERC20CallBuilder { contract_address: token }.transfer(to, amount)
}
```

This removes the hand-rolled boilerplate of building `Call { to, selector, calldata }` (manual `selector!` + manual `Serde`/`u256` splitting) for multicall and account `__execute__` flows.

## Why

Building a `Call` for an interface method today means manually replicating exactly what the dispatcher already generates internally — the selector and the argument serialization — which is verbose and error-prone (e.g. splitting `u256` into `low`/`high` by hand). The "build phase" is literally the first half of an existing dispatcher method, so this reuses the dispatcher's own serialization and `selector!` generation rather than introducing a parallel encoding path.

## What's generated

For `trait IFoo<T> { fn transfer(ref self: T, to: ContractAddress, amount: u256); ... }`:

```cairo
#[doc(group: "dispatchers")]
trait IFooCallBuilderTrait<T> {
    #[unstable(feature: "call_builder")]
    fn transfer(self: T, to: ContractAddress, amount: u256) -> starknet::account::Call;
    // ...one method per interface fn, return type forced to Call
}

#[doc(group: "dispatchers")]
#[derive(Copy, Drop, starknet::Store, Serde)]
struct IFooCallBuilder {
    pub contract_address: starknet::ContractAddress,
}

#[doc(group: "dispatchers")]
impl IFooCallBuilderImpl of IFooCallBuilderTrait<IFooCallBuilder> {
    fn transfer(self: IFooCallBuilder, to: ContractAddress, amount: u256) -> starknet::account::Call {
        let mut __calldata__ = core::traits::Default::default();
        core::serde::Serde::<ContractAddress>::serialize(@to, ref __calldata__);
        core::serde::Serde::<u256>::serialize(@amount, ref __calldata__);
        starknet::account::Call {
            to: self.contract_address,
            selector: selector!("transfer"),
            calldata: core::array::ArrayTrait::span(@__calldata__),
        }
    }
}
```

## Design notes

- **Gated behind `#[unstable(feature: "call_builder")]`** — mirrors the `safe_dispatcher` precedent; lets the API iterate before stabilizing. (Open to shipping it stable instead.)
- **Contract-address only, single builder** — `Call` has no class-hash field, so there is deliberately no Library/Safe matrix; library calls don't map to `Call`.
- **Return type is always `Call`** — the interface method's declared return type is irrelevant when only building a call.

## Does NOT affect the ABI

The contract ABI is generated from the interface trait, not the dispatcher structs (`abi.rs` has zero references to them). This change is purely additive Cairo codegen — no on-chain interface change, nothing for SDKs/indexers/explorers to adapt to. Unused builders are dead-code-eliminated and never reach Sierra.

## Verification

- `cargo test -p cairo-lang-starknet` — **57 passed, 0 failed** (without fix mode; deterministic green).
- The `expand_contract::*` tests run full crate diagnostics (parser + semantic + lowering) with `expect_diagnostics: false`, so the generated Cairo type-checks against corelib (the `starknet::account::Call` path resolves, `Serde`/derives resolve, and implementing the `#[unstable]` trait raises no gate error).
- Regenerated all 14 affected expansion fixtures via `CAIRO_FIX_TESTS=1`. Verified the diff is purely additive: zero pre-existing generated declarations removed and identical `//! >` test-case counts per fixture (the "deletions" in the diff are reordering of derive-generated impls).

> Note: building requires `rustc 1.94` (repo `rust-version`); `cargo +1.94` was used.

## TODO before un-drafting

- [ ] Add a runtime `cairo_level_test` (in `abi_dispatchers_tests.cairo`) asserting `call.to` / `call.selector` / `call.calldata` for a builder-constructed `Call`.
- [ ] corelib doc example + feature note for `call_builder`.
- [ ] Decide stable vs. `#[unstable]` for the initial release.
- [ ] Heads-up to OpenZeppelin Cairo (heaviest interface consumer) re: expansion snapshot tests and the (low) risk of an `I…CallBuilder` name collision.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/crates/cairo-lang-starknet/src/plugin/dispatcher.rs
+++ b/crates/cairo-lang-starknet/src/plugin/dispatcher.rs
@@ -87,6 +87,8 @@ pub fn handle_trait<'db>(
     let mut library_caller_method_impls = vec![];
     let mut safe_contract_caller_method_impls = vec![];
     let mut safe_library_caller_method_impls = vec![];
+    let mut call_builder_signatures = vec![];
+    let mut call_builder_method_impls = vec![];
     let mut forward_method_wrappers = vec![];
     let mut forward_external_fns = vec![];
     let mut forward_impl_method_stubs = vec![];
@@ -101,6 +103,8 @@ pub fn handle_trait<'db>(
     let safe_contract_caller_name = format!("{base_name}SafeDispatcher");
     let library_caller_name = format!("{base_name}LibraryDispatcher");
     let safe_library_caller_name = format!("{base_name}SafeLibraryDispatcher");
+    let call_builder_trait_name = format!("{base_name}CallBuilderTrait");
+    let call_builder_name = format!("{base_name}CallBuilder");
     for item_ast in body.iter_items(db) {
         match item_ast {
             ast::TraitItem::Function(func) => {
@@ -341,6 +345,19 @@ pub fn handle_trait<'db>(
                     ret_decode.clone(),
                     false,
                 ));
+                call_builder_signatures.push(RewriteNode::interpolate_patched(
+                    "\n    #[unstable(feature: \"call_builder\")]\n$func_decl$;",
+                    &[(
+                        "func_decl".to_string(),
+                        call_builder_signature(db, &declaration, "T"),
+                    )]
+                    .into(),
+                ));
+                call_builder_method_impls.push(call_builder_method_impl(
+                    call_builder_signature(db, &declaration, &call_builder_name),
+                    entry_point_selector.clone(),
+                    serialization_code.clone(),
+                ));
                 safe_library_caller_method_impls.push(declaration_method_impl(
                     dispatcher_signature(db, &declaration, &safe_library_caller_name, false),
                     entry_point_selector,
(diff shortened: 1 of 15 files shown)


Yoni-Starkware:
Superseded by a 2-PR stack: #10157 (prep refactor: extract `replace_self_param`) → #10158 (the CallBuilder feature + `.builder()` accessor). Closing this single-PR version.

PR #10159, opened 2026-06-24
feature(sierra-gen): Added support for injecting constant values externally.

orizi:
## Summary

Introduces a reserved extern function `__externally_provided_const__` that, when called in Cairo code, is intercepted during Sierra generation and replaced with a `const_as_immediate` instruction whose value is supplied by a caller-installed `ExternalConstProvider` callback. The provider receives the full path of the extern function and its declared return type, and returns a `ConstValueId` to substitute. The mechanism is wired into both the code generator (`block_generator.rs`) and the local-variables analysis (`local_variables.rs`), where calls to this extern are treated as constants with known AP change rather than resolved as real libfuncs.

A new Salsa input (`SierraGenGroupInput`) stores the optional provider on the database, exposed via `set_external_const_provider` / `external_const_provider` methods on `SierraGenGroup`. A test provider resolving every call to the constant `7777` is installed in the test harness, and a new test file (`externally_provided_const`) verifies that the generated Sierra contains `const_as_immediate<Const<felt252, 7777>>`.

---

## Type of change

Please check **one**:

- [x] New feature
- [ ] Performance improvement

---

## Why is this change needed?

There was no mechanism for build drivers or tooling to inject a compile-time constant into a Cairo program without modifying source code. This is needed when the value of a constant is only known at build time (e.g. a contract address, a configuration value, or a version hash) and must be embedded into the compiled Sierra output without requiring the Cairo source to hard-code it.

---

There was no way to externally supply a constant value during Sierra generation. Any constant had to be written directly in Cairo source.

---

A Cairo program may declare `extern fn __externally_provided_const__() -> T nopanic;` and call it. During Sierra generation, if an `ExternalConstProvider` is installed on the database, the call is replaced with a `const_as_immediate` of the value the provider returns for that function's full path and return type. If no provider is installed and the extern is called, Sierra generation panics.

---

---

## Additional context

The provider's returned `ConstValueId` type is validated against the declared return type of the extern; a mismatch panics rather than producing a type-incorrect Sierra program. The extern function is recognized solely by its name and never appears in the final Sierra output.


orizi:
* **#10163** <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/10163?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10160** <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/10160?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10159** <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/10159?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/10159?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

crates/cairo-lang-sierra-generator/src/block_generator.rs
```diff
@@ -393,6 +405,53 @@ fn generate_statement_call_code<'db>(
...
+            // one, as a mismatch would produce a type-incorrect Sierra program.
+            if value.ty(db)? != return_type {
+                panic!(
+                    "`{EXTERNALLY_PROVIDED_CONST}` provider returned a value whose type does not \
+                     match the declared return type."
+                );
+            }
+            value
+        }
+        None => ConstValueId::from_int(db, return_type, &BigInt::ZERO),
```
Arcticae, line 443: This should emit warnings imo

crates/cairo-lang-sierra-generator/src/block_generator.rs
```diff
@@ -393,6 +405,53 @@ fn generate_statement_call_code<'db>(
...
+    let [output] = statement.outputs[..] else {
+        panic!("`{EXTERNALLY_PROVIDED_CONST}` must have exactly one output.");
+    };
+    // The declared return type, which the provided value must match.
+    let return_type = context.get_lowered_variable(output).ty;
+
+    let value = match db.external_const_provider() {
+        Some(provider) => {
+            let (extern_id, _) = statement.function.get_extern(db).unwrap();
+            let value = provider(db, &extern_id.full_path(db), return_type)?;
```
Arcticae, line 432: I presume `extern_id.full_path` will yield the externs' call site? 
Don't we want more flexibility here and add some parameters to the method?

orizi: it would emit the full path of the extern - which is basically the same path as the contract (+ a bit)

crates/cairo-lang-sierra-generator/src/block_generator.rs
orizi: impossible to emit diagnotics at this stage - so at most - emit a log warning.

orizi (commented):
> @orizi made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/10159)* status: 0 of 8 files reviewed, 2 unresolved discussions (waiting on Arcticae).

PR #10255, opened 2026-07-27
fix(semantic): report an unresolved trait in a negative impl bound

orizi:
## Summary

When a negative impl parameter references an unresolved trait (e.g., a typo like `-MissingTrait<T>`), the compiler was silently swallowing the "Trait not found" diagnostic. This happened because the code used `?` to propagate the `Err` from `neg_impl.concrete_trait`, causing an early return before the diagnostic could be emitted.

The fix wraps the trait param inspection in an `if let Ok(...)` guard, so that the unresolved-trait diagnostic is preserved and surfaced to the user, while the invalid-param-kind check is only performed when the trait resolved successfully.

Two new test cases are added: one for a negative impl with an unresolved trait, and one for the equivalent positive impl parameter (as a control case), both confirming that `E0006: Trait not found` and the resulting `E2311: Trait has no implementation` diagnostics are correctly reported.

---

## Type of change

Please check **one**:

- [ ] New feature
- [ ] Performance improvement

---

## Why is this change needed?

When a negative impl generic parameter contained a misspelled or otherwise unresolved trait, the `?` operator caused an early return from the function, discarding the "Trait not found" diagnostic that had already been recorded. Users would see no actionable error pointing to the typo in the negative impl constraint.

---

A negative impl with an unresolved trait (e.g., `impl Foo<T, -MissingTrait<T>>`) would silently fail to emit the `E0006: Trait not found` diagnostic, leaving the user without a clear indication of what went wrong.

---

The `E0006: Trait not found` diagnostic is now correctly emitted when a negative impl parameter references an unresolved trait, consistent with how positive impl parameters (`+MissingTrait<T>`) already behave.

---

N/A

---

## Additional context

The behavior for positive impl parameters (`+Trait`) was already correct; this fix brings negative impl parameters (`-Trait`) in line with that behavior.

### DIFF
--- a/crates/cairo-lang-semantic/src/expr/test_data/neg_impl
+++ b/crates/cairo-lang-semantic/src/expr/test_data/neg_impl
@@ -392,3 +392,77 @@ error[E2313]: Negative impl has an unsupported generic argument GenericParamCons
  --> lib.cairo:10:5
     bar::<[u32; N]>();
     ^^^
+
+//! > ==========================================================================
+
+//! > Negative impl with an unresolved trait.
+
+//! > test_runner_name
+test_function_diagnostics(expect_diagnostics: true)
+
+//! > function_code
+fn foo() -> felt252 {
+    MyTrait::<felt252>::f()
+}
+
+//! > function_name
+foo
+
+//! > module_code
+trait MyTrait<T> {
+    fn f() -> felt252;
+}
+
+impl BlanketNonCopy<T, -MissomgTrait<T>> of MyTrait<T> {
+    fn f() -> felt252 {
+        7
+    }
+}
+
+//! > expected_diagnostics
+error[E0006]: Trait not found.
+ --> lib.cairo:5:25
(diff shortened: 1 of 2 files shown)


orizi:
* **#10255** <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/10255?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/10255?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

TomerStarkware (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/10255#-:-OyXiFpzATpinI52gXHr:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on eytan-starkware).

PR #10265, opened 2026-07-27
(bug fix): complete all nightly benchmark stages even when one alerts on degradation

orizi:
## Summary

Adds `if: ${{ !cancelled() }}` conditions to all benchmark steps after the first `fail-on-alert` store step in the nightly workflow. This ensures that a performance regression detected in one benchmark suite does not prevent the remaining benchmark suites from running and storing their results, and that the final `git push` to `gh-pages` still executes.

---

## Type of change

Please check **one**:

- [ ] New feature
- [ ] Performance improvement

---

## Why is this change needed?

When the first benchmark store step triggers a `fail-on-alert` degradation alert, GitHub Actions marks the job as failed and skips all subsequent steps by default. This caused the heap benchmarks, LS-flow heap benchmarks, language-server re-execution benchmarks, and the final `gh-pages` push to be silently dropped whenever a single regression was detected.

---

A single benchmark regression would cause all later benchmark suites to be skipped entirely, and their results would never be stored or pushed to `gh-pages`.

---

All benchmark suites run and store their results regardless of whether an earlier suite triggered a degradation alert. The `gh-pages` push also runs unconditionally (as long as the job was not explicitly cancelled), so the full set of benchmark results is always recorded.

---

N/A

---

## Additional context

The `!cancelled()` condition distinguishes between a step failure (which should not block subsequent steps) and an explicit job cancellation (which should still halt execution).

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -88,10 +88,15 @@ jobs:
           fail-on-alert: true
           alert-threshold: '150%'
 
+      # The remaining suites and the final push run even if an earlier store step failed on a
+      # degradation alert (fail-on-alert), so a single regression doesn't drop the results of the
+      # other benchmarks.
       - name: Run heap benchmarks
+        if: ${{ !cancelled() }}
         run: cargo bench --bench dhat_compile --features dhat
 
       - name: Store heap benchmark results
+        if: ${{ !cancelled() }}
         uses: benchmark-action/github-action-benchmark@52576c92bccf6ac60c8223ec7eb2565637cae9ba
         with:
           name: Cairo Heap Benchmarks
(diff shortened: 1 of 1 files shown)


orizi:
* **#10265** <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/10265?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/10265?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

TomerStarkware (approved):
:lgtm:

> @TomerStarkware reviewed 1 file and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/10265#-:-OybtFs93SlKUgi7m6Wz:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on eytan-starkware).

PR #10273, opened 2026-07-29
fix(starknet): don't let the embeddable_as wrapper's component local shadow a parameter

orizi:
## Summary

The generated local variable in `#[embeddable_as]` wrapper functions has been renamed from `component` to `__component__`. This prevents a name collision when the wrapped function itself has a parameter named `component`, which would previously cause the generated code to shadow or conflict with that parameter.

A new constant `COMPONENT_VAR` (`"__component__"`) is introduced in `component.rs` to hold this identifier, following the `__name__` reserved-looking convention used elsewhere in the plugin. All test snapshots have been updated to reflect the new generated variable name, and a new test case is added covering the scenario where a trait method has a parameter named `component`.

---

## Type of change

Please check **one**:

- [ ] New feature
- [ ] Performance improvement

---

## Why is this change needed?

When a component trait method declared a parameter named `component`, the `#[embeddable_as]` plugin would generate a local variable with the same name (`let component = ...` or `let mut component = ...`), causing a name collision. The generated wrapper would then pass the wrong value — the function parameter instead of the component state — to the inner implementation call.

---

The generated wrapper used `component` as the local variable name for the component state, which conflicted with any function parameter also named `component`.

---

The generated wrapper now uses `__component__` as the local variable name, which is unlikely to collide with user-defined parameter names. A dedicated test case verifies that `embeddable_as` works correctly when a method has a parameter named `component`.

---

N/A

---

## Additional context

The fix follows the existing `__name__` naming convention already used for other generated identifiers in the StarkNet plugin (e.g., `__calldata__`, `__dispatcher_return_data__`).

### DIFF
--- a/crates/cairo-lang-starknet/src/plugin/plugin_test_data/components/component
+++ b/crates/cairo-lang-starknet/src/plugin/plugin_test_data/components/component
@@ -1144,13 +1144,13 @@ impl TContractStateDrop: Drop<TContractState>
 > of MyTrait<TContractState> {
     
     fn get(self: @TContractState, addr: ContractAddress) -> u32 {
-        let component = HasComponent::get_component(self);
-        MyInnerImpl::get(component, addr)
+        let __component__ = HasComponent::get_component(self);
+        MyInnerImpl::get(__component__, addr)
     }
     
     fn set(ref self: TContractState, addr: ContractAddress, value: u32) {
-        let mut component = HasComponent::get_component_mut(ref self);
-        MyInnerImpl::set(ref component, addr, value)
+        let mut __component__ = HasComponent::get_component_mut(ref self);
+        MyInnerImpl::set(ref __component__, addr, value)
     }
 }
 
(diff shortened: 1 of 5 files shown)


orizi:
* **#10273** <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/10273?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/10273?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

eytan-starkware (approved):
:lgtm:

> @eytan-starkware reviewed 5 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/10273#-:-OynLDEB6F-A8S-Cj88z:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on TomerStarkware).

PR #10281, opened 2026-08-02
fix(semantic): expand a nested repetition over its own group of captures

orizi:
Stacked on #10270.

A nested repetition now expands over **its own group** of captures rather than the whole flat list.

Captures are stored flat, one list per placeholder, so a nested repetition had no way to say "this iteration's slice". With more than one group at a level the inner block re-walked every capture on every outer step:

```
($($a:expr => [$($b:expr),*]),*) => { $($a + $($b +)*)* 0 }
groups!(1 => [7, 8], 2 => [9])
  before: 1 + 7+8+9+2 + 7+8+9+0
  after:  1 + 7+8+2 + 9+0
```

`MatcherContext` gains `rep_ids`, `rep_parents` and `rep_group_lens`. A repetition is entered once per iteration of its parent, so the parent's current index selects the group and the lengths of preceding groups give the offset. `rep_ids` is keyed by the pattern's syntax node rather than by traversal order — a nested repetition is re-walked on every outer iteration and would otherwise get a fresh id, and no stable identity, each time.

### Tests

Goldens in `expansion_test_data/inline_macros`, all expected values cross-checked against `macro_rules!` under rustc:

- nested repetition matching a different number of times per outer iteration
- three repetition levels, each with several groups
- `three_levels` strengthened to repeat at every level (per review) — `1 => [2 => (3, 4), 5 => (6, 7)], 8 => [9 => (10, 11)]`
- `unsorted_lhs` — a pattern whose placeholders are not ordered by repetition depth
- `multi_vars` — several placeholders bound at each of two depths

The last three were requested on #10270 but cannot live there: that commit still walks the flat list, so they emit wrong output and landing them there would mean blessing it.

PR #10282, opened 2026-08-02
fix(semantic): drive a macro expansion block by the repetition at its own depth

orizi:
Stacked on #10281.

A `$()` expansion block whose only placeholder is bound to a **deeper** repetition than the block's own level expanded wrongly:

```
($([$($x:expr),*]),*) => { $($($x +)*)* 0 }
deeper_only!([1, 2], [3, 4])
  before: 1+2+1+2+0    // group [3,4] dropped entirely, group [1,2] emitted twice
  after:  1+2+3+4+0
```

Nothing was bound at the outer block's depth, so it fell back to driving on `$x` too and both loops walked the same group.

`macro_rules!` accepts this shape and expands it correctly at any depth (verified under rustc 1.96), so the fix is to iterate, not to reject. What rustc *does* reject — `($($a:expr),*) => { $($a + $($a +)*)* 0 }`, "attempted to repeat an expression containing no syntax variables matched as repeating at this depth" — is exactly our E2202, and the existing `max_path_len_in_block >= curr_rep_depth` check is a faithful port of that rule.

### Change

A block nested in `d` blocks drives the pattern repetition at depth `d`, reached by climbing `rep_parents` from a placeholder's own repetition (`repetition_at_depth`). `depth` is threaded through `expand_macro_rule_ex`.

This makes the `(is_iterated, pattern_depth)` ranking from #10270 dead — a block at depth `d` drives a repetition no enclosing block can be iterating — so `find_repetition_driver` loses it, and `repetition_group`'s parent lookup tightens to an `expect` (the no-parent arm stays: every depth-1 block takes it).

`register_repetition_placeholders` no longer lets a shallower repetition overwrite a deeper binding. It re-runs on every iteration of the enclosing repetition, and a nested repetition matching zero times that iteration is never entered, so it would not restore what was overwritten. Without this the depth rule regresses `two_deeper_only!([(1)], [])` into expanding not at all.

### Tests

Seven goldens in `expansion_test_data/inline_macros`, every expected value cross-checked against rustc. Four fail at the parent commit; the other three are guards on the new `expect` and on zero-match inputs, which have masked three separate bugs in this code already.

### DIFF
--- a/crates/cairo-lang-semantic/src/expr/expansion_test_data/inline_macros
+++ b/crates/cairo-lang-semantic/src/expr/expansion_test_data/inline_macros
@@ -627,10 +627,136 @@ macro deeper_only {
 }
 
 //! > expr_code
-deeper_only!([1])
+deeper_only!([1, 2], [3, 4])
+
+//! > expanded_code
+1+2+3+4+0
+
+//! > diagnostics
+
+//! > ==========================================================================
+
+//! > Test a block whose only placeholder is bound two repetition levels deeper.
+
+//! > test_runner_name
+test_expand_expr(expect_diagnostics: false)
+
+//! > module_code
+#[feature("user_defined_inline_macros")]
+macro two_deeper_only {
+    ($([$(($($x:expr),*)),*]),*) => { $($($($x +)*)*)* 0 };
+}
+
+//! > expr_code
+two_deeper_only!([(1, 2), (3)], [(4, 5)])
+
+//! > expanded_code
+1+2+3+4+5+0
+
+//! > diagnostics
+
+//! > ==========================================================================
+
+//! > Test a block driven by a deeper repetition, over an outer repetition matching zero times.
+
+//! > test_runner_name
+test_expand_expr(expect_diagnostics: false)
+
+//! > module_code
+#[feature("user_defined_inline_macros")]
+macro deeper_only {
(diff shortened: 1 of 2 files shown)


orizi:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/10282?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#10282** <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/10282?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/10282?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#10281** <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/10281?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10270** <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/10270?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

PR #10300, opened 2026-08-04
test(semantic): pin the macro repetition operator and trailing separator behaviors

orizi:
Tests only - no production code changes. These goldens pass unchanged on this
branch; their value is pinning behaviors that the upcoming per-group expansion
rewrite must preserve, so that its review shows them as zero-churn.

Repetition operator constraints on a nested, NON-final group
------------------------------------------------------------
Cairo enforces `+`/`?` in validate_repetition_operator_constraints(), after the
whole pattern matched, over every repetition occurrence - including the inner
repetition's occurrence in each group of an outer repetition. The new pins cover
the case where the violating group is not the last one, each with a `*`
counterpart proving the rejection comes from the operator constraint and not from
a plain match failure:

* `($([$($x:ident),+]),*)` on `m!([], [a, b])` -> E2158 no-matching-rule.
* `($([$($x:ident),*]),*)` on `m!([], [a, b])` -> accepted.
* `($([$($x:ident)?]),*)`  on `m!([a b], [c])` -> E2158 no-matching-rule.
* `($([$($x:ident)*]),*)`  on `m!([a b], [c])` -> accepted.

Cross-checked on rustc 1.96.0 (ac68faa20 2026-05-25), every program run with an
invocation because macro_rules! transcriber errors are lazy:

  macro_rules! m { ($([$($x:ident),+]),*) => { 0 }; }
  fn main() { let _ = m!([], [a, b]); }
    error: no rules expected `]`
      note: while trying to match meta-variable `$x:ident`

  macro_rules! m { ($([$($x:ident),*]),*) => { 0 }; }
  fn main() { let _ = m!([], [a, b]); }
    compiles, rustc exit 0

  macro_rules! m { ($([$($x:ident)?]),*) => { 0 }; }
  fn main() { let _ = m!([a b], [c]); }
    error: no rules expected `b`
      note: while trying to match `]`

  macro_rules! m { ($([$($x:ident)*]),*) => { 0 }; }
  fn main() { let _ = m!([a b], [c]); }
    compiles, rustc exit 0

The `?` cases carry no separator because rustc rejects one at definition time:
`($([$($x:ident),?]),*)` gives "error: the `?` macro repetition operator does not
take a separator". Cairo accepts `,?`, so the pins use the rustc-legal form.

The outer level holds 2 groups in every case, and the non-violating group holds 2
captures wherever the operator allows it - `?` caps its own group at 1 by
construction. The `+` invocation is `m!([], [a, b])` rather than `m!([], [a])` to
[...]

PR #10299, opened 2026-08-04
refactor(semantic): build a nested capture tree alongside the flat macro captures

orizi:
Introduces `CaptureTree` (`Leaf`/`Seq`), a per-placeholder tree mirroring the
repetition nesting of the macro pattern that matched it, and has the matcher
build one per placeholder name in `MatcherContext` alongside the existing flat
`captures`, which stays authoritative. Group `k` of a pattern repetition is
element `k` of its `Seq`, so a repetition that matched zero times leaves an
empty `Seq` in the position its groups would have taken - something the flat
representation cannot express. Expansion is untouched.

A debug assertion at the end of a successful match verifies the invariant tying
the two representations together: the in-order leaves of a name's tree are
exactly the name's flat capture list, every captured name has a tree, and a
tree's leaves are all at one nesting level.

Pattern param names are not verified to be unique, so a name may be used at
conflicting nesting positions, which a single tree cannot mirror. Such a name is
recorded as poisoned - its tree is not authoritative and consumers must fall back
to the flat captures - rather than tripping the assertion on legal input.

Pure refactor with zero behavior change: `git diff -- '*test_data*'` is empty,
`cargo test -p cairo-lang-semantic` is green, as are the `cairo-lang-lowering`,
`cairo-lang-plugins`, `cairo-lang-doc` and `cairo-lang-compiler` suites,
`cairo-test tests/bug_samples --starknet` and the corelib macro tests.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

PR #10298, opened 2026-08-04
fix(semantic): reject a macro expansion block with nothing repeating at its depth

orizi:
`ExpansionCheckCtx` already tracks, for every placeholder, its pattern repetition depth
(`placeholder_paths[name].len()`) and the expansion depth it is used at (`curr_rep_depth`).
It now also validates each `$()` block of the expansion itself: a block nested in
`enclosing_depth` other blocks is legal only if it holds a placeholder whose pattern depth
exceeds `enclosing_depth`, i.e. something that actually repeats at this level and can drive
the block. Otherwise the number of repetitions is undetermined, and the new
declaration-time error E2203 is reported on the offending `MacroRepetition` and sets
`rule.err`, so the rule never expands.

Placeholders nested deeper inside the block count as drivers: in `$($($x),*),*` with `$x` at
pattern depth 1, `$x` sits in the inner block but is consumed by the outer repetition too, so
the outer block is legal and only the inner one is rejected.

E2203 was unallocated (main's general band ended at E2201, the parent commit took E2202;
E2300-E2315 is the `InferenceError` sub-band and is not part of this band). The next free
general semantic code is E2204.

## Semantics mirrored from rustc

Probed with rustc 1.96.0 (ac68faa20 2026-05-25) in this change. rustc's transcriber-side
definition errors are LAZY, so every probe below invokes the macro; rustc then reports the
definition-site error upon invocation. Probe sources are transcribed inline.

REJECTED - all four report
`attempted to repeat an expression containing no syntax variables matched as repeating at this
depth`:

* no metavariable: `macro_rules! m1 { () => { $(foo)* }; }` + `m1!();`
  -> caret on `(foo)`.
* depth-0-only metavariable: `macro_rules! m2 { ($a:ident) => { $($a)* }; }` + `m2!(x);`
  -> caret on `($a)`, plus `this similarly named macro metavariable is unrepeatable`.
* over-deep inner block: `macro_rules! m3 { ($($x:ident),*) => { $($($x),*),* }; }`
  + `m3!(a, b)` -> caret on the INNER `($x)`, not the outer block.
* offender not first: `macro_rules! m9 { ($($x:ident),*) => { stringify!($($x),* ; $(bar)*) }; }`
  + `m9!(a, b)` -> caret on `(bar)`; the leading well-formed block is silent.

ACCEPTED:

* broadcast: `macro_rules! m4 { ($p:ident, $($x:ident),*) => { stringify!($($p $x),*) }; }`
[...]

PR #10301, opened 2026-08-04
fix(semantic): expand a macro repetition over its own group of captures

orizi:
Macro expansion read the flat capture list, so a `$()` block was iterated over
*every* value a placeholder ever captured instead of over the values of the group
being expanded. With `($([$($x:expr),*]),*) => { ($([$($x),*]),*) }`,
`m!([1, 2], [3, 4])` expanded to `([1,2,3,4],[1,2,3,4],[1,2,3,4],[1,2,3,4])`; it
now expands to `([1,2],[3,4])`. Expansion now walks the nested `CaptureTree` the
matcher builds (previous node), and the flat machinery is deleted.

The algorithm
-------------
`ExpansionContext` (replacing the `matcher_ctx`-threading `expand_macro_rule_ex`)
keeps `group_indices`: the index of the group being expanded, one per `$()`
expansion block currently entered, outermost first.

* `MacroParam`: `CaptureTree::at(&group_indices)` descends one `Seq` level per
  index. Reaching a `Leaf` before the indices run out is a broadcast - the
  placeholder repeats less deeply than the block, so its single value is used in
  every group. Reaching a `Seq` after the last index means the placeholder is
  nested deeper than the blocks it is used in, which E2198 rejects at declaration
  time; if it happens anyway the E2202 `MissingCapture` error is returned rather
  than panicking, likewise for an out of range index, which E2199 rejects.
* `MacroRepetition`: the number of groups is the length of the `Seq` reached by
  descending with the current index prefix, for each placeholder used anywhere in
  the block whose captures still repeat at its depth. E2203 (previous node)
  guarantees at least one such placeholder exists and E2199 that they all come
  from one pattern repetition, hence agree; both are verified rather than assumed,
  and a violation returns E2202 instead of expanding to something arbitrary.
  Separator emission and its `i + 1 < len` condition are unchanged.

Deleted: `Captures` and `MatcherContext::captures`, `placeholder_to_rep_id` (its
two per-capture inserts and the blanket `for placeholder_name in
ctx.captures.keys()` rebind loop), `repetition_indices` (and the dead
`for i in 0..match_count` fill loop), `find_first_repetition_param`, the previous
node's `capture_trees_are_valid` assertion with `CaptureTree::leaves` /
`is_uniformly_nested`, and `poisoned_capture_trees`. `current_repetition_stack:
Vec<RepetitionId>` became `repetition_depth: usize`, its only remaining use being
the nesting depth; `open_repetition_groups` still runs strictly before the
increment. `is_macro_rule_match` returns `CaptureTrees`, so `MatcherContext` and
`RepetitionId` are private again and both callers (expr and item position) just
pass the trees.

`repetition_match_counts` / `repetition_operators` keep per-ENCOUNTER
`RepetitionId`s exactly as on main - each group's inner repetition gets a fresh
id, which is what validates `+`/`?` per group - and were deliberately not
stabilized across groups. The trailing-separator acceptance and the
separator-consumption loop in the matcher are untouched.

One behavior change on a pattern no check covers yet: with duplicate placeholder
names at conflicting depths (legal today - see the `TODO(Dean): Verify uniqueness
of param names.`) the conflicting values are dropped from the tree instead of
being emitted in an arbitrary order. Where the surviving tree cannot drive the
expansion block, as in `($y:ident, $($y:ident),*) => { $($y),* }`, that is now
reported as E2202; where it is still well formed, as in
`($($p:expr),* ; $($($p:expr),*),*)`, the dropped values simply do not appear and
no diagnostic is produced - E2202 is not a guaranteed backstop for duplicate
names. E2202's `RepetitionWithoutPlaceholder` reason, dead text since E2203 moved
that defect to declaration time, is replaced by `UndeterminedRepetitionCount`, of
which the first shape is the reachable case.

Goldens - expansion (src/expr/expansion_test_data/inline_macros)
---------------------------------------------------------------
Six new cases. Every expected text was derived by running the equivalent
`macro_rules!` on rustc 1.96.0 (ac68faa20 2026-05-25), transcribing through
`stringify!` so the transcriber output itself is observed. The Cairo expansions
are wrapped in a tuple so that they are single expressions. Verbatim rustc runs:

  macro_rules! nested    { ($([$($x:expr),*]),*) => { stringify!($([$($x),*]),*) }; }
  macro_rules! bcast_tok { ($p:ident, $($x:ident),*) => { stringify!($($p $x),*) }; }
  macro_rules! bcast_add { ($p:expr, $($x:expr),*) => { stringify!($($p + $x),*) }; }
  macro_rules! sums      { ($([$($x:expr),*]),*) => { stringify!($((0 $(+ $x)*)),*) }; }
  macro_rules! flatten   { ($([$($x:expr),*]),*) => { stringify!($($($x),*),*) }; }

  nested!([1, 2], [3, 4])   -> [1, 2], [3, 4]
  bcast_tok!(f, a, b)       -> f a, f b
  bcast_add!(10, 1, 2)      -> 10 + 1, 10 + 2
  sums!([1, 2], [])         -> (0 + 1 + 2), (0)
  sums!([], [])             -> (0), (0)
  sums!()                   -> (empty)
  flatten!([1, 2], [3, 4])  -> 1, 2, 3, 4

(a) nested groups, `nested!([1, 2], [3, 4])` -> `([1,2],[3,4])`.
(b) broadcast, `broadcast!(10, 1, 2)` -> `(10+ 1,10+ 2)`, the `$p + $x` form of the
    rustc-verified `f a, f b`.
(c) zero match: `sums!([1, 2], [])` -> `((0 + 1+ 2),(0 ))` and `sums!([], [])` ->
    `((0 ),(0 ))` - an empty inner group expands to nothing while its outer group
    is still expanded. `sums!()` -> `()` is a PRESERVATION pin, not fix evidence:
    with no captures at all the old code also emitted nothing, so it passes on the
    pre-fix branch too.
(d) depth flattening, `flatten!([1, 2], [3, 4])` -> `(1,2,3,4)`, an inner-depth
    placeholder emitted by two nested expansion blocks over a nested pattern.

Every level of every case holds >= 2 captures, and the varying group is not the
first one in (c).

Bless-then-revert: with only this node's four production hunks reverted
(diagnostic.rs, expr/compute.rs, items/macro_call.rs, items/macro_declaration.rs
checked out from the previous node, all three earlier foundation nodes applied),
the five fix-evidence cases fail as `expanded_code` output-tag mismatches -
(a) `([1,2,3,4],[1,2,3,4],[1,2,3,4],[1,2,3,4])`, (b) `10+ 1`, (c)
`((0 + 1+ 2),(0 + 1+ 2))` and `()`, (d) `(1,2,3,4,1,2,3,4,1,2,3,4,1,2,3,4)`.

Goldens - diagnostics (src/expr/test_data/inline_macros)
-------------------------------------------------------
Two new cases, both failing on the reverted branch:

(g) item position, `($([$($name:ident),*]),*) => { $($(fn $name() {})*)* }` with
    `make_fns!([first, second], [third, fourth]);` - no diagnostics. Pre-fix each
    of the four functions is generated once per group, giving four E2118 "The name
    `first` is defined multiple times." errors. Items generated by an item-position
    call are not resolvable from the calling module, so this collision, rather than
    calls to the generated functions, is what makes the case depend on per-group
    expansion.
(h) `($y:ident, $($y:ident),*) => { $($y),* }` -> the new E2202 message, the
    duplicate-name case described above. Pre-fix it emitted `base,one,two` and
    produced E2117 "Parser error in macro-expanded code".

The behavior-pins goldens from the preceding node have ZERO diffs, as does every
other pre-existing case in both files. Exactly one pre-existing golden is churned:

* "Test depth-0 placeholder used inside a repetition driven by a depth-1 var
  (valid)." - its expansion `$($x + $y),*` is now correctly expanded once per
  capture of `y`, so `m!(1 + 2, 3, 4);` yields a two element comma list, which is
  not a single expression. What the case asserts - that the declaration-time
  checks accept a depth-0 placeholder broadcast into a depth-1 block - is
  unchanged, so the expansion was wrapped in a tuple, as its neighbour
  `array![$($base + $x),*]` already is, instead of flipping a "(valid)" case into
  an error. The old single-iteration output was the bug.

Gates: cargo test --profile=ci-dev -p cairo-lang-semantic (106 passed);
./scripts/rust_fmt.sh; ./scripts/clippy.sh --profile=ci-dev;
scripts/check_comment_punctuation.py crates;
cargo run --profile=ci-dev --bin cairo-test -- tests/bug_samples --starknet (68
passed). Fallout beyond the listed gates: cargo check --workspace --all-targets,
./scripts/validate_error_codes.sh, cairo-test corelib (737 passed) and cargo test
over cairo-lang-compiler / -doc / -plugins / -lowering / -starknet / -defs.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>


orizi:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/10301?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#10315** <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/10315?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10314** <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/10314?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10313** <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/10313?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10312** <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/10312?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10311** <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/10311?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10310** <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/10310?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10309** <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/10309?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10308** <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/10308?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10307** <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/10307?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10306** <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/10306?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10305** <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/10305?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10304** <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/10304?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10303** <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/10303?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10302** <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/10302?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10301** <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/10301?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/10301?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#10300** <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/10300?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10299** <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/10299?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10298** <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/10298?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10354** <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/10354?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

crates/cairo-lang-semantic/src/items/macro_declaration.rs
```diff
@@ -110,79 +94,53 @@ impl<'db> CaptureTree<'db> {
...
 
     /// Store the repetition operator for each repetition.
-    pub repetition_operators: OrderedHashMap<RepetitionId, ast::MacroRepetitionOperator<'db>>,
+    repetition_operators: OrderedHashMap<RepetitionId, ast::MacroRepetitionOperator<'db>>,
 }
 
 impl<'db> MatcherContext<'db> {
-    /// Records `value` as the next capture of `name`, in both the flat [`Self::captures`] and the
-    /// nested [`Self::capture_trees`].
+    /// Records `value` as the next capture of `name` in [`Self::capture_trees`].
```
orizi, line 121: over doc

crates/cairo-lang-semantic/src/items/macro_declaration.rs
```diff
@@ -828,10 +742,11 @@ pub struct MacroExpansionResult {
 /// The reason the expansion of a macro rule could not be performed.
 #[derive(Clone, Debug, Eq, Hash, PartialEq, salsa::SalsaValue)]
 pub enum MacroExpansionFailure<'db> {
-    /// A `$( ... )` block in the expansion holds no placeholder, so there is nothing to determine
-    /// how many times it should be repeated.
-    RepetitionWithoutPlaceholder,
-    /// A placeholder in the expansion has no captured value at the current repetition indices.
+    /// The captures of a `$( ... )` block in the expansion do not say how many groups to expand it
```
orizi: can we easily separate these 2 cases?

crates/cairo-lang-semantic/src/items/macro_declaration.rs
```diff
@@ -859,137 +774,167 @@ impl<'db> MacroExpansionError<'db> {
...
-            ast::MacroElement::Repetition(repetition) => {
-                let inner_elements = repetition.elements(db).elements(db);
-                if let Some(param) = find_first_repetition_param(db, inner_elements) {
-                    return Some(param);
-                }
+        }
+        Ok(())
+    }
+
+    /// The number of groups `repetition` is expanded over: the number of groups the placeholders
```
orizi: over doc - reduce all around.

crates/cairo-lang-semantic/src/items/macro_declaration.rs
orizi: Split into `MissingRepetitionDriver` and `ConflictingRepetitionDrivers`, each with its own message.

crates/cairo-lang-semantic/src/items/macro_declaration.rs
orizi: Trimmed - `record_capture` and `group_count` docs reduced (and the stale duplicate-name clause dropped).

crates/cairo-lang-semantic/src/items/macro_declaration.rs
orizi: Trimmed.

crates/cairo-lang-semantic/src/expr/test_data/inline_macros
```diff
@@ -3000,3 +3002,47 @@ fn foo() {
...
+// repetitions of the expansion block cannot be determined.
+macro m {
+    ($y:ident, $($y:ident),*) => { $($y),* };
+}
+fn foo() {
+    m!(base, one, two);
+}
+
+//! > expected_diagnostics
+error[E2202]: The number of repetitions of this macro expansion block cannot be determined from the captured values.
```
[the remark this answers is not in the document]
orizi: Fixed - the golden is re-blessed at this PR's state (the duplicate-name case now reports `MissingRepetitionDriver`'s message here, and E2204 once #10302 lands).

crates/cairo-lang-semantic/src/items/macro_declaration.rs
orizi: Trimmed - the `record_capture` and `group_count` docs are reduced and the stale duplicate-name clause is gone.

crates/cairo-lang-semantic/src/items/macro_declaration.rs
orizi: Split into `MissingRepetitionDriver` and `ConflictingRepetitionDrivers`, each with its own message.

crates/cairo-lang-semantic/src/items/macro_declaration.rs
orizi: Trimmed - the `record_capture` and `group_count` docs are reduced and the stale duplicate-name clause is gone.

orizi (commented):
> @orizi+AGNT made 4 comments and resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/10301)* status: 0 of 6 files reviewed, all discussions resolved.

eytan-starkware (commented):
> @eytan-starkware+AGNT made 5 comments.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/10301)* status: 0 of 6 files reviewed, 4 unresolved discussions (waiting on eytan-starkware, orizi, and TomerStarkware).

___
*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/10301#-P-98-M14RuxKLcuWvZu:-P-98-M14RuxKLcuWvZv:bnabdgl) (no related file):*
Note: the comments below are from an automatic orizi-review run (Claude agents reviewing in Ori's style, findings adversarially verified before posting). Treat with the usual bot skepticism.
___
*[`crates/cairo-lang-semantic/src/expr/expansion_test_data/inline_macros` line 64 at r4](https://reviewable.io/reviews/starkware-libs/cairo/10301#-P-94Ejx5layH3LMRkIf:-P-94Ejx5layH3LMRkIg:bink427) ([raw file](https://github.com/starkware-libs/cairo/blob/043f5545b5312bc64284ad40c0769f3e0be282ef/crates/cairo-lang-semantic/src/expr/expansion_test_data/inline_macros#L64)):*
> ```
> //! > ==========================================================================
> 
> //! > Test nested expansion blocks each expanding over their own group of captures.
> ```

nothing lands in `corelib/src/test/language_features/macro_test.cairo` - it has repetition macros but not a single nested `$( $( ) )`, so the fixed semantics are never actually compiled and run, only text-compared. add the `nested` / `flatten` shapes there too.
___
*[`crates/cairo-lang-semantic/src/items/macro_declaration.rs` line 514 at r4](https://reviewable.io/reviews/starkware-libs/cairo/10301#-P-94Gln3wGUdMlk1t0U:-P-94Gln3wGUdMlk1t0V:bc6fako) ([raw file](https://github.com/starkware-libs/cairo/blob/043f5545b5312bc64284ad40c0769f3e0be282ef/crates/cairo-lang-semantic/src/items/macro_declaration.rs#L514)):*
> ```XML
>     rule: &MacroRuleData<'db>,
>     input: &ast::TokenTreeNode<'db>,
> ) -> Option<CaptureTrees<'db>> {
> ```

doc above is stale now that this returns trees, and the sentence was already broken ("checks if the input the given rule"):

```suggestion
/// Given a macro rule and an input token tree, checks whether the input matches the rule, and
/// returns the values it captured if it does.
```
___
*[`crates/cairo-lang-semantic/src/items/macro_declaration.rs` line 846 at r4](https://reviewable.io/reviews/starkware-libs/cairo/10301#-P-94Cfb-Ql6i4HBkh_p:-P-94Cfb-Ql6i4HBkh_q:bqxatzo) ([raw file](https://github.com/starkware-libs/cairo/blob/043f5545b5312bc64284ad40c0769f3e0be282ef/crates/cairo-lang-semantic/src/items/macro_declaration.rs#L846)):*
> ```XML
>         // The trees are borrowed out of `self` so that appending to the buffers below may borrow it
>         // mutably.
>         let capture_trees = self.capture_trees;
> ```

not needed - the field is a `&'a` reference, so the reborrow is independent of the `&mut self` borrow. drop the local and the two comment lines, `self.capture_trees.get(&name).and_then(...)` compiles as is (checked).
___
*[`crates/cairo-lang-semantic/src/items/macro_declaration.rs` line 879 at r4](https://reviewable.io/reviews/starkware-libs/cairo/10301#-P-94AeK6IChF8HgvwuI:-P-94AeK6IChF8HgvwuJ:bsl27y2) ([raw file](https://github.com/starkware-libs/cairo/blob/043f5545b5312bc64284ad40c0769f3e0be282ef/crates/cairo-lang-semantic/src/items/macro_declaration.rs#L879)):*
> ```XML
>         for index in 0..group_count {
>             self.group_indices.push(index);
>             let expanded = elements
> ```

the pop-before-`?` dance is pointless - the error aborts the whole expansion and `ctx` is dropped, so nothing observes `group_indices`. also hoist `separator(db)` out of the loop instead of re-reading it per group. compiles and the goldens pass:

```suggestion
        let separator = match repetition.separator(db) {
            ast::OptionTerminalComma::TerminalComma(sep) => Some(sep.as_syntax_node().get_text(db)),
            ast::OptionTerminalComma::Empty(_) => None,
        };
        for index in 0..group_count {
            self.group_indices.push(index);
            for element in elements.elements(db) {
                self.expand_node(element.as_syntax_node())?;
            }
            self.group_indices.pop();
            if index + 1 < group_count && let Some(sep) = separator {
                self.res_buffer.push_str(sep);
            }
        }
```

orizi (commented):
> @orizi+AGNT made 4 comments and resolved 4 discussions.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/10301)* status: 0 of 7 files reviewed, all discussions resolved (waiting on TomerStarkware).

___
*[`crates/cairo-lang-semantic/src/expr/expansion_test_data/inline_macros` line 64 at r4](https://reviewable.io/reviews/starkware-libs/cairo/10301#-P-94Ejx5layH3LMRkIf:-P-9MZEBBXztar-Wg-t3:byftqdo) ([raw file](https://github.com/starkware-libs/cairo/blob/043f5545b5312bc64284ad40c0769f3e0be282ef/crates/cairo-lang-semantic/src/expr/expansion_test_data/inline_macros#L64)):*
<details><summary><i>Previously, eytan-starkware+AGNT (Agent AGNT for eytan-starkware) wrote…</i></summary><blockquote>

nothing lands in `corelib/src/test/language_features/macro_test.cairo` - it has repetition macros but not a single nested `$( $( ) )`, so the fixed semantics are never actually compiled and run, only text-compared. add the `nested` / `flatten` shapes there too.
</blockquote></details>

Added `nested_repetitions::test_nested_repetition_expansion` to corelib's macro_test.cairo - a nested `$([$($x:expr),*]),*` flattened and compared at runtime. Note: including an empty `[]` group fails e2e with the spurious legacy-macro E2200 even though the same shape passes as a semantic golden - that pipeline divergence is queued as a separate investigation, and the corelib test uses non-empty groups meanwhile.
___
*[`crates/cairo-lang-semantic/src/items/macro_declaration.rs` line 514 at r4](https://reviewable.io/reviews/starkware-libs/cairo/10301#-P-94Gln3wGUdMlk1t0U:-P-9MZiSBBZwVc2_bMFC:bye7a8t) ([raw file](https://github.com/starkware-libs/cairo/blob/043f5545b5312bc64284ad40c0769f3e0be282ef/crates/cairo-lang-semantic/src/items/macro_declaration.rs#L514)):*
<details><summary><i>Previously, eytan-starkware+AGNT (Agent AGNT for eytan-starkware) wrote…</i></summary><blockquote>

doc above is stale now that this returns trees, and the sentence was already broken ("checks if the input the given rule"):

```suggestion
/// Given a macro rule and an input token tree, checks whether the input matches the rule, and
/// returns the values it captured if it does.
```
</blockquote></details>

Fixed, with the suggested wording (and `captured params` -> the trees it actually returns).
___
*[`crates/cairo-lang-semantic/src/items/macro_declaration.rs` line 846 at r4](https://reviewable.io/reviews/starkware-libs/cairo/10301#-P-94Cfb-Ql6i4HBkh_p:-P-9MYl15EY2juRYwdZZ:b-57bze8) ([raw file](https://github.com/starkware-libs/cairo/blob/043f5545b5312bc64284ad40c0769f3e0be282ef/crates/cairo-lang-semantic/src/items/macro_declaration.rs#L846)):*
<details><summary><i>Previously, eytan-starkware+AGNT (Agent AGNT for eytan-starkware) wrote…</i></summary><blockquote>

not needed - the field is a `&'a` reference, so the reborrow is independent of the `&mut self` borrow. drop the local and the two comment lines, `self.capture_trees.get(&name).and_then(...)` compiles as is (checked).
</blockquote></details>

Dropped the local and the comment.
___
*[`crates/cairo-lang-semantic/src/items/macro_declaration.rs` line 879 at r4](https://reviewable.io/reviews/starkware-libs/cairo/10301#-P-94AeK6IChF8HgvwuI:-P-9MYH97QIScPGZX1pb:b-k1imlt) ([raw file](https://github.com/starkware-libs/cairo/blob/043f5545b5312bc64284ad40c0769f3e0be282ef/crates/cairo-lang-semantic/src/items/macro_declaration.rs#L879)):*
<details><summary><i>Previously, eytan-starkware+AGNT (Agent AGNT for eytan-starkware) wrote…</i></summary><blockquote>

the pop-before-`?` dance is pointless - the error aborts the whole expansion and `ctx` is dropped, so nothing observes `group_indices`. also hoist `separator(db)` out of the loop instead of re-reading it per group. compiles and the goldens pass:

```suggestion
        let separator = match repetition.separator(db) {
            ast::OptionTerminalComma::TerminalComma(sep) => Some(sep.as_syntax_node().get_text(db)),
            ast::OptionTerminalComma::Empty(_) => None,
        };
        for index in 0..group_count {
            self.group_indices.push(index);
            for element in elements.elements(db) {
                self.expand_node(element.as_syntax_node())?;
            }
            self.group_indices.pop();
            if index + 1 < group_count && let Some(sep) = separator {
                self.res_buffer.push_str(sep);
            }
        }
```
</blockquote></details>

Done - the `?` aborts directly now, and the separator is read once before the loop.

orizi (commented):
> @orizi+AGNT made 4 comments.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/10301)* status: 0 of 7 files reviewed, all discussions resolved (waiting on TomerStarkware).

___
*[`crates/cairo-lang-semantic/src/expr/expansion_test_data/inline_macros` line 64 at r4](https://reviewable.io/reviews/starkware-libs/cairo/10301#-P-94Ejx5layH3LMRkIf:-P-KC0yf4NF-i1ZcNCNr:b-a7l6fd) ([raw file](https://github.com/starkware-libs/cairo/blob/043f5545b5312bc64284ad40c0769f3e0be282ef/crates/cairo-lang-semantic/src/expr/expansion_test_data/inline_macros#L64)):*
<details><summary><i>Previously, orizi+AGNT (Agent AGNT for orizi) wrote…</i></summary><blockquote>

Added `nested_repetitions::test_nested_repetition_expansion` to corelib's macro_test.cairo - a nested `$([$($x:expr),*]),*` flattened and compared at runtime. Note: including an empty `[]` group fails e2e with the spurious legacy-macro E2200 even though the same shape passes as a semantic golden - that pipeline divergence is queued as a separate investigation, and the corelib test uses non-empty groups meanwhile.
</blockquote></details>

Added `nested_repetitions::test_nested_repetition_expansion` to corelib's macro_test.cairo. Note: an empty `[]` group fails e2e with the spurious legacy-macro E2200 even though the same shape passes as a semantic golden - that divergence is queued as a separate investigation; the corelib test uses non-empty groups meanwhile.
___
*[`crates/cairo-lang-semantic/src/items/macro_declaration.rs` line 514 at r4](https://reviewable.io/reviews/starkware-libs/cairo/10301#-P-94Gln3wGUdMlk1t0U:-P-KC1SwAr3lOP-8HR_1:b-nhcmq1) ([raw file](https://github.com/starkware-libs/cairo/blob/043f5545b5312bc64284ad40c0769f3e0be282ef/crates/cairo-lang-semantic/src/items/macro_declaration.rs#L514)):*
<details><summary><i>Previously, orizi+AGNT (Agent AGNT for orizi) wrote…</i></summary><blockquote>

Fixed, with the suggested wording (and `captured params` -> the trees it actually returns).
</blockquote></details>

Fixed with the suggested wording (and the return is described as the trees it actually is).
___
*[`crates/cairo-lang-semantic/src/items/macro_declaration.rs` line 846 at r4](https://reviewable.io/reviews/starkware-libs/cairo/10301#-P-94Cfb-Ql6i4HBkh_p:-P-KC0TJ6aAkXWTgLl4-:b-57bze8) ([raw file](https://github.com/starkware-libs/cairo/blob/043f5545b5312bc64284ad40c0769f3e0be282ef/crates/cairo-lang-semantic/src/items/macro_declaration.rs#L846)):*
<details><summary><i>Previously, orizi+AGNT (Agent AGNT for orizi) wrote…</i></summary><blockquote>

Dropped the local and the comment.
</blockquote></details>

Dropped the local and the comment.
___
*[`crates/cairo-lang-semantic/src/items/macro_declaration.rs` line 879 at r4](https://reviewable.io/reviews/starkware-libs/cairo/10301#-P-94AeK6IChF8HgvwuI:-P-KC02j0SaBcqoQgNwD:bt2mpvq) ([raw file](https://github.com/starkware-libs/cairo/blob/043f5545b5312bc64284ad40c0769f3e0be282ef/crates/cairo-lang-semantic/src/items/macro_declaration.rs#L879)):*
<details><summary><i>Previously, orizi+AGNT (Agent AGNT for orizi) wrote…</i></summary><blockquote>

Done - the `?` aborts directly now, and the separator is read once before the loop.
</blockquote></details>

Done - `?` aborts directly, separator read once before the loop.

crates/cairo-lang-semantic/src/items/macro_declaration.rs
```diff
@@ -110,79 +94,51 @@ impl<'db> CaptureTree<'db> {
...
     /// including ones whose repetition matched zero times.
-    ///
-    /// Mirrors [`Self::captures`], which remains the authoritative representation: the leaves of a
-    /// name's tree, in order, are exactly the name's flat capture list.
-    pub capture_trees: OrderedHashMap<SmolStrId<'db>, CaptureTree<'db>>,
-
-    /// The names whose [`Self::capture_trees`] entry could not mirror the pattern, as the pattern
-    /// uses the name at conflicting nesting positions - at two different repetition depths, or
-    /// twice outside any repetition. Param name uniqueness is not verified, so such a pattern
-    /// is accepted; a single tree cannot hold all of the name's values, so consumers must fall
```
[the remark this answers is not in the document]
orizi: Verified at this PR's revision: the dropped value cannot be expanded silently - `$x` at depth 0 finds a `Seq` where a `Leaf` is expected and reports "Macro placeholder 'x' has no captured value in this repetition" (the expansion-failure backstop from the PR below). And a pattern reusing a placeholder name is the very thing #10302 - the next PR in this stack - rejects at declaration time, foreclosing the path entirely; the drop site's doc there states exactly this reachability (the rule is still matched against calls before the declaration error is honored). So within this PR the hole is diagnostic-covered, and one PR up it is gone.

orizi (commented):
> @orizi+AGNT made 1 comment and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/10301)* status: 0 of 7 files reviewed, all discussions resolved (waiting on TomerStarkware).

PR #10330, opened 2026-08-10
(fix): don't prelude-shorten diagnostic paths shadowed by local items

orizi:
### TL;DR

Fix incorrect short-path resolution when a local item shadows a prelude item with the same name.

### What changed?

`check_prelude` now accepts a `Resolver` and verifies that the prelude item's name is not shadowed by a different item in the context module before returning a short (prelude-based) path. If a local item with the same name exists and resolves to a different item than the prelude one, `check_prelude` returns `None`, falling back to the full path. The `Resolver` is also initialized before the first `check_prelude` call so it can be reused across both call sites.

A new diagnostic test case was added to confirm that when a local `struct u128 {}` shadows the prelude's `u128`, the compiler correctly uses the full path `core::integer::u128` and emits a type mismatch error rather than silently resolving to the wrong type.

### How to test?

Run the semantic diagnostic tests:

```
cargo test -p cairo-lang-semantic
```

The new test case `Testing prelude name shadowed by a local item` should pass, confirming that a locally defined type shadowing a prelude type does not cause the short prelude path to be used incorrectly.

### Why make this change?

Previously, if a module defined an item with the same name as a prelude item (e.g., `struct u128 {}`), the path resolver could still emit the short prelude-based name for the original prelude item, even though that short name would resolve to the local item in that context. This produced misleading or incorrect paths. The fix ensures the short path is only used when it unambiguously refers to the intended item.

### DIFF
--- a/crates/cairo-lang-semantic/src/diagnostic_test_data/tests
+++ b/crates/cairo-lang-semantic/src/diagnostic_test_data/tests
@@ -1094,6 +1094,34 @@ INACCESSIBLE
 
 //! > ==========================================================================
 
+//! > Testing prelude name shadowed by a local item
+
+//! > test_runner_name
+test_expr_diagnostics(expect_diagnostics: true)
+
+//! > crate_settings
+edition = "2024_07"
+
+//! > module_code
+struct u128 {}
+
+//! > function_body
+
+//! > expr_code
+{
+    let _x: core::integer::u128 = 1_felt252;
+}
+
+//! > expected_semantics
+
+//! > expected_diagnostics
+error[E2041]: Unexpected argument type. Expected: "core::integer::u128", found: "felt252".
+ --> lib.cairo:4:35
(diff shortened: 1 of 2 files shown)


orizi:
* **#10330** <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/10330?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/starkware-libs/cairo/10330?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

TomerStarkware (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/10330#-:-OzjwvMz0XfGpl2dMlaO:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on orizi).

PR #10360, opened 2026-08-31
Parse the generated function paths in generated Sierra

orizi:
Based on `main`, so it can merge independently of #10359 - and it should merge before or with it, since on that PR alone the compiler emits Sierra its own parser cannot read.

## Summary

Extends the Sierra parser to accept path segments that are braces token trees, by accepting:
- a braces token tree as a path segment of a label - `foo::{closure#0}`, where previously only `ConcreteLabel` immediately followed by braces (`foo{...}`) was accepted;
- `#` within a token tree, which was not a token of the grammar at all.

Both are additions - the existing bracketed (`test::foo[57-168]`) and braced (`{closure@lib.cairo:3:5: 3:7}`) forms still parse.

## Type of change

Please check **one**:

- [ ] New feature
- [ ] Performance improvement

## Why is this change needed?

#10359 names generated functions `test::foo::{closure#0}` / `test::foo::{loop#0}`, and those names reach Sierra function ids and type debug names. With that PR and without this one, the parser cannot read the Sierra the compiler emits. On the `libfuncs_coverage` Sierra as regenerated there, which contains `core::sha256::compute_sha256_byte_array::{loop#0}`:

```
$ cargo run --bin sierra-compile -- \
    crates/cairo-lang-starknet/test_data/libfuncs_coverage__libfuncs_coverage.sierra out.casm
Error: Failed to parse sierra program with: `UnrecognizedToken { token: (233278, Token(35, "{"), 233279),
expected: ["r#\"[a-zA-Z_][a-zA-Z_0-9]*\"#", "\"<\""] }`.
```

It fails at the `{` after `::`, before `#` is even reached. With this change the same file parses and compiles to CASM.

Note that CI on #10359 is fully green, `sierra-updated-check` included - that check verifies the `.sierra` goldens are up to date, not that they can be read back.

A label's path segments had to be identifiers, so a name with a `{...}` segment was a parse error.

A path segment may also be a braces token tree, and `#` is a token tree token, so these all parse:

```
test::foo::{closure#0}
test::foo::{loop#0}
test::foo::{closure#0}::{loop#0}
core::sha256::compute_sha256_byte_array::{loop#0}
```

#10359 - which produces such names. This PR only widens what the parser accepts, so it is harmless on its own.

## Additional context

The cases live in the existing round-trip test, `crates/cairo-lang-sierra/tests/format_test.rs`, which parses Sierra and compares its printed form - so the grammar and the printer are kept in step. Generated paths are added there in each position they occur: a `ut@` user type, a `user@` generic arg of a libfunc, and a function id, with loose spacing around `::` so normalization is covered too. The added function is anchored at the last statement so that no existing statement label is renumbered.

`cargo test -p cairo-lang-sierra` passes in full on `main` with this change, so no existing input parses differently.

https://claude.ai/code/session_01MMkzrvP4FJZyzEYn9B4j7W

### DIFF
--- a/crates/cairo-lang-sierra/src/parser.lalrpop
+++ b/crates/cairo-lang-sierra/src/parser.lalrpop
@@ -108,6 +108,7 @@ TokenTree: String = {
     "/" => <>.to_string(),
     "\\" => <>.to_string(),
     "-" => <>.to_string(),
+    "#" => <>.to_string(),
 }
 
 BracesTokenTree: String = {
@@ -219,6 +220,7 @@ ConcreteLabel: String = {
     "(" <generic_args:GenericArgsString> "," ")" => format!("({generic_args})"),
     "(" ")" => format!("()"),
     <base:ConcreteLabel> "::" <next:BasicLabel> => format!("{base}::{next}"),
+    <base:ConcreteLabel> "::" <next:BracesTokenTree> => format!("{base}::{next}"),
     <base:ConcreteLabel> "::" "<" <generic_args:GenericArgsString> ">" => format!("{base}::<{generic_args}>"),
     <base:ConcreteLabel> "<" <generic_args:GenericArgsString> ">" => format!("{base}<{generic_args}>"),
 }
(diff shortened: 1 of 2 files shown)


TomerStarkware (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/10360#-:-P0WK8PWDknAZ94AcdPj:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on eytan-starkware).

PR #10374, opened 2026-09-12
fix(sierra-to-casm): prevent wide frame offset overflow

DicksonWu654:
## Summary

Prevent valid wide Sierra function parameter and return layouts from panicking during Sierra-to-CASM reference construction. Frame layout now uses wider signed arithmetic, converts each final CASM offset to `i16` with a checked conversion, and returns existing structured compilation errors when an offset is not encodable.

---

## Type of change

- [ ] New feature
- [ ] Performance improvement

---

## Why is this change needed?

A valid 32,766-cell function parameter frame reaches offset `fp - 32768`, but the previous `i16` expression computed an out-of-range intermediate before adding one. Function-call return construction and stack-contiguity validation had symmetric intermediate or sentinel underflows. With overflow checks enabled these layouts panic even though every emitted reference offset is representable.

---

Wide but representable parameter and return layouts could panic while constructing or validating reference expressions.

---

Frame offsets are accumulated in `i64`; final dereference offsets are checked when converted to `i16`. Representable boundary layouts compile normally, while layouts outside the CASM offset range produce structured errors. Regression coverage exercises the exact parameter and function-call return boundaries, including output ordering and full-width stack validation.

---

None.

---

## Additional context

Validation performed on `sierra-minor-update`:

- `./scripts/rust_fmt.sh`
- `./scripts/rust_fmt.sh --check`
- `RUSTUP_TOOLCHAIN=nightly-2024-08-22 cargo test -p cairo-lang-sierra-to-casm wide_` (2 passed)
- `RUSTUP_TOOLCHAIN=nightly-2024-08-22 cargo test -p cairo-lang-sierra-to-casm invocations::function_call::tests` (2 passed)
- `git diff --check origin/sierra-minor-update...HEAD`

### DIFF
--- a/crates/cairo-lang-sierra-to-casm/src/invocations/function_call.rs
+++ b/crates/cairo-lang-sierra-to-casm/src/invocations/function_call.rs
@@ -1,5 +1,3 @@
-use std::collections::VecDeque;
-
 use cairo_lang_casm::casm;
 use cairo_lang_casm::operand::Register;
 use cairo_lang_sierra::extensions::function_call::SignatureAndFunctionConcreteLibfunc;
@@ -8,7 +6,7 @@ use cairo_lang_sierra::extensions::ConcreteLibfunc;
 use super::{
     check_references_on_stack, CompiledInvocation, CompiledInvocationBuilder, InvocationError,
 };
-use crate::references::build_deref_reference;
+use crate::references::{build_deref_reference, ReferenceExpression};
 use crate::relocations::{Relocation, RelocationEntry};
 
 /// Handles a function call.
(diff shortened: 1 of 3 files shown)


orizi (changes requested):
> @orizi made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo/10374)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on DicksonWu654).

___
*[a discussion](https://reviewable.io/reviews/starkware-libs/cairo/10374#-P1TDi440SUxkPAYzAth:-P1TDi456o7lWPIFGgYI:b91alya) (no related file):*
Rather not extend the supported sierra programs without specific reason.
the i16 overflow is a valid failure for using the program.

